PR #102, opened 2022-08-03
feat: Support locations for screens

cmaddox5:
**Asana task**: [[Screenplay] Indicate location on screen name](https://app.asana.com/0/1185117109217413/1202505849835058/f)

A new file needs to be created for this feature to work as expected. That file (`screen_locations.json`) will live in the same directory as `places_and_screens.json`. The contents are a list of objects with the shape `{ "id": "1", "location": "Test" }`. Each object will map to screens with the same id in `places_and_screens.json`. The frontend will then render the location in the accordion panel. If no location exists for a screen, only the screen type is displayed. 

No locations are needed for the config to work. `screen_locations.json` can be an empty list and each screen will render only the screen type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,6 @@ npm-debug.log
 # Local storage of takeover messages
 /priv/alerts.json
 /priv/places_and_screens.json
+/priv/screen_locations.json
 
 .envrc
--- a/assets/js/components/Dashboard/ScreenDetail.tsx
+++ b/assets/js/components/Dashboard/ScreenDetail.tsx
@@ -13,11 +13,14 @@ const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
       ?.label;
   };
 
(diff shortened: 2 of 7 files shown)


PaulJKim:
Code looks good to me, but what was the basis again for adding a new file instead of adding the location as an attribute in the existing `places_and_screens.json` file?

cmaddox5:
> Code looks good to me, but what was the basis again for adding a new file instead of adding the location as an attribute in the existing `places_and_screens.json` file?

I think the primary reason for that is location names will likely be decided by the hardware folks, not us product folks. So we want to have this data live in a more neutral location. Another option would be to add it to the `screens` config, but that would require a lot more changes than what we have here, most of the changes being in `screens`. I don't really have a strong opinion either way, so am happy to bring this up to the team if you think a different option is more appropriate/easier to maintain.

PaulJKim (approved):

PR #103, opened 2022-08-03
feat: DUP simulations

cmaddox5:
**Asana task**: [DUP simulation](https://app.asana.com/0/1185117109217413/1202505849835050/f)

Places with DUPs now show a simulation for the screen. This uses a new `screens` endpoint that displays all three rotations at the same time, formatted to match designs.


hannahpurcell (approved):
Great! I'm impressed with how easy it is to just plug in a new sim (because all the hard work was done in `screens`).

PR #104, opened 2022-08-05
fix: Sort order label and related minor sorting bugs

jzimbel-mbta:
**Asana task**: ad hoc

- `sortLabel` is now computed from `modeLineFilterValue` and `sortDirection`,
  rather than being an independently managed state value. This addressed some
  inconsistency issues.
- `sortDirection` always maps to the correct "Xbound" sort label (based on GTFS data)
  when filtering by line
- "ZXY" label corrected to "ZYX"
- sorting fixed when filtering by bus, commuter rail, and other non-subway routes/modes

❓ **One question**: `sortDirection` is preserved when we change `modeLineFilterValue`. For example if we're sorting in reverse alphabetical order ("ZYX") and switch the mode filter to "Red Line", it starts out in "Northbound" sort order because `sortDirection` is still 1. Do we want that? Or should we always reset `sortDirection` to 0 when changing `modeLineFilterValue`?

(This was already the case before my changes, it was just getting incorrectly labeled. It would be labeled as "Southbound" despite the sort order being "Northbound")


assets/js/components/Dashboard/Dashboard.tsx
```diff
@@ -25,7 +42,8 @@ const Dashboard = (props: { page: string }): JSX.Element => {
   );
   const [statusFilterValue, setStatusFilterValue] = useState(STATUSES[0]);
   const [activeEventKeys, setActiveEventKeys] = useState<string[]>([]);
-  const [sortLabel, setSortLabel] = useState("ABC");
+
+  const sortLabel = getSortLabel(modeLineFilterValue, sortDirection);
```
cmaddox5, line 46: This is more of a React question than a suggestion. Is there a particular advantage to passing state values as parameters? Is that just to allow for calling the function with values that are not in state?

cmaddox5 (approved):
Looks great! Just one question.

assets/js/components/Dashboard/Dashboard.tsx
jzimbel-mbta: It's mostly 6 and 1/2 dozen, but there are a few reasons I prefer to do it this way when possible:
- it forces you to explicitly state in the argument list what variables the function depends on, as opposed to a [closure](https://en.wikipedia.org/wiki/Closure_(computer_programming)) that uses one or more free variables from its containing scope. You get a pure function: variable(s) in -> value out.
- (much less important) a slight performance gain. See below 👇

Technically, the entire body of a component runs anew every time it re-renders, so when you define a helper function inside of the component, that definition is recreated on every re-render, as opposed to a function defined outside at the module level which is defined only once. Overall it doesn't make a huge difference performance-wise, though. Function definitions are fast.

PR #107, opened 2022-08-09
Screen Simulations - PAESS

PaulJKim:
**Asana task**: [PA/ESS simulation](https://app.asana.com/0/1185117109217413/1202505849835052/f)

**Description**
This PR adds PAESS simulations in iframes to Screenplay.

The signs are grouped by zone into columns and each column is always present, even if there are no signs in one of the zone. This was done to mimic how Signs UI lays out it's UI which is that it dedicates the left column to south and westbound sign zones, the center column for mezzanine and center platform sign zones, and the right column for north and eastbound sign zones.

The caveat to this pattern is that in the cases where there are only two sign zones present, it's usually because there is one sign zone for the mezzanine and one sign zone for a single platform. This is the case at Malden Center on the OL, for example. In these cases in Signs UI, we typically use the available zones i.e. "m" or "n" but we re-label them to what reflects the actual station i.e. "lobby" and "platform" in Malden Center's case. This is why I added the `label` field in the screen model so that we can add custom labeling when necessary by making changes to the `places_and_screen.json` file.


assets/js/components/Dashboard/PaessDetailContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+        key={`${stationCode}-left`}
+        className={`screen-detail__iframe-container screen-detail__iframe-container--pa_ess`}
+        data-testid="paess-col-left"
+      >
+        {leftScreens.length > 0 && (
+          <div>
+            {leftScreens.map(
+              (screen) =>
+                screen.station_code != null &&
+                screen.zone != null && (
```
PaulJKim, line 43: If there's a cleaner, more javascript-y way to avoid doing explicit null checks for these nullable properties, please let me know 😃

assets/js/components/Dashboard/PaessScreenDetail.tsx
```diff
@@ -0,0 +1,49 @@
...
+  stationCode: string;
+  zone: string;
+  label?: string;
+}
+
+const PaessScreenDetail = (props: PaessScreenDetailProps): JSX.Element => {
+  const generateSource = () => {
+    // @ts-ignore Suppressing "object could be null" warning
+    const { environmentName } = document.getElementById("app").dataset;
+    return environmentName === "dev"
```
cmaddox5, line 13: In the near future, I believe the goal is to add a `dev-green` to Screenplay. Is there also a `dev-green` for SignsUI? If so, we should add a conditional for that URL. If not, we'll need to make sure the `dev-green` does not use `prod`.

```suggestion
    return environmentName === "dev" || environmentName === "dev-green"
```

assets/js/components/Dashboard/PaessDetailContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  return (
+    <Row className="h-100" md={3}>
+      <Col
+        key={`${stationCode}-left`}
+        className={`screen-detail__iframe-container screen-detail__iframe-container--pa_ess`}
+        data-testid="paess-col-left"
+      >
+        {leftScreens.length > 0 && (
+          <div>
```
cmaddox5, line 39: Should be able to remove this `div`.

assets/js/components/Dashboard/PaessDetailContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+          </div>
+        )}
+      </Col>
+      <Col
+        key={`${stationCode}-center`}
+        className={`screen-detail__iframe-container screen-detail__iframe-container--pa_ess`}
+        data-testid="paess-col-center"
+      >
+        {centerScreens.length > 0 && (
+          <div>
```
cmaddox5, line 61: Same as above.

assets/js/components/Dashboard/PaessDetailContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+          </div>
+        )}
+      </Col>
+      <Col
+        key={`${stationCode}-right`}
+        className={`screen-detail__iframe-container screen-detail__iframe-container--pa_ess`}
+        data-testid="paess-col-right"
+      >
+        {rightScreens.length > 0 && (
+          <div>
```
cmaddox5, line 83: Same as above.

assets/js/components/Dashboard/PlaceRow.tsx
```diff
@@ -89,7 +89,20 @@ const PlaceRow = (props: PlaceRowProps): JSX.Element => {
...
+
+          paEssGroupedByRoute.has(routeLetter)
+            ? paEssGroupedByRoute.get(routeLetter)?.push(paEssScreen)
+            : paEssGroupedByRoute.set(routeLetter, [paEssScreen]);
+        }
+      });
+
+      paEssGroupedByRoute.forEach((screens) => {
+        groupedScreens.push(screens);
+      });
```
cmaddox5, line 105: To help simplify the logic that lives in this function, I would pull this logic out into it's own helper function.

assets/js/components/Dashboard/PaessDetailContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const centerScreens = props.screens.filter(
+    (screen) =>
+      screen.zone != null && zonePositions.center.includes(screen.zone)
+  );
+  const rightScreens = props.screens.filter(
+    (screen) => screen.zone != null && zonePositions.right.includes(screen.zone)
+  );
+
+  return (
+    <Row className="h-100" md={3}>
```
cmaddox5, line 32: We do not currently support anything under the `lg` breakpoint, so I would change this to `lg` for consistency. If we want to support others, we can make changes everywhere at the same time.

assets/js/components/Dashboard/ScreenDetail.tsx
```diff
@@ -14,6 +15,14 @@ const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
...
 
+  const isPaess = props.screens.some((screen) => screen.type === "pa_ess");
+
+  const getPaessRoute = () => {
+    return props.screens[0].station_code
+      ? props.screens[0].station_code.charAt(0).toLowerCase()
+      : "";
+  };
+
   const getScreenLocation = () =>
```
cmaddox5, line 26: This function needs to be changed to allow for getting the "location" of the countdown clocks. The designs show the route/mode as the location.

assets/css/screen-detail.scss
```diff
@@ -3,6 +3,34 @@
   border-radius: 8px;
   padding: 24px 16px;
 
+  &--paess-g {
```
cmaddox5, line 6: Colors look good, but there seems to be some vertical alignment issues. In the designs, the location "header" is a bit slimmer than what is currently being shown, and the `margin-top` is designed to be less than other screens. The height can be solved by changing `70px` to `40px` in the below `linear-gradient` functions (is that what they are called in CSS?). The margins may take a little more work. I think I have the needed changes but it's across multiple files so the suggestion feature won't work. Just reach out on Slack when you get time and I'll show you what I have.

assets/js/components/Dashboard/ScreenDetail.tsx
```diff
@@ -41,27 +50,44 @@ const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
...
+        <div
+          className={`screen-detail__screen-type-location${
+            isPaess && getPaessRoute() == "s"
+              ? " screen-detail__screen-type-location--paess-s"
+              : ""
+          }`}
+        >
           {translateScreenType()} {getScreenLocation()}
         </div>
         <div className="screen-detail__report-a-problem-button">
```
cmaddox5, line 68: This button should be hidden for PA/ESS blocks.

cmaddox5 (changes requested):
This is looking great! I have a few suggestions first. Most won't require much, but one might take a few changes in a couple of different files.

assets/js/components/Dashboard/PaessDetailContainer.tsx
PaulJKim: So I mainly added this `div` to achieve the center alignment that I was looking for in the cases where there are two signs in a given column. With one sign, the center alignment looks just fine, but when there's two they end up side by side in the column instead of vertically stacked. Adding the div seemed to do the trick, but certainly let me know if there's a better way to go about it!

assets/js/components/Dashboard/ScreenDetail.tsx
PaulJKim: Ah yeah I was wondering about that since an alternative solution would be to add locations data to the new json file for each of the PAESS screens. That sounds like a lot of annoying work for something that we could do fairly easily in this helper function though so I'll go ahead and add that here

assets/css/screen-detail.scss
PaulJKim: Had to do a little bit extra to get around the padding/margin stuff. Basically, removing the padding caused other simulations to end up more flushed to the bottom of their containers so instead, I just added some PAESS specific overrides of the padding values and then set the margins to the right values below in `screen-detail__screen-type-location--paess`. That seems to have worked in keeping the rest of the screen previews the same while letting us adjust the PAESS header to the way we need it

assets/js/constants/constants.ts
```diff
@@ -37,7 +37,7 @@ export const SCREEN_TYPES = [
   },
   { label: "Elevator", ids: ["elevator"] },
   { label: "Outfront Media", ids: ["ofm"] },
-  { label: "PA/ESS", ids: ["pa_ess"] },
+  { label: "PA ESS", ids: ["pa_ess"] },
```
PaulJKim, line 40: Removed this slash to match the figma designs

PaulJKim: That said, it does mean that this will affect what is in the drop down as well.

cmaddox5 (approved):
Looks great!

PR #110, opened 2022-08-10
polish: Align filter dropdowns

cmaddox5:
**Asana task**: [📺▶️ : 🇵🇱 | Align filter controls with Places table columns](https://app.asana.com/0/1185117109217413/1202737926527313/f)

Filter dropdowns are now aligned with the column that they are a filter for. I also cleaned up/renamed a few CSS classes so that there is more consistency.

### DIFF
--- a/assets/css/dashboard.scss
+++ b/assets/css/dashboard.scss
@@ -1,8 +1,7 @@
 .place-list__header-row {
-  margin-left: 12px;
-  margin-bottom: 24px;
   display: flex;
   align-items: center;
+  margin-bottom: 24px;
 }
 
 .place-list__sort-label {
@@ -13,4 +12,5 @@
   letter-spacing: 1px;
   margin-right: auto;
   cursor: pointer;
+  width: fit-content;
 }
--- a/assets/css/dashboard.scss
+++ b/assets/css/dashboard.scss
@@ -2,15 +2,15 @@
   display: flex;
   align-items: center;
   margin-bottom: 24px;
-}
 
-.place-list__sort-label {
-  color: white;
-  font-weight: 900;
-  font-size: 16px;
-  line-height: 24px;
-  letter-spacing: 1px;
-  margin-right: auto;
-  cursor: pointer;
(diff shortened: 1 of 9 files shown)


PaulJKim (commented):
Sorry for the late review on this! Overall it looks great, but I wonder if we should make the drop-downs go the full length of the column that they're set over, or just wide enough to fit the widest possible text on the dropdown + the (X) button. I found it a little weird when the mode dropdown extends to the left after selecting a mode. The other two dropdowns are a bit better imo since they're justified on the left. Happy to leave this for another polish task if we think it's necessary though.

cmaddox5:
> Sorry for the late review on this! Overall it looks great, but I wonder if we should make the drop-downs go the full length of the column that they're set over, or just wide enough to fit the widest possible text on the dropdown + the (X) button. I found it a little weird when the mode dropdown extends to the left after selecting a mode. The other two dropdowns are a bit better imo since they're justified on the left. Happy to leave this for another polish task if we think it's necessary though.

Sorry for the delay. Finally getting back around to Screenplay stuff. This is a good call out. I don't think I ever really paid attention to the width before but now that you mentioned it, I can't unsee it. Went ahead and got that fixed. Widths I picked come from the designs. I also fixed a border radius issue when the clear button is visible.

PaulJKim:
Looking great! Only thing now is that the following two pieces of text seem to overflow out of the dropdown button:
Screen Shot 2022-08-23 at 11 01 42 AM
Screen Shot 2022-08-23 at 11 01 59 AM

cmaddox5:
> Looking great! Only thing now is that the following two pieces of text seem to overflow out of the dropdown button: Screen Shot 2022-08-23 at 11 01 42 AM Screen Shot 2022-08-23 at 11 01 59 AM

Yikes! Good catch. Should be good now. Also, I noticed in the designs that `Emergency takeovers` has actually been shortened to just `Emergency`. Got that changed, too.

PaulJKim (approved):
Looks perfect!

PR #115, opened 2022-08-25
Adjust height of screens that need a scrollbar; change default dev port

hannahpurcell:
**Asana task**: [📺▶️ : 🇵🇱 | Scrollbar in simulations](https://app.asana.com/0/1185117109217413/1202736912335100)

Pairs with: https://github.com/mbta/screens/pull/1481.

TBD whether there should be more or less padding on the sims, now that a scrollbar padding has been added.

### DIFF
--- a/assets/css/screen-detail.scss
+++ b/assets/css/screen-detail.scss
@@ -75,7 +75,7 @@
 
   &--bus_eink_v2,
   &--gl_eink_v2 {
-    height: 480px;
+    height: 503px;
   }
 
   &--bus_eink,
@@ -96,7 +96,7 @@
   }
 
   &--dup {
-    width: 832px;
+    // width: 832px;
   }
 
   &--pa_ess {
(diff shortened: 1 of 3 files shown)


assets/css/screen-detail.scss
```diff
@@ -75,7 +75,7 @@
 
   &--bus_eink_v2,
   &--gl_eink_v2 {
-    height: 480px;
+    height: 503px;
```
cmaddox5, line 78: Are these heights changed to keep the vertical scrollbar hidden?

assets/css/screen-detail.scss
```diff
@@ -96,7 +96,7 @@
   }
 
   &--dup {
-    width: 832px;
+    // width: 832px;
```
cmaddox5, line 99: Can this be removed?

cmaddox5 (approved):
Looks good! Just a couple of questions.

assets/css/screen-detail.scss
hannahpurcell: Sorta, but also no -- I changed these values and the heights for `.screen-detail__iframe` to remove the vertical scrollbar, but it turns out that I only needed to change the values for the iframe and not the iframe-container. So I've reverted these ones, because it helps remove unnecessary padding.

PR #117, opened 2022-08-25
polish: Disable status filter.

cmaddox5:
**Asana task**: [Filter dropdowns: Disable Status filter](https://app.asana.com/0/0/1202839698975985/f)

Status won't be used until overrides are implemented. Disabling the filter until then.

### DIFF
--- a/assets/css/filter-dropdown.scss
+++ b/assets/css/filter-dropdown.scss
@@ -12,11 +12,6 @@
     }
   }
 
-  &.disabled {
-    opacity: 0.4;
-    pointer-events: none;
-  }
-
   .filter-dropdown__clear-button {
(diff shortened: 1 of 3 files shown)


cmaddox5:
@hannahpurcell That's a good idea. I added a commit giving the disabled filter dropdown a "Coming Soon!" tooltip. Do you want to go ahead and include it and ask design if we should keep it, or hold off merging until we get an answer?

hannahpurcell:
@cmaddox5 I think we'll need to make the "coming soon"s match each other, but leave it with the tooltip as-is and we can ask design which they prefer. Regardless, let's merge it, so they can see everything they need to make a decision

PR #119, opened 2022-08-29
Restyled sidebar and columns for 1280px viewport

hannahpurcell:
**Asana task**: [Make Screenplay work in a 1280px wide window](https://app.asana.com/0/inbox/1199335143657941/1202858275726791/1202886247076499)

Sidebar has been shrunk to minimum decent-looking width, and columns were adjusted so that widths don't need to be hardcoded but the longest text never gets cut off

### DIFF
--- a/assets/css/filter-dropdown.scss
+++ b/assets/css/filter-dropdown.scss
@@ -1,4 +1,5 @@
 .filter-dropdown__button-group {
+  width: 100%;
   .btn-group .btn {
     & .btn {
       background-color: #165c96 !important;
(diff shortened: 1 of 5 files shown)


cmaddox5 (approved):
Very slick! Looks great at the smaller size. I will say that the buttons look a little too big at 1920, but if we are wanting to focus on 1280 then this will work great.

hannahpurcell:
Mmm yeah, I agree the buttons look shrimpy at bigger sizes. Guess we'll wait to see what the PIOs see on the reg and adjust sizing and layout

PR #120, opened 2022-09-13
refactor: Backend file organization

cmaddox5:
**Asana task**: ad-hoc

Because we are about to start on alerts for Screenplay, I think it's a good idea to separate Screenplay and OFT files into their own folders. No logic changes were made, just file locations and module name changes.


hannahpurcell (approved):
I think this works. Eventually, the Outfront tool will be re-integrated into Screenplay, and then this separation won't be needed. At first, I wondered if the directory split could happen higher up, instead of having two directories in each of `controllers`, `templates`, and `views`, but I think that might actually cause more work than help. So I'm ok to give this a try

cmaddox5:
@hannahpurcell My worry with splitting it higher up is breaking some requirements of Phoenix. I took this route because our needs are somewhat similar to how v1 and v2 are split up in screens. Another idea I had was to leave all things in the new Screenplay where they were  (i.e. `/dashboard`, `/alerts`, etc.) and move all OFT related files into their own folder. That way when we merge the two, we can just refactor by deleting the OFT folders. Would that be better?

hannahpurcell (approved):
Mmm yes, I prefer this! Thanks!

PR #121, opened 2022-09-13
feat: Places list filter details / action bar

jzimbel-mbta:
**Asana task**: [Filter details row](https://app.asana.com/0/1185117109217413/1202322822031428/f)

This adds an "action bar" that appears at the top of the places list when one or more filters are selected. The bar shows a count of matching places and a count of the total screens at those places, as well as additional controls to hide screenless places, bulk edit (currently always disabled), and reset all filters.

The "toggle screenless places" button only appears when at least one place in the filtered list is screenless.

- [x] fix existing and add new tests
- [x] address merge conflicts

:warning::question: One question I have that I couldn't find the answer to in the designs: when the user is filtering by a specific screen type, should the total count of screens at all filtered places respect that?

For example, user is filtering by route/mode: GREEN LINE and screen type: PRE-FARE. We currently show Government Center only, but the screen count is 8, not 2, because all screens at Gov Center (including PA/ESS) are still included.
image

## More screenshots

Screen Shot 2022-09-13 at 5 10 27 PM
Route filter applied.

Screen Shot 2022-09-13 at 5 10 35 PM
Route filter applied, screenless places hidden.

Screen Shot 2022-09-13 at 5 10 46 PM
Route filter and screen type filter applied. ("Screenless places" button hidden since the filtered list contains no screenless places)


assets/js/components/Dashboard/PlacesActionBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+  onClickToggleScreenlessPlaces: () => void;
+}
+
+const getPlaceCount = (places: Place[]) => {
+  return places.length;
+};
+
+const getScreenCount = (places: Place[]) => {
+  // Get all screens within the filtered places.
+  // TODO: Should we also limit the count to only include screens of the filtered screen type, when that filter is selected?
```
cmaddox5, line 26: I don't think we've addressed this particular question before, but a similar question was brought up for alerts. For alerts, someone asked if you are filtering by a specific alert should only screens showing that alert be displayed. The answer to that was no, show all screens at all places showing that alert on at least one screen.

assets/js/components/Dashboard/Dashboard.tsx
```diff
@@ -40,9 +41,27 @@ const Dashboard = (props: { page: string }): JSX.Element => {
...
+    setModeLineFilterValue(MODES_AND_LINES[0]);
+    setScreenTypeFilterValue(SCREEN_TYPES[0]);
+    setStatusFilterValue(STATUSES[0]);
+    setShowScreenlessPlaces(true);
+  };
+
+  const goToHome = () => {
+    // This approach is difficult to maintain--every time we add new state values,
+    // we need to remember to add a line here to "reset" it.
+    // TODO: Should the home button simply be a link to /dashboard? That would reload the page,
```
cmaddox5, line 58: Ahh, that sounds like exactly what we want. I know when the button is clicked, all state should be wiped out. I believe that will include filters on the Alerts page as well. If that link will handle all of that, that would be great.

cmaddox5 (approved):
Looks great! Once CI is passing, this should be good to go. I left a couple of comments on some TODOs, but I don't see anything that needs changing as is.

PR #123, opened 2022-09-16
polish: Capitalize terminal stops

cmaddox5:
[Notion Page](https://www.notion.so/mbta-downtown-crossing/Columns-on-places-list-first-one-is-too-short-mode-is-too-long-13867c42f2a2408094d2784abc5e45fa)

Because we need the list of stations in stop order, we can add a boolean to terminal stop entries. Also no longer need to capitalize names in data. Let the helper function do it.


assets/js/components/Dashboard/PlaceRow.tsx
```diff
@@ -151,6 +151,25 @@ const PlaceRow = (props: PlaceRowProps): JSX.Element => {
     )?.inlineMap;
   };
 
+  const formatStationName = (stationName: string) => {
+    let isTerminalStop = false;
+    if (props.filteredLine && props.filteredLine !== null) {
```
hannahpurcell, line 156:
```suggestion
    if (props.filteredLine) {
```
^ this verifies `filteredLine` exists _and_ is truthy

hannahpurcell, line 156:
```suggestion
    if (props.filteredLine) {
```
^ this verifies filteredLine exists and is truthy

assets/tests/components/dashboard.test.tsx
```diff
@@ -260,32 +260,37 @@ describe("Dashboard", () => {
...
           ]);
         });
 
         fireEvent.click(getByTestId("sort-label"));
         await waitFor(() => {
+          console.log(
+            getAllByTestId("place-name").map(
+              (placeName) => placeName.textContent
+            )
+          );
```
hannahpurcell, line 280: remove console log?

assets/js/components/Dashboard/PlaceRow.tsx
```diff
@@ -151,6 +151,25 @@ const PlaceRow = (props: PlaceRowProps): JSX.Element => {
...
+      const line = STATION_ORDER_BY_LINE[props.filteredLine.toLowerCase()];
+      const terminalStops = line.filter((line) => line.isTerminalStop);
+      isTerminalStop = terminalStops.some((stop) => stationName === stop.name);
+    } else {
+      isTerminalStop = Object.keys(STATION_ORDER_BY_LINE)
+        .filter((key) => !["silver", "mattapan"].includes(key))
```
hannahpurcell, line 162: Why weed out these two routes?

assets/js/components/Dashboard/PlaceRow.tsx
hannahpurcell: If the only thing that impacts whether a stop is capitalized / not capitalized is that they are a terminal stop, and this value is hardcoded in `stationOrder.ts`, then I'm not sure why the if/else is needed

assets/js/components/Dashboard/PlaceRow.tsx
cmaddox5: Mostly just because they are unnecessary lookups. Nothing will return true as a terminal stop in these in the current code. For SL, I think that makes sense because we do not show the SL in its entirety in the place list. Does that make sense for Mattapan? Do we have terminal stops for that line? I know there is a start and end to it but I wasn't sure if they are considered that in the same way subway does.

assets/tests/components/dashboard.test.tsx
cmaddox5: Oof. Thanks.

assets/js/components/Dashboard/PlaceRow.tsx
cmaddox5: I could have sworn that wasn't working at one point... Definitely is the right way to do it though.

As for the reason for the if/else, it was mostly to avoid terminal stop lookups in lines that are not currently in the filter. If we are filtering on RL, we don't need to know the terminal stops for GL. Does that make sense?

assets/js/components/Dashboard/PlaceRow.tsx
hannahpurcell: No matter what, every place row requires exactly 1 lookup to decide whether its name should be capitalized or not, right? The lookup should just be "am I a terminal stop?" And this code gets executed for every single PlaceRow individually, so no place row is aware of the "terminal-ness" of the other rows. I'm not sure how this conditional prevents any lookups.

cmaddox5: We can only reliably look in the list for a single route if a route filter is active. Just knowing if a stop is a terminal stop is not quite enough when a filter is applied. For Gov't Center, it is a terminal stop on the GL but not on the BL. If we went with the "am I a terminal stop?" approach, it would be capitalized for BL when it should not be. That's the primary reason for the conditional. So when no filter is applied, we find if a stop is a terminal stop for any line. If a filter is applied, we find if a stop is a terminal stop for only the line in the filter.

hannahpurcell: OHHH! Multi-route stops totally went over my head. Gov't Center was the perfect example, thank you

assets/js/constants/stationOrder.ts
```diff
@@ -1,6 +1,7 @@
 export interface Station {
   name: string;
   inlineMap?: string;
+  isTerminalStop?: boolean;
```
hannahpurcell, line 4: Blue line needs isTerminalStop values

assets/js/components/Dashboard/PlaceRow.tsx
```diff
@@ -152,6 +152,25 @@ const PlaceRow = (props: PlaceRowProps): JSX.Element => {
...
+        .some((key) => {
+          const line = STATION_ORDER_BY_LINE[key];
+          const terminalStops = line.filter((line) => line.isTerminalStop);
+          return terminalStops.some((stop) => stationName === stop.name);
+        });
```
hannahpurcell, line 168: Mind leaving in-line comments for some of this? The "why we do this" won't be clear down the road.

hannahpurcell (approved):
thanks!

PR #125, opened 2022-09-19
polish: PlaceRow hover

cmaddox5:
[Notion](https://www.notion.so/mbta-downtown-crossing/Rows-don-t-have-hover-state-bdd69570112543bfabbf0b054e5672f2)

I think the only hover effect is `background-color`. Added that to places that have screens.

### DIFF
--- a/assets/css/place-row.scss
+++ b/assets/css/place-row.scss
@@ -5,6 +5,10 @@
   margin-bottom: 2px;
   border-bottom: 0px;
 
+  &:hover {
+    background-color: $cool-gray-40;
+  }
+
   &.open {
     background-color: $cool-gray-20;
     padding: 0px 0px 16px;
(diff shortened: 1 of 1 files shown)


assets/css/place-row.scss
```diff
@@ -5,6 +5,10 @@
   margin-bottom: 2px;
   border-bottom: 0px;
 
+  &:hover:not(.disabled) {
+    background-color: $cool-gray-40;
+  }
```
jzimbel-mbta, line 10: Does applying this to `.place-row` make the hover effect apply to whole expanded rows? Do we want that?

jzimbel-mbta (approved):
Approved assuming my Q doesn't require any changes, otherwise feel free to re-request after updating

assets/css/place-row.scss
cmaddox5: Good catch. Based on the designs, we don't want hover effects on expanded rows.

jzimbel-mbta (approved):

PR #126, opened 2022-09-19
tweak: Omit hidden-from-screenplay screens when generating places_and_screens.json

jzimbel-mbta:
**Asana task**: ad hoc / pre-launch polish

This adds a line to the script that generates places_and_screens.json, making it respect the new `hidden_from_screenplay` config field when fetching screens config.

Plus one small refactor to the script: replaced an `Enum.reduce` with an `Enum.flat_map` that does the same thing.

Tested informally by hiding 2 screens in screens-dev admin and running the script pointed at dev with some debug lines:
```
$ mix run scripts/fetch_places_and_screens.exs --environment dev
BEFORE screen count: 206
AFTER screen count: 204
```


hannahpurcell (approved):
Looks great!

PR #129, opened 2022-09-20
Update fetch script for PAESS Bus and Silver Line

PaulJKim:
**Asana task**: [Update fetch script for PA/ESS signs](https://app.asana.com/0/1185117109217413/1202997227409777/f)

**Description**
The fetching script was reading from the realtime-signs config which only covers subway. There was a JSON file added to the Signs UI repo that contains bus configs as well so this PR adds logic to leverage that to automatically add bus/Silver line PAESS configs.

The hurdle was that this new JSON file doesn't contain any GTFS stop Ids that we could use to map them to the content being built in the script. The script instead tried to match the realtime-signs id on the `name` field for each `place` being built out by the script. While this approach is a bit loose, it works out since the number of places with busways is fairly small so there aren't any duplicate matches.

I also added a new JSON file in the screenplay repo for custom labeling. This is something we'll need to more closely match Signs UI as some stations require these label in order to more clearly reflect the actual station layout. Some examples are Kenmore, Park Street, and JFK/UMass. I found the most straightforward approach to be to just manually define these in a JSON file, but I'm open to other suggestions.


scripts/fetch_places_and_screens.exs
```diff
@@ -369,6 +369,9 @@ platform_to_stop_map =
   end)
   |> Enum.into(%{})
 
+{:ok, labels} = File.read("scripts/paess_labels.json")
```
cmaddox5, line 372: Just out of curiosity, do these custom labels change often? Or are changes pretty well communicated so we know to change them here when they change in Signs UI?

cmaddox5 (approved):
Looks great! Just have one question, but the answer has no impact on the code itself.

PR #130, opened 2022-09-20
Add filter dropdowns and sort for alert page

PaulJKim:
**Asana task**: [[Screenplay] Add primary filter dropdowns / sort ](https://app.asana.com/0/1185117109217413/1202932249366383/f)

This PR adds filter dropdowns and enabled sorting of alerts by End time, either ascending (soonest end first) or descending (latest end first), and defaults to comparing alert start time when the End times are similar (for example if they are both `null`).

The filtering by mode or line works by leveraging the `informed_entity` field for Alerts data. If any of the informed entities in an alert match the Line or mode selected in the dropdown, it will be included in the result set. For Commuter Rail, Bus, and Ferry, we have some special handling to use the `route_type` field instead since those modes are not separated out by line in Screenplay.


assets/tests/components/placesPage.test.tsx
```diff
@@ -11,27 +11,11 @@ beforeAll(() => {
...
-      Promise.resolve({
-        json: () => Promise.resolve(placesAndScreens),
-      })
-    ) as jest.Mock;
-  });
-
-  afterEach(() => {
-    global.fetch = originalFetch;
-  });
-
```
PaulJKim, line 29: So I did remove this fetching logic since now the Places array is passed as a prop to the component that this file is for. I'm not sure if that move makes sense so let me know if we should keep it.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -1,11 +1,130 @@
...
+          case "Ferry":
+            return informedEntity.route_type === 4;
+          default:
+            return ids.includes(informedEntity.route);
+        }
+      });
+    });
+  };
+
+  const filterAlertsByScreenType = (alerts: Alert[]) => {};
```
PaulJKim, line 96: This is on hold until we know what the data will look like once we're able to map alerts to screen types.

assets/tests/components/placesPage.test.tsx
hannahpurcell: I'm not sure, but I'm fine with this. The when and where will the fetch be removed (I think it's still there in master)

hannahpurcell (commented):
Looks great, with only a few tweaks. Thanks for the jest tests!

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -1,11 +1,133 @@
...
+    MODES_AND_LINES[0]
+  );
+  const [alertScreenTypeFilterValue, setAlertScreenTypeFilterValue] = useState(
+    SCREEN_TYPES[0]
+  );
+  const [alertStatusFilterValue, setAlertStatusFilterValue] = useState(
+    STATUSES[0]
+  );
+
+  const alertSortLabel = getAlertSortLabel(alertSortDirection);
```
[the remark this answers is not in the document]
PaulJKim: Works for me!

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -14,7 +136,57 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
...
-        <div style={{ color: "white" }}>This will be posted alerts</div>
+        <Container fluid>
+          <Row className="place-list__header-row">
+            <Col lg={3}>
+              <div
+                className="place-list__sort-label d-flex align-items-center"
+                onClick={sortLabelOnClick}
+                data-testid="sort-label"
+              >
+                {alertSortLabel}{" "}
```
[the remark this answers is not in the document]
PaulJKim: Oh, nope that must've snuck in through auto-formatting or something! Deleting

assets/tests/components/placesPage.test.tsx
PaulJKim: So, I _think_ the mock fetch needed to be there because it was called each time the dashboard component was rendered and that's how the places list was populated. Now that the separate Places and Alerts pages are basically just viewer components that take in their respective lists of things as props, the fetch doesn't run each time its re-rendered.

Now the Dashboard component is a lot thinner and doesn't do anything besides calling the backend or API for data, which we can't really unit test. So until there's some unit-testable logic that we add into the Dashboard component in a subsequent PR, I want to say that we can leave this mock fetching stuff out.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -14,7 +126,67 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
...
+              />
+            </Col>
+          </Row>
+        </Container>
+        {filterAlerts()
+          .sort((a: Alert, b: Alert) => {
+            return alertSortDirection === 0
+              ? compareAlerts(a, b)
+              : compareAlerts(b, a);
+          })
```
hannahpurcell, line 173: Ah, I realize the original commit I made was not correct because it didn't include def of `a` and `b`. Can still consolidate this, though
```suggestion
          .sort((a: Alert, b: Alert) =>
            alertSortDirection === 0 ? compareAlerts(a, b) : compareAlerts(b, a)
          )
```

hannahpurcell (approved):
Great! A JS syntax tweak that is unneeded, but just fyi

hannahpurcell:
Also, reminder to NOT merge this, though it may be tempting

PaulJKim:
@hannahpurcell reviving this one! I just added a commit that adds logic for filtering based on screen type which leverages (dummy) screens-by-alert data. I assume the logic will need to be revisited and tweaked once Jon's solution is fully implemented

hannahpurcell (approved):
Great, works for me! Could you sync up with @cmaddox5 before merging, because I know there was some overlap in representing alerts

PaulJKim:
Discussed with @cmaddox5 in a huddle and our changes only slightly overlap in the `alert.ts` model, but any merge conflicts should be pretty straightforward to resolve. Merging this PR into the `alerts-feature` branch!

PR #131, opened 2022-09-20
[Screenplay Polish] - Missing subtle dimension on rows

PaulJKim:
[Task in Notion](https://www.notion.so/mbta-downtown-crossing/Missing-subtle-dimension-on-rows-7e189e6ff11447648f1075044e3bacfa)

Evidently `border-radius`, `border-image-source`, and `linear-gradient` don't play too nicely together. The current solution works but is missing the gradient on the sides and only shows it along the top of a row.

Screen Shot 2022-09-20 at 4 43 13 PM

### DIFF
--- a/assets/css/place-row.scss
+++ b/assets/css/place-row.scss
@@ -108,7 +108,3 @@
   margin: 0 12px 0 12px;
   border-radius: 8px;
 }
-
-.place-row::after {
-  border-radius: 8px;
-}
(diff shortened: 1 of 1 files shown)


assets/css/place-row.scss
```diff
@@ -1,7 +1,14 @@
 .place-row {
+  border: 1px transparent;
```
hannahpurcell, line 2: Doesn't look like this line is actually impacting the row styling
```suggestion
```

hannahpurcell (approved):
👍

PR #132, opened 2022-09-21
polish: button styling

cmaddox5:
[Notion](https://www.notion.so/mbta-downtown-crossing/Text-styling-on-buttons-weight-and-character-spacing-527faaaf40c24d5ab6f37bbd6d802a02)

Primary focus of this ticket was text styling, so I added some explicit font-related styles. I also noticed that our buttons were not colored correctly. The background color and hover effect needed a slight adjustment for the default button.

Screen Shot 2022-09-21 at 1 49 33 PM


hannahpurcell (commented):
The main to-do for this polish task is to reduce the font-weight to 400 and condense the letter spacing to .5. The implementation screenshot still looks pretty thicc and spaced out compared to the design.

On the colors on hover: the button needs to be brightly colored when a filter is selected. So, unfortunately, the opaque vs bright version will all need to be handled in javascript. There is a separate polish task for this, so it's ok that the color isn't working quite right in this branch.

cmaddox5:
> The main to-do for this polish task is to reduce the font-weight to 400 and condense the letter spacing to .5. The implementation screenshot still looks pretty thicc and spaced out compared to the design.
> 
> On the colors on hover: the button needs to be brightly colored when a filter is selected. So, unfortunately, the opaque vs bright version will all need to be handled in javascript. There is a separate polish task for this, so it's ok that the color isn't working quite right in this branch.

I totally missed that Notion doc reference in the Notion doc for this task. My bad!

hannahpurcell (approved):
Beautiful! Thanks for also fixing the sizing of the x button. This does not totally finish the "hover" task, though, so that can happen in another branch.

PR #136, opened 2022-09-23
polish: Convert route pills from PNGs to SVGs

cmaddox5:
[Notion](https://www.notion.so/mbta-downtown-crossing/Replace-mode-pill-PNG-icons-2693e740a0084b29af39d69a974b18f3)

Pills definitely look better in this format. Another change in this is the removal of the Ferry pill. We have no need for it at the moment. Will need to revisit this to decide how we will show ferry on a place.


hannahpurcell (commented):
We'll need the ferry pill back 😬

cmaddox5:
> We'll need the ferry pill back 😬

Requested the SVG from Mindy. Will add it when I get it.

assets/css/place-row.scss
```diff
@@ -94,8 +94,14 @@
     }
   }
 
-  .place-row__mode-line-icon:not(:last-child) {
-    margin-right: 8px;
+  .place-row__mode-line-icon {
+    &--blue,
+    &--bus {
+      box-shadow: 0px 0px 2px 0px #2e3e4d66;
+    }
```
hannahpurcell, line 101: I'm not sure this is doing anything? I tried toggling it and see no changes.

hannahpurcell (approved):
Great!

assets/css/place-row.scss
cmaddox5: The drop shadow was originally added to the SVG itself, but I was seeing some issues with the pills that were not present in pills w/o drop shadow. I tried out the original SVGs with the drop shadow again just in case and they seem good to me. I must have made another change along the way that corrected the original problem. Do you mind pulling it up and checking that BL and Bus pills look good on your machine?

hannahpurcell: 👍 all route pills look good

PR #138, opened 2022-09-26
polish: Map segment updates

cmaddox5:
[Notion](https://www.notion.so/mbta-downtown-crossing/Replace-line-map-PNGs-and-adjust-z-index-781b3c5f063944a3b955ffccadf8cc70)

Map segments are now bundled SVGs. I was able to generalize the SVGs used to allow for reuse regardless of line/branch. Still requires a little bit of hardcoding, but uses way less assets.


hannahpurcell (changes requested):
Hmmmm looks like the "click to open dropdown" functionality has been put on the map segment only

assets/css/map-segment.scss
```diff
@@ -0,0 +1,11 @@
+.map-segment {
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translateY(-50%) translateX(-50%);
+  z-index: 999;
+
+  &--blue {
+    filter: drop-shadow(0px 0px 8px rgba(46, 62, 77, 0.4));
+  }
```
hannahpurcell, line 10: The stick color is also different for the blue line (unlike the other lines). Should be `#004BCC` while the route pill color is `#003DA5`.

Just out of curiosity, where does this drop-shadow live in the spec? I see the comment saying there should be a drop-shadow, but I can't find the values in the spec

hannahpurcell (commented):
Great! Just one fix, and the rest is gold

assets/css/map-segment.scss
cmaddox5: Yeesh, so subtly but I still shouldn't have missed that. Unfortunately, I think that means we can't reliably use that util function anymore. We can't change it because OFT uses it. I made a new function in the component itself. What do you think about that?

Also, the drop-shadow in design can be found [here](https://www.figma.com/file/AXmojXuKz4TRVhznOIMuBJ/Screenplay-Design?node-id=1149%3A54197).

assets/js/components/Dashboard/MapSegment.tsx
```diff
@@ -0,0 +1,74 @@
...
+      return "#7C878E";
+    case "green":
+    case "green-b":
+    case "green-c":
+    case "green-d":
+    case "green-e":
+      return "#00843D";
+    default:
+      return "#FFFFFF";
+  }
```
hannahpurcell, line 58: how about
```suggestion
    case "blue":
      return "#004BCC";
    default:
      color(line);
  }
```

hannahpurcell: and the util function would need to be imported again. This lets us get closer to "single source of truth" for those colors.

hannahpurcell (approved):

assets/js/components/Dashboard/MapSegment.tsx
cmaddox5: Very clean. Love that.

PR #140, opened 2022-09-27
Add special layout for Solari for screen cards

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/Solari-screens-should-stack-side-by-side-072a894f2d8440a4ae262f6b1fb04e4c)

If the screen type is Solari, we want to wrap all said Solari screens in a different layout component that lets the screen cards flex and fill a row. (Currently, we only ever expect to see <2 Solari screens, and the flex method looks good up to 3 screens.)

### DIFF
--- a/assets/js/components/Dashboard/ScreenDetail.tsx
+++ b/assets/js/components/Dashboard/ScreenDetail.tsx
@@ -13,28 +13,28 @@ interface ScreenDetailProps {
 const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
   const isSolari = props.screens.every((screen) => screen.type === "solari");
 
-  return (
-    isSolari
-    ? <div className="screen-detail__solari-layout">
+  return isSolari ? (
+    <div className="screen-detail__solari-layout">
       {props.screens.map((screens, index) => (
-        <ScreenCard
-          {...props}
-          key={index}
-          screens={[screens]}
-        />
+        <ScreenCard {...props} key={index} screens={[screens]} />
       ))}
     </div>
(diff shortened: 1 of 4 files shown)


assets/js/components/Dashboard/ScreenDetail.tsx
```diff
@@ -11,18 +11,30 @@ interface ScreenDetailProps {
...
-  };
+const ScreenCard = (props: ScreenDetailProps) => {
+  const { screens, isOpen } = props;
+  const isPaess = screens.every((screen) => screen.type === "pa_ess");
+  const isSolari = screens.every((screen) => screen.type === "solari");
+  const paessRouteLetter = screens[0].station_code
+    ? screens[0].station_code.charAt(0).toLowerCase()
+    : "";
+
+  const translateScreenType = SCREEN_TYPES.find(({ ids }) =>
```
cmaddox5, line 35: Super nitpicky, but the name of this still reads like it is a function. What do you think about `translatedScreenType`?

cmaddox5 (approved):
Looks great! Just have one optional suggestion.

assets/js/components/Dashboard/ScreenDetail.tsx
hannahpurcell: oh yeah, definitely worth mentioning! thanks

hannahpurcell, line 35:
```suggestion
  const translatedScreenType = SCREEN_TYPES.find(({ ids }) =>
```

PR #148, opened 2022-09-29
tweak: Add Ferry to stops in script

cmaddox5:
**Asana task**: ad-hoc

For Aquarium (and potentially others in the future), we need a way to add Ferry to the list of routes. The most straightforward way to do this is to add these special cases as patterns in a function. The `add_routes_to_stop` function here will be the function we use to do that.


jzimbel-mbta (approved):
Looks good, just a comment wording suggestion!

PR #150, opened 2022-09-29
feature: Add Sentry for frontend logs

cmaddox5:
**Asana task**: [Set Up Sentry for Screenplay](https://app.asana.com/0/1185117109217413/1201365885881926/f)

This sets up Sentry for our front-end for Screenplay. No source mapping yet, but at the very least we'll have a way to tell if errors are occurring. DSNs are already set in ECS.

### DIFF
--- a/assets/js/components/App.tsx
+++ b/assets/js/components/App.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement } from "react";
 import { Routes, Route, BrowserRouter } from "react-router-dom";
-import * as Sentry from "@sentry/react";
 
 const OutfrontTakeoverTool = React.lazy(
   () => import("./OutfrontTakeoverTool/OutfrontTakeoverTool")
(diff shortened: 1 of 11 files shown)


assets/js/components/App.tsx
```diff
@@ -30,6 +31,19 @@ class AppRoutes extends React.Component {
...
+
+  if (sentryDsn) {
+    Sentry.init({
+      dsn: sentryDsn,
+    });
+
+    if (username) {
+      Sentry.setUser({ username: username });
+    }
+  }
```
jzimbel-mbta, line 45: Slightly worried about putting this right in the render logic of a React component, since if it ever re-renders, these lines will run again. I don't think the top-level component will ever re-render, but still...

Unless this is exactly what the Sentry docs recommend as a best practice, would you mind doing one of the following?
1. Move this into a `useEffect` with `[]` as its dependency list, or
2. Call the Sentry initialization logic at the module level, in this module, right after the import statements.

jzimbel-mbta (approved):
Approved assuming you respond as you feel fit to the one comment!

PR #155, opened 2022-10-11
feat: Posted alerts backend

cmaddox5:
**Asana task**: [[Screenplay] Basic Alert Card list](https://app.asana.com/0/1185117109217413/1202932249366374/f)

This adds all of the backend logic we will need to add `AlertCard`s to the Posted Alerts page. A new endpoint has been added `/api/alerts` for retrieving all alerts from the API. No filtering has been done yet as that will come in a later task.


lib/screenplay/outfront_takeover_tool/alerts/reminders.ex
```diff
@@ -22,7 +22,7 @@ defmodule Screenplay.Alerts.Reminders do
     schedule_work()
 
     if url == "" do
-      Logger.info("No Slack Webhook URL found")
+      # Logger.info("No Slack Webhook URL found")
```
jzimbel-mbta, line 25: Any reason for commenting this out?

jzimbel-mbta (commented):
Just two things. Man, GitHub's diff viewer is pretty inconsistent about indicating when a file has simply been renamed/moved.

lib/screenplay/outfront_takeover_tool/alerts/reminders.ex
cmaddox5: Oops, that shouldn't have been committed. Log message was just getting in the way when debugging so I meant for that to be temporary.

lib/screenplay/alerts/alert.ex
```diff
@@ -1,204 +1,105 @@
...
-  @type update_map :: %{
-          optional(:message) => canned_message() | custom_message(),
-          optional(:stations) => list(station),
-          optional(:schedule) => schedule()
+  @moduledoc false
+
+  alias Screenplay.Alerts.Parser
+  alias Screenplay.V3Api
+
+  @derive Jason.Encoder
```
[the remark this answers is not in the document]
jzimbel-mbta: Right right—I see that in the current code, we explicitly convert the struct to a map before passing it off to `Jason` to convert to a JSON string.

But the `@derive Jason.Encoder` line in this module says "this struct can be directly converted to JSON using `Jason`'s provided functions". If we actually tried `Jason.encode(%Alert{...})`, however, it would fail when trying to use its default encoder functions.

If we're not planning to ever call `Jason.encode` directly on an Alert struct, I'd recommend deleting the `@derive Jason.Encoder` line.

jzimbel-mbta (approved):
🚀

PR #157, opened 2022-10-18
feat: Posted alerts list

cmaddox5:
**Asana task**: [[Screenplay] Basic Alert Card list](https://app.asana.com/0/1185117109217413/1202932249366374/f)

This PR adds the frontend code needed to display a list of `AlertCard`s on the `AlertsPage`. This includes all of the filter and sort work PaulK added in a previous PR. The code here communicates with V3 to get the alerts, so a new ENV var is needed (same `API_V3_KEY` used in screens). `devops` PR already merged and applied, but this will be the first time the key will be used. 

The route for the `AlertsPage` is now `/alerts` instead of `/dashboard/alerts`.


assets/css/alert-card.scss
```diff
@@ -0,0 +1,95 @@
+.alert-card {
```
hannahpurcell, line 1: This is appropriately named -- BUT there's another `alert-card` on the OFT side that is also adding styles.

For example, each alert card has a margin-top of 24px, but there should only be an 8px between cards.

assets/js/components/Dashboard/AlertCard.tsx
```diff
@@ -0,0 +1,124 @@
...
+        <Row>
+          <Col lg="auto">
+            {alert.affected_list.map((icon: string) => (
+              <img
+                className="alert-card__alert-details__pill"
+                key={`${alert.id}-${icon}`}
+                src={`/images/pills/${icon.toLowerCase()}.svg`}
+                alt={icon}
+              />
+            ))}
```
hannahpurcell, line 83: Multiple route pills will need to stack vertically

Screen Shot 2022-10-20 at 2 29 11 PM

hannahpurcell (commented):
Couple missed things, and otherwise very good. Thanks for ensuring total test coverage of AlertCard.tsx

assets/css/alert-card.scss
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    &__header {
+      margin-top: 16px;
+      font-family: Inter, sans-serif;
+      font-size: 14px;
+      font-weight: 500;
+      line-height: 24px;
+      letter-spacing: 0px;
+    }
```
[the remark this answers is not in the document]
cmaddox5: Yeah this just came directly from Figma, so I didn't think to do any refactoring. Happy to do the refactoring if you think that would be better.

lib/screenplay/outfront_takeover_tool/alerts/reminders.ex
```diff
@@ -22,7 +22,7 @@ defmodule Screenplay.OutfrontTakeoverTool.Alerts.Reminders do
     schedule_work()
 
     if url == "" do
-      Logger.info("No Slack Webhook URL found")
+      # Logger.info("No Slack Webhook URL found")
```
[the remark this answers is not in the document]
cmaddox5: Could've sworn I uncommented that. Was only there because that log was getting in the way while doing some debugging. My bad.

assets/css/alert-card.scss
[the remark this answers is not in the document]
hannahpurcell: Nope, in many cases concise code is important, but CSS is the exception (as is doing CSS efficiently). Refactoring wouldn't improve much

assets/css/alert-card.scss
cmaddox5: Fixed this with a bit of a band-aid. Changed the OFT `.alert-card` to `.oft-alert-card`. You think that's enough or should I find a better solution?

hannahpurcell: This whole file belongs to OFT, so you can nest all these styles (or at least just the alert-card style) in a selector for `outfront-container`, the wrapper for the whole OFT side of the app

cmaddox5: Good thinking. There are some CSS classes near the bottom of the file that don't directly live in the container so I left them out just in case. They should not conflict with anything in the Screenplay code though.

hannahpurcell: No, those directly live in the container, too. We'll have to build override previews eventually, so in case anything else gets named `pager` or `portrait-png`, let's wrap it all in the selector.

hannahpurcell (approved):
One more comment about the CSS, but going ahead and approving.

hannahpurcell (approved):
Sorry, I dismissed my review, but my branch was stale -- username is still present. There's some styling that's a bit off, but I'll fix it in my branch

PR #158, opened 2022-10-19
Alert places list

hannahpurcell:
**Asana task**: [[Screenplay] Alert Places List](https://app.asana.com/0/1185117109217413/1202932249366386)

Definitely a big PR, but a large part of it is rearranging components to make them more reusable, changing the test data, and writing tests. If it helps, review this PR commit by commit, because I tried to split it up into logical smaller units.

The PlacesPage was split into a PlacesPage and PlacesList so that the PlacesList could be reused on the AlertsPage.

~Last piece that hasn't been added: the Alert details at the top of the Alert Places List. The component was created in Christian's PR, so I'll wait to plug it in until that one is reviewed.~ done!


assets/js/components/OutfrontTakeoverTool/AlertDashboard/AlertDetails.tsx
```diff
@@ -65,7 +65,7 @@ const AlertDetails = (props: AlertDetailsProps): JSX.Element => {
   );
 
   return (
-    <div className="alert-card">
+    <div className="oft-alert-card">
```
cmaddox5, line 68: You might need to merge in changes from the feature branch one more time. This was a change from one of my commits that I ended up changing before merging.

assets/tests/components/alertsPage.test.tsx
```diff
@@ -32,14 +32,21 @@ describe("Alerts Page", () => {
...
         <AlertsPage
           places={placesAndScreens as Place[]}
           screensByAlertId={alertsOnScreens as ScreensByAlert}
           isVisible={true}
         />
       );
 
+      // First, verify expired alerts are not visible
+      // It's possible this filtering will get done in Christian's branch; check
+      // expect(queryByTestId("9")).not.toBeInTheDocument();
```
cmaddox5, line 45: I ran the test after uncommenting this and it passes. So you should be good on uncommenting and removing comments.

assets/tests/components/alertsPage.test.tsx
```diff
@@ -86,4 +136,27 @@ describe("Alerts Page", () => {
...
+    test("navigating to / from the places list for an alert", async () => {
+      const { getByText, getByTestId, queryByText } = render(
+        <AlertsPage
+          places={placesAndScreens as Place[]}
+          screensByAlertId={alertsOnScreens as ScreensByAlert}
+          isVisible={true}
+        />
+      );
+
+      await act(async () => {
```
cmaddox5, line 150: Not sure what the problem is, but I'm having trouble getting this test to pass. For some reason, the test doesn't think any alerts are in the document when it tries to click `10`.

config/runtime.exs
```diff
@@ -29,4 +29,5 @@ config :screenplay,
   slack_webhook_url: System.get_env("SLACK_WEBHOOK_URL"),
   environment_name: System.get_env("ENVIRONMENT_NAME"),
   sentry_frontend_dsn: System.get_env("SENTRY_DSN"),
-  api_v3_key: System.get_env("API_V3_KEY")
+  api_v3_key: System.get_env("API_V3_KEY"),
+  alerts_ui_url: System.get_env("ALERTS_UI_URL")
```
cmaddox5, line 33: Just a thought, but could the URLs live in this file (or some other config file) where they can just be conditionally chosen based on `environment_name`? I know there was a convo in Slack about the best place to put things like this, so this may push back on what was decided there. But `environment_name` is ultimately what we use to get the API URL, so we could do something similar with this. This would allow us to set `environment_name` and all other URLs will be picked correctly according to the value we use there.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -171,61 +265,58 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
...
-            alertSortDirection === 0 ? compareAlerts(a, b) : compareAlerts(b, a)
-          )
-          .map((alert: Alert) => (
-            <AlertCard key={alert.id} alert={alert} />
-          ))}
-      </div>
-    </div>
+    <>
+      <Container fluid>
+        <Row className="place-list__header-row">
```
cmaddox5, line 270: I don't think you set these class names, but it looks like they might be wrong due to some copy/pasting. These should be `alert-list...`.

cmaddox5 (changes requested):
Awesome! Love how all of the pieces are starting to fall into place! Just a few suggestions/questions before approving.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -37,6 +51,86 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
...
+        {selectedAlert ? (
+          <div>
+            <Button
+              className="back-button"
+              data-testid="places-list-back-button"
+              onClick={() => setSelectedAlert(null)}
+            >
+              <ArrowLeft /> Back
+            </Button>
+            <span>Service Change #{selectedAlert.id}</span>
```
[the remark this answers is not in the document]
hannahpurcell: Ohhhh yes! I misinterpreted the docs. Great, thanks

assets/js/components/OutfrontTakeoverTool/AlertDashboard/AlertDetails.tsx
hannahpurcell: I think I merged poorly. Hope nothing else got through 😓

assets/js/components/Dashboard/AlertsPage.tsx
hannahpurcell: Ah! https://github.com/mbta/screenplay/blame/alerts-feature/assets/js/components/Dashboard/AlertsPage.tsx#L182

cmaddox5 (approved):
Looks great! One last suggestion, but nothing big enough to hold back an approval.

PR #159, opened 2022-10-27
feat: Add place and screen count to AlertCard

cmaddox5:
**Asana task**: [[Screenplay] Include "places" and "screens" counts on alerts](https://app.asana.com/0/1185117109217413/1202932249366380/f)

`AlertCard`s now display the number of screens and places that are affected by the alert. 3 new ENV variables have also been added: `API_V3_URL`, `SCREENS_URL`, and `SIGNS_UI_URL`. These URLs  used to be picked conditional in code based on environment. They now live in terraform so each environment will automatically pick the correct URL. That PR is [here](https://github.com/mbta/devops/pull/675) and will need to be merged and applied before merging this PR.

### DIFF
--- a/assets/js/components/Dashboard/AlertsPage.tsx
+++ b/assets/js/components/Dashboard/AlertsPage.tsx
@@ -145,7 +145,6 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
   ) => {
     return alerts.filter((alert) => {
       // Get this alert's list of affected screens.
-      // Later should be replaced with a call to memcached?
       const screensWithAlert = screensByAlertMap[alert.id];
 
       return screensWithAlert
--- a/assets/js/components/Dashboard/AlertsPage.tsx
+++ b/assets/js/components/Dashboard/AlertsPage.tsx
@@ -178,6 +178,22 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
     return Date.parse(end1) - Date.parse(end2);
   };
 
+  const getNumberOfAffectedPlaces = (screensByAlert: string[]) => {
+    // Using a Set so we don't have to worry about counting a place more than once.
+    const placesCounted = new Set();
+
+    props.places.forEach((place) => {
+      if (
+        place.screens.filter((screen) => screensByAlert.includes(screen.id))
+          .length > 0
+      ) {
+        placesCounted.add(place.id);
+      }
+    });
+
+    return placesCounted.size;
+  };
+
   return (
     <div
       className={classNames("alerts-page", {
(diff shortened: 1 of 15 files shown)


assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -267,6 +278,22 @@ const AlertsList: ComponentType<AlertsListProps> = ({
...
+      if (
+        place.screens.filter((screen) => screensByAlert.includes(screen.id))
+          .length > 0
+      ) {
+        placesCounted.add(place.id);
+      }
+    });
+
+    return placesCounted.size;
+  };
```
PaulJKim, line 295: `getNumberOfAffectedPlaces` and `placesWithAffectedAlert` seem to have fairly similar logic. I know they're a little different (one returns a list of the actual places and is meant for a single selected alert while the other just returns the count of the affected places), but I'm wondering if there's a way to consolidate the logic for the sake of maintainability.

If it seems like more trouble than it's worth though, it's probably okay to leave it as is since I can't really foresee a business reason for either logic to change in the future.

lib/screenplay/alerts/screens_by_alert.ex
```diff
@@ -0,0 +1,48 @@
...
+      {:ok, parsed}
+    else
+      {:http_request, e} ->
+        {:error, httpoison_error} = e
+        log_api_error({:http_fetch_error, e}, message: Exception.message(httpoison_error))
+
+      {:response_success, %{status_code: status_code, body: body}} = response ->
+        _ = log_api_error({:bad_response_code, response}, status_code: status_code, body: body)
+
+        :bad_response_code
```
PaulJKim, line 24: What's the reason for returning this atom in this branch considering that we don't do anything similar in the other error cases? I think the end result would be the same - that is, a no match exception would be raised from the controller, but just curious.

lib/screenplay_web/controllers/alerts_api_controller.ex
```diff
@@ -2,9 +2,17 @@ defmodule ScreenplayWeb.AlertsApiController do
...
 
   alias Screenplay.Alerts.Alert
+  alias Screenplay.Alerts.ScreensByAlert
 
   def index(conn, _params) do
     {:ok, alerts} = Alert.fetch()
-    json(conn, Enum.map(alerts, &Alert.to_full_map/1))
+
+    {:ok, screens_by_alerts} =
+      alerts |> Enum.map(& &1.id) |> ScreensByAlert.get_screens_by_alert()
```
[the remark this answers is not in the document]
cmaddox5: A big advantage to having all caching logic live in the screens server is it gives us a super easy way to set everything up locally. If screenplay always communicated with memcache, it would have to look at data from dev-green or dev because local would not write to memcache. A very specific advantage to us devs but still helps a lot. 

And I think you are right about that downside being ok. Screenplay's goal is to represent the screens in a way that assumes all hardware is working as it should. Not the case for software. So if the screens app is down, the simulations won't work/counts won't show and PIOs will be able to see first hand that there is a problem on live screens.

assets/js/components/Dashboard/AlertsPage.tsx
cmaddox5: Good catch. I was avoiding that function at first because it looked at component state, but it's super easy to change it so it takes state as a variable instead.

lib/screenplay/alerts/screens_by_alert.ex
cmaddox5: There's no good reason for that. I changed all returns to `:error` in the `else` and added a default return in the controller.

assets/js/components/Dashboard/AlertsPage.tsx
PaulJKim: Ah nice. Glad it was an easy tweak!

PaulJKim (approved):
Seems good to me! Thanks for taking care of moving the URLs to TF!

PR #161, opened 2022-11-04
tweak: Metadata Module Plug

cmaddox5:
**Asana task**: [[Screenplay] 🐞 Username not initialized from alerts page](https://app.asana.com/0/1185117109217413/1203236363940422/f)

The solution for communicating ENV variables and similar constants was originally borrowed from the way the screens app. That method is fine for applications that only use a single page with no routing. For Screenplay, when the application loaded from the `/alerts` URL, some things like `username` and `environment_name` were not being picked up because that controller did not add them to `conn.assigns`. A new module plug has been added so we no longer have to remember to add values at the controller level. In this plug, it grabs all necessary values the same way the `DashboardController` was. Instead of adding these as data attributes on the `dashboard` template, they are now meta attributes on the main `app.html.heex`. This allows all of these values to be accessed from any page, regardless of the URL used to get there.

### DIFF
--- a/assets/js/app.tsx
+++ b/assets/js/app.tsx
@@ -20,8 +20,12 @@ import ReactDOM from "react-dom";
 import App from "./components/App";
 import * as Sentry from "@sentry/react";
 
-const sentryDsn = document.getElementById("app")?.dataset.sentry;
-const username = document.getElementById("app")?.dataset.username;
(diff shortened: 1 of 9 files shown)


hannahpurcell (approved):
Yeah, I think this is great

PR #164, opened 2022-11-14
tweak: Refactor navigation

cmaddox5:
**Asana task**: ad-hoc

In order to complete [this task](https://app.asana.com/0/1185117109217413/1203236363940419/f), the way we implement navigation needed to be refactored a bit. 

Before this change, we were doing some work manually that `react-router` normally takes care of. To fix this, I removed all custom conditional rendering on the `Dashboard` and instead added some additional routes to `App.tsx`. The top level `*` route renders the `Dashboard`. Under that are two nested routes for `PlacesPage` and `AlertsPage`. When using nested routes, you no longer reference components in the parent route's component. Instead, you just render `<Outlet />`. `<Outlet />` will be replaced by the child route's element when navigating between routes. 

In lieu of passing props to `AlertsPage` and `PlacesPage`, the `Outlet` now holds a context that holds the result of the `/api/dashboard` fetch (the `placesAndScreens` data). Each child route's element will be able to grab this data using `useOutletContext`. This change required some test changes. The `RenderRouteWithOutletContext` util component should be used as a wrapper for any child route so that the context can be initialized correctly.

### DIFF
--- a/assets/js/components/App.tsx
+++ b/assets/js/components/App.tsx
@@ -6,6 +6,8 @@ const OutfrontTakeoverTool = React.lazy(
   () => import("./OutfrontTakeoverTool/OutfrontTakeoverTool")
 );
 const Dashboard = React.lazy(() => import("./Dashboard/Dashboard"));
+const PlacesPage = React.lazy(() => import("./Dashboard/PlacesPage"));
+const AlertsPage = React.lazy(() => import("./Dashboard/AlertsPage"));
 const clarityTag = document
   .querySelector("meta[name=clarity-tag]")
   ?.getAttribute("content");
@@ -23,11 +25,10 @@ class AppRoutes extends React.Component {
             path="/emergency-takeover"
             element={<OutfrontTakeoverTool />}
           ></Route>
-          <Route
-            path="/dashboard"
-            element={<Dashboard page="places" />}
-          ></Route>
-          <Route path="/alerts" element={<Dashboard page="alerts" />}></Route>
+          <Route path="*" element={<Dashboard />}>
+            <Route path="dashboard" element={<PlacesPage />}></Route>
+            <Route path="alerts" element={<AlertsPage />}></Route>
+          </Route>
         </Routes>
       </React.Suspense>
     );
--- a/assets/js/components/Dashboard/AlertsPage.tsx
+++ b/assets/js/components/Dashboard/AlertsPage.tsx
@@ -29,6 +29,7 @@ import { PlacesList } from "./PlacesPage";
 import classNames from "classnames";
 import AlertCard from "./AlertCard";
 import { formatEffect } from "../../util";
+import { usePlaces } from "./Dashboard";
 
 type DirectionID = 0 | 1;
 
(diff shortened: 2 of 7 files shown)


assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -37,28 +38,21 @@ interface AlertsResponse {
...
-  isVisible: boolean;
-}
-
-const AlertsPage: ComponentType<Props> = (props: Props) => {
-  const { places, isVisible } = props;
+const AlertsPage: ComponentType = () => {
+  const { places } = useOutletContext<{ places: Place[] }>();
 
   const [alerts, setAlerts] = useState<Alert[]>([]);
   const [screensByAlertMap, setScreensByAlertMap] = useState<ScreensByAlert>(
```
jzimbel-mbta, line 45: We eventually plan to move this state up as well, right? (Just double checking)

assets/js/components/Dashboard/PlacesPage.tsx
```diff
@@ -30,23 +31,18 @@ const getSortLabel = (
...
-  <div
-    className={classNames("places-page", {
-      "places-page--hidden": !props.isVisible,
-    })}
-  >
-    <div className="page-content__header">Places</div>
-    <div className="page-content__body">
-      <PlacesList places={props.places} />
+  return (
+    <div className={classNames("places-page")}>
```
jzimbel-mbta, line 38:
```suggestion
    <div className="places-page">
```

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -81,11 +75,7 @@ const AlertsPage: ComponentType<Props> = (props: Props) => {
     ?.getAttribute("content");
 
   return (
-    <div
-      className={classNames("alerts-page", {
-        "alerts-page--hidden": !isVisible,
-      })}
-    >
+    <div className={classNames("alerts-page")}>
```
jzimbel-mbta, line 78:
```suggestion
    <div className="alerts-page">
```

jzimbel-mbta (approved):
Looks great! Just some small suggestions. Nice work updating the tests, as well.

assets/js/components/App.tsx
```diff
@@ -23,11 +25,10 @@ class AppRoutes extends React.Component {
...
             element={<OutfrontTakeoverTool />}
           ></Route>
-          <Route
-            path="/dashboard"
-            element={<Dashboard page="places" />}
-          ></Route>
-          <Route path="/alerts" element={<Dashboard page="alerts" />}></Route>
+          <Route path="*" element={<Dashboard />}>
+            <Route path="dashboard" element={<PlacesPage />}></Route>
+            <Route path="alerts" element={<AlertsPage />}></Route>
```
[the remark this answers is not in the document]
cmaddox5: I've been working on that today in a different branch and it's not easy to find a good place for the `Outlet` to live on the `AlertsPage`. I'm going to try to make it work, but it may be easier to make it its own `Route` to reduce the amount of conditional elements that would need to live in the `AlertsPage`.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -37,28 +38,21 @@ interface AlertsResponse {
...
 
-interface Props {
-  places: Place[];
-  isVisible: boolean;
-}
-
-const AlertsPage: ComponentType<Props> = (props: Props) => {
-  const { places, isVisible } = props;
+const AlertsPage: ComponentType = () => {
+  const { places } = useOutletContext<{ places: Place[] }>();
```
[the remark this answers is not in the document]
cmaddox5: I did try that first, and it does work great. Only issue was tests. Using a custom hook from `Dashboard` means that the `render` function in each test would need to reference `Dashboard`. My first instinct was that isn't ideal, but it may be ok for the reference to it to live in the util Component I made for routing. What do you think?

assets/js/components/Dashboard/AlertsPage.tsx
cmaddox5: Yep! I was just going to leave that out of scope for this particular task. It will be done in URL change part.

PR #165, opened 2022-11-16
tweak: Alert Detail URL change

cmaddox5:
**Asana task**: [Alert details / alert places list to be at new URL](https://app.asana.com/0/1185117109217413/1203236363940419/f)

Task above lays out the reason for this change very well.

Implementation for it was fairly straightforward after the refactor to navigation. Only thing that may look odd is this new URL is its own `Route` on the same level as `/alerts`. I could not find a good place to render the `<Outlet />` on the `AlertsPage` when I tried to add it as a nested route. It would require more conditional rendering, which we are trying to avoid. `AlertDetails` also needs all of the same context values present in the `Outlet` in `Dashboard`. If you feel strongly above making it a nested route, I can give that a shot and we can decide the best method.

This new URL works if you navigate straight to a specific alert. And if the ID in the URL does not exist, it takes you back to `/alerts` (with a new modal to be displayed in a future task).

### DIFF
--- a/assets/js/components/App.tsx
+++ b/assets/js/components/App.tsx
@@ -6,6 +6,8 @@ const OutfrontTakeoverTool = React.lazy(
   () => import("./OutfrontTakeoverTool/OutfrontTakeoverTool")
 );
 const Dashboard = React.lazy(() => import("./Dashboard/Dashboard"));
+const PlacesPage = React.lazy(() => import("./Dashboard/PlacesPage"));
+const AlertsPage = React.lazy(() => import("./Dashboard/AlertsPage"));
 const clarityTag = document
   .querySelector("meta[name=clarity-tag]")
   ?.getAttribute("content");
@@ -23,11 +25,10 @@ class AppRoutes extends React.Component {
             path="/emergency-takeover"
             element={<OutfrontTakeoverTool />}
           ></Route>
-          <Route
-            path="/dashboard"
-            element={<Dashboard page="places" />}
-          ></Route>
-          <Route path="/alerts" element={<Dashboard page="alerts" />}></Route>
+          <Route path="*" element={<Dashboard />}>
+            <Route path="dashboard" element={<PlacesPage />}></Route>
+            <Route path="alerts" element={<AlertsPage />}></Route>
+          </Route>
         </Routes>
       </React.Suspense>
     );
--- a/assets/js/components/App.tsx
+++ b/assets/js/components/App.tsx
@@ -8,6 +8,7 @@ const OutfrontTakeoverTool = React.lazy(
 const Dashboard = React.lazy(() => import("./Dashboard/Dashboard"));
 const PlacesPage = React.lazy(() => import("./Dashboard/PlacesPage"));
 const AlertsPage = React.lazy(() => import("./Dashboard/AlertsPage"));
+const AlertDetails = React.lazy(() => import("./Dashboard/AlertDetails"));
 const clarityTag = document
   .querySelector("meta[name=clarity-tag]")
   ?.getAttribute("content");
@@ -28,6 +29,7 @@ class AppRoutes extends React.Component {
           <Route path="*" element={<Dashboard />}>
             <Route path="dashboard" element={<PlacesPage />}></Route>
             <Route path="alerts" element={<AlertsPage />}></Route>
+            <Route path="alerts/:id" element={<AlertDetails />}></Route>
           </Route>
         </Routes>
       </React.Suspense>
--- a/assets/js/components/Dashboard/AlertDetails.tsx
+++ b/assets/js/components/Dashboard/AlertDetails.tsx
@@ -1,0 +1,72 @@
+import React, { ComponentType } from "react";
+import AlertCard from "./AlertCard";
+import { PlacesList } from "./PlacesPage";
+import { useNavigate, useOutletContext, useParams } from "react-router-dom";
+import { Alert } from "../../models/alert";
+import { Place } from "../../models/place";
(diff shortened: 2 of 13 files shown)


assets/js/components/App.tsx
```diff
@@ -8,6 +8,7 @@ const OutfrontTakeoverTool = React.lazy(
 const Dashboard = React.lazy(() => import("./Dashboard/Dashboard"));
 const PlacesPage = React.lazy(() => import("./Dashboard/PlacesPage"));
 const AlertsPage = React.lazy(() => import("./Dashboard/AlertsPage"));
+const AlertDetails = React.lazy(() => import("./Dashboard/AlertDetails"));
```
PaulJKim, line 11: I haven't seen this approach before. So is the aim to improve performance by not rendering all of the components at the initial page load and only render them when a use select to go to either page?

assets/js/components/App.tsx
```diff
@@ -28,6 +29,7 @@ class AppRoutes extends React.Component {
           <Route path="*" element={<Dashboard />}>
             <Route path="dashboard" element={<PlacesPage />}></Route>
             <Route path="alerts" element={<AlertsPage />}></Route>
+            <Route path="alerts/:id" element={<AlertDetails />}></Route>
```
PaulJKim, line 32: Just trying to understand React Router and outlet a bit. Is this bit of code here how the child components know where to go to fetch state when we call `useDashboardContext`?

assets/js/components/Dashboard/AlertDetails.tsx
```diff
@@ -0,0 +1,77 @@
...
+const AlertDetails: ComponentType = () => {
+  const { places, alerts, screensByAlertMap } = useDashboardContext();
+  const { id } = useParams();
+  const [selectedAlert, setSelectedAlert] = useState<Alert>();
+
+  const alertsUiUrl = document
+    .querySelector("meta[name=alerts-ui-url]")
+    ?.getAttribute("content");
+
+  const navigate = useNavigate();
```
PaulJKim, line 20: Does this just wrap `useNavigate()` in a variable? If so, what's the benefit we get from doing that?

lib/screenplay_web/router.ex
```diff
@@ -35,32 +35,32 @@ defmodule ScreenplayWeb.Router do
...
     get("/emergency-takeover", PageController, :index)
   end
 
   scope "/", ScreenplayWeb do
-    pipe_through [:redirect_prod_http, :browser, :auth, :ensure_auth]
+    pipe_through([:redirect_prod_http, :browser, :auth, :ensure_auth])
 
     get("/dashboard", DashboardController, :index)
-    get("/alerts", AlertsController, :index)
+    get("/alerts/*id", AlertsController, :index)
```
PaulJKim, line 58: Do we need this `*/id` in the route definition here? Actually I'm slightly confused because the index function in `AlertsController` doesn't seem to do anything 🤔 at least in this branch?

PaulJKim (commented):
In the Asana task it says
> Visiting a URL for an alert which exists in Screenplay takes the user directly to that alert's details / places page.

But when I try that on my local instance of Screenplay running off this branch, it actually just takes me back to the posted alerts page. Based on the routing, that seems like a bug of some sort or maybe it has something to do with how my local instance is set up?

assets/js/components/App.tsx
cmaddox5: This is actually the way we accomplished splitting the code between the new Screenplay and the old OFT. The bootstrap we use in Screenplay messes up the styling in OFT. If we lazy load the components, they only render when visited, so the CSS does not load and apply on OFT components. In the future, this will go away and we will render the components normally.

assets/js/components/App.tsx
cmaddox5: The state retrieved from `useDashboardContext` is defined in `Dashboard`. It is passed into the `<Outlet />` component. `Outlet` is basically a short way of rendering a child route's component on the parent route. So `Dashboard` is the parent route and when navigating to `/alerts`, it will replace `<Outlet />` with `<AlertsPage />`.

assets/js/components/Dashboard/AlertDetails.tsx
cmaddox5: That is the only way to consume the hook from a function within a function component. Hooks can only be called from within custom hook or at the top level of a function component. So here, I save the hook logic in a variable for later and that allows me to call it from a function within the component.

cmaddox5:
> In the Asana task it says
> 
> > Visiting a URL for an alert which exists in Screenplay takes the user directly to that alert's details / places page.
> 
> But when I try that on my local instance of Screenplay running off this branch, it actually just takes me back to the posted alerts page. Based on the routing, that seems like a bug of some sort or maybe it has something to do with how my local instance is set up?

Hmm, have you checked the ENV you are using for the API URL? If you are pointing API at `dev-green`, it will not have alert IDs from prod, and vice versa. If that all looks good, you're right there is likely a bug that needs to be addressed.

cmaddox5:
After loading this up again, the redirect is happening to me too. Looking into it now.

cmaddox5:
@PaulJKim Ended up being an easy fix. In the `useEffect` hook in `AlertDetails`, we need to wait for the list of alerts to populate before making a decision on if the page exists or not. In the logic for this, I messed up and looked at the array object instead of the length of the array. That is fixed now.

PaulJKim (approved):
After the small bugfix, looks like the behavior checks out with what's described in the Asana task. LGTM!

PR #166, opened 2022-11-18
bugfix: http to https redirect

hannahpurcell:
**Asana task**: [[Screenplay] Fix http / https routing](https://app.asana.com/0/1185117109217413/1203259339371911)

Setting the https port and setting force_ssl like so seems to resolve the redirect issue in a production test env. This follows part of the pattern suggested by the Phoenix boilerplate `prod.exs`, but some of the fields (keyfile / certfile) were not needed.

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -12,7 +12,6 @@ import Config
 config :screenplay, ScreenplayWeb.Endpoint,
   cache_static_manifest: "priv/static/cache_manifest.json",
   http: [:inet6, port: 4000],
-  redirect_http?: true,
   server: true,
   url: [port: 80]
 
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -62,7 +62,7 @@ unless System.get_env("PORT") do
   config :screenplay, SiteWeb.Endpoint, url: [scheme: "https", port: 443]
 
   config :screenplay, :secure_pipeline,
-    force_ssl: [host nil, rewrite_on: [:x_forwarded_proto]]
+    force_ssl: [host: nil, rewrite_on: [:x_forwarded_proto]]
 end
 
 # Check `Plug.SSL` for all available options in `force_ssl`.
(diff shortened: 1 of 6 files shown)


lib/screenplay_web/controllers/auth_controller.ex
```diff
@@ -37,9 +39,19 @@ defmodule ScreenplayWeb.AuthController do
...
 
   def callback(
-        conn = %{assigns: %{ueberauth_failure: %Ueberauth.Failure{}}},
+        conn = %{assigns: %{ueberauth_failure: %Ueberauth.Failure{errors: errors}}},
         _params
       ) do
+    error_messages =
+      errors
+      |> Enum.flat_map(fn
+        %Ueberauth.Failure.Error{message: message} when is_binary(message) -> [message]
```
jzimbel-mbta, line 48: Looks like the `message` field is _supposed_ to [always be a string](https://github.com/ueberauth/ueberauth/blob/master/lib/ueberauth/failure/error.ex#L11), but I suppose it doesn't hurt to make sure.

config/prod.exs
```diff
@@ -12,9 +12,10 @@ import Config
 config :screenplay, ScreenplayWeb.Endpoint,
   cache_static_manifest: "priv/static/cache_manifest.json",
   http: [:inet6, port: 4000],
-  redirect_http?: true,
   server: true,
-  url: [port: 80]
+  url: [port: 443],
```
jzimbel-mbta, line 16: It feels so odd that we (as in, computers) use different ports to indicate HTTP vs. HTTPS. I guess I'm still a little fuzzy on port mechanics, but after reading more on this it seems to make sense.

lib/screenplay_web/router.ex
```diff
@@ -51,30 +44,29 @@ defmodule ScreenplayWeb.Router do
...
   end
 
   scope "/auth", ScreenplayWeb do
-    pipe_through([:redirect_prod_http, :browser])
+    pipe_through([:browser])
 
     get("/:provider", AuthController, :request)
     get("/:provider/callback", AuthController, :callback)
-    get("/:provider/logout", AuthController, :logout)
+    # get("/:provider/logout", AuthController, :logout)
```
jzimbel-mbta, line 65: Remove commented code

jzimbel-mbta (approved):
Really only one actionable comment. Looks good to me!

PR #167, opened 2022-11-18
Updated station list for GLX

hannahpurcell:
Second part of [this task](https://app.asana.com/0/1185117109217413/1203364297898378)

Requires an update of the places_and_screens.json to see the updated stick map.


cmaddox5 (approved):
Looks good! Is a script re-run something that can be done now or do we need to wait until we are closer to launch?

hannahpurcell:
> *Just noticed a test is failing. I'll leave this approval but let me know if reapproval is needed after it is fixed.

nahhh just forgot to run prettier. thanks!

PR #168, opened 2022-11-18
feat: Posted/edited/closed alert banner

cmaddox5:
**Asana task**: [[Screenplay] Update Banner](https://app.asana.com/0/1185117109217413/1203279317437018/f)

Added a banner to the top of every page showing if an alert has been posted/edited/closed in the last two minutes. Only triggers for significant alerts (see `util.ts` for how we choose significant alerts).

### DIFF
--- a/assets/js/components/Dashboard/AlertBanner.tsx
+++ b/assets/js/components/Dashboard/AlertBanner.tsx
@@ -1,0 +1,48 @@
+import React, { ComponentType } from "react";
+import { Alert } from "../../models/alert";
+import { ArrowRepeat } from "react-bootstrap-icons";
+import {
+  convertArrayToListString,
+  formatEffect,
+  translateRouteID,
+} from "../../util";
+
+interface Props {
+  alert: Alert;
+}
+
+const AlertBanner: ComponentType<Props> = ({ alert }: Props) => {
+  const wasCreated = alert.created_at === alert.updated_at;
+  const getAffectedListString = () => {
+    const translatedAffectList = alert.affected_list.map((routeId) =>
+      translateRouteID(routeId)
+    );
+
+    return convertArrayToListString(translatedAffectList);
+  };
+  const getInfoSentence = () => {
+    if (wasCreated) {
+      return "Posting a new alert may cause others to appear on fewer screens.";
+    } else {
+      return "Some edits may cause other alerts to appear on different screens than before.";
+    }
+  };
+  return (
+    <div className="alert-banner">
+      <div className="alert-banner__icon-container">
+        <ArrowRepeat />
+      </div>
+      <div className="alert-banner__text">
+        <span className="bold">
+          A {getAffectedListString()} {formatEffect(alert.effect)} alert was
+          just {wasCreated ? "posted" : "edited"}.
+        </span>{" "}
+        {getInfoSentence()} It could take up to{" "}
+        <span className="bold">2 minutes</span> for any impacted screens to
+        update.{" "}
+      </div>
+    </div>
+  );
+};
+
+export default AlertBanner;
--- a/assets/js/components/Dashboard/AlertBanner.tsx
+++ b/assets/js/components/Dashboard/AlertBanner.tsx
@@ -39,11 +39,11 @@ const AlertBanner: ComponentType<Props> = ({ alert, isClosedAlert }: Props) => {
       <div className="alert-banner__text">
         <span className="bold">
           A {getAffectedListString()} {formatEffect(alert.effect)} alert was
-          just {wasCreated ? "posted" : "edited"}.
+          just {isClosedAlert ? "closed" : wasCreated ? "posted" : "edited"}.
         </span>{" "}
         {getInfoSentence()} It could take up to{" "}
         <span className="bold">2 minutes</span> for any impacted screens to
-        update.{" "}
+        update.
       </div>
     </div>
   );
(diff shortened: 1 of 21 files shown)


assets/js/components/Dashboard/Dashboard.tsx
```diff
@@ -6,45 +6,134 @@ import { Alert } from "../../models/alert";
...
+          moment(postedOrEditedAlert.updated_at).isAfter(
+            moment(bannerAlert.closedAt)
+          )
+      )
+    ) {
+      setBannerAlert({ alert: postedOrEditedAlert });
+    }
+    // If no other alert is eligible to be the bannerAlert, expire the current bannerAlert if two minutes have passed since it started displaying.
+    else if (
+      moment(now).isSameOrAfter(moment(bannerAlert.closedAt).add(2, "minutes"))
```
PaulJKim, line 85: Would this check handle posted/edited alerts? Since those alerts wouldn't have a defined `closedAt` value according to the comment on lines 65-67

assets/js/components/Dashboard/AlertBanner.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { ComponentType } from "react";
+import { Alert } from "../../models/alert";
+import { ArrowRepeat } from "react-bootstrap-icons";
+import { formatEffect, translateRouteID } from "../../util";
+import { useParams } from "react-router-dom";
+
+interface Props {
+  alert: Alert;
+  isClosedAlert: boolean;
```
PaulJKim, line 9: Definitely a nit but does this need to be prop? We could just do the check on line 39 that we do on line 134 in Dashboard.tsx, I think

assets/js/util.ts
```diff
@@ -46,6 +46,65 @@ export const abbreviation = (line: string) => {
...
+      return "Orange Line";
+    case "blue":
+      return "Blue Line";
+    case "silver":
+      return "Silver Line";
+    case "green":
+    case "green-b":
+    case "green-c":
+    case "green-d":
+    case "green-e":
```
PaulJKim, line 65: hmm I'm not sure how much of this is specified in the spec, but would it be worth including strings for the different green branches? I think it's useful to distinguish when an alert is affecting just one branch on not other on the green line trunk, for example

assets/js/util.ts
```diff
@@ -46,6 +46,65 @@ export const abbreviation = (line: string) => {
...
+    case "bus":
+      return "Bus";
+    case "cr":
+      return "Commuter Rail";
+    default:
+      return "";
+  }
+};
+
+export const getModeFromAffectedList = (affectedList: string[]) => {
```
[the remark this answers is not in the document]
cmaddox5: Hmm, that's a good question. I think that normally alerts only affect one mode, but I'll need to check with product on that.

assets/js/components/Dashboard/Dashboard.tsx
cmaddox5: That scenario should not get through the first `else if` on line 68. If `bannerAlert` is `undefined` or `bannerAlert.closedAt` is `undefined` and there is no new closed alert, it will set the `bannerAlert` to the result of `getPostedOrEditedAlert`. That result can either be a posted/edited alert from the last two minutes or `undefined` if there is no recent alert. 

After not looking at this for a week, this conditional block is hard to follow, but I'm really not sure how to simplify it 😬 . If you have any suggestions, let me know.

assets/js/util.ts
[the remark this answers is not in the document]
cmaddox5: Had a talk with PaulH about this in Slack ([thread](https://mbta.slack.com/archives/C039ARUM48H/p1669821605654029)). An alert created for subway _could_ also affect bus (i.e. Station Closure), but not the other way around. And the current logic tries to match on subway first, so this logic works for our needs. As a test, PaulH created a Station Closure alert at Harvard and a banner was displayed for it.

assets/js/components/Dashboard/Dashboard.tsx
PaulJKim: Hmm yeah I've been looking at it and can't think of any ways to make any significant simplifications. One small suggestion that might help is putting the `moment` comparisons on Lines 74-77 in a separate helper function to help the conditional block read a bit better, but I think that's the best I can come up with at the `moment` 🤷 

One thing I'm a little confused on is how a posted/edited alert banner will be closed after two minutes. I see that for a closed alert, we're constantly checking if two minutes has passed since the `closedAt` time, but a posted or edited alert wouldn't have that field, right? Oh, or does that check implicitly happen in `getPostedOrEditedAlert` and then the banner alert would just get set as `undefined` on line 69?

cmaddox5: Yep, it happens in `getPostedOrEditedAlert`. For recently posted/edited alerts, we are able to check if `created_at` or `updated_at` is in the last two minutes. Because the alert itself has that data, we don't need to store it and constantly check against the time we first saw it. For a closed alert, those fields are unreliable and would be lost on the second refresh after the alert closed anyway.

assets/js/components/Dashboard/AlertBanner.tsx
cmaddox5: Good call. I updated the logic a bit so that the `AlertBanner` takes the whole `bannerAlert` as a prop.

PaulJKim (approved):
LGTM!

PR #171, opened 2022-12-01
tweak: Lift state up

cmaddox5:
**Asana task**: [[Screenplay] Lift up state for persistent filters](https://app.asana.com/0/1185117109217413/1203303529042768/f)

This PR adds a new `context` and `reducer` to maintain state that should persist. This includes anything that alerts the list of places/alerts and open panels on the `PlacesPage`. The only new functionality that should be added here is persistent filter values. All other functionality should work exactly as it did before.


assets/js/hooks/useScreenplayContext.tsx
```diff
@@ -0,0 +1,226 @@
...
+  placesPageDirectionId: DirectionID;
+  placesPage: PageProps & {
+    showScreenlessPlaces: boolean;
+    activeEventKeys: string[];
+  };
+  alertsPage: PageProps;
+  bannerAlert?: BannerAlert;
+}
+
+const reducer = (state: any, action: any) => {
```
PaulJKim, line 46: It seems like we always send a string for action. Should we type check it as a string?

assets/js/components/Dashboard/AlertBanner.tsx
```diff
@@ -2,15 +2,19 @@ import React, { ComponentType } from "react";
...
 import { useParams } from "react-router-dom";
-import { BannerAlert } from "./Dashboard";
+import { useScreenplayContext } from "../../hooks/useScreenplayContext";
+import { Alert } from "../../models/alert";
 
-interface Props {
-  bannerAlert: BannerAlert;
+interface BannerAlert {
+  alert: Alert;
+  closedAt?: string;
```
PaulJKim, line 10: I _think_ the `closedAt` prop was removed in the banner alert PR. Was it added back intentionally or does the branch just need an update from master?

assets/js/hooks/useScreenplayContext.tsx
```diff
@@ -0,0 +1,226 @@
...
+  switch (action.type) {
```
PaulJKim, line 47: Hmm I could see an alternative approach where we have reducers for each page. Mainly I think it would just remove the need for the if-else blocks and reduce the size of the reducer(s) but it might be a little cleaner looking. Not sure if that would work in practice without playing around with `context` a bit, but what I have in mind is have like a `dashboardReducer` with the SET_PLACES, SET_ALERTS, and SET_SCREENS_BY_ALERT actions and then have a `placesReducer` and `alertsReducer` with basically the same set of actions but export those to their respective components and remove the need to pass the action.page field.

Or are we only allowed to have one reducer for the entire context?

assets/js/components/Dashboard/AlertBanner.tsx
cmaddox5: That interface is now used to represent a `BannerAlert` object and not the props for `AlertBanner`. So the props are still gone, but it made more sense to have the `BannerAlert` interface exported from this file instead of `Dashboard`.

assets/js/hooks/useScreenplayContext.tsx
cmaddox5: I gave that a shot. It felt messy putting it together, but I think clarity may be a little better. It's all in one commit so if we don't like it, it's easy to change back. What do you think?

assets/js/hooks/useScreenplayContext.tsx
cmaddox5: I added a type for the reducer action so all potential dispatch types are documented.

assets/js/components/Dashboard/AlertBanner.tsx
PaulJKim: Ah I see that makes sense

PR #172, opened 2022-12-05
HTTP redirect bug, take two

hannahpurcell:
**Asana task**: [[Screenplay] Fix http / https routing](https://app.asana.com/0/1185117109217413/1203259339371911)

Chased so many red herrings and the solution? Was to put the config variable `redirect_http` into the correct config category (can't nest it under endpoint). I feel so so silly.


jzimbel-mbta (approved):
🥳

PR #173, opened 2022-12-07
feat: Data Update Animations

cmaddox5:
**Asana task**: [[Screenplay] Update data animation](https://app.asana.com/0/1185117109217413/1203279317437020/f)

This PR adds a brief animation to `AlertCard`s and `PlaceRow`s on `AlertDetails`. When an alert has been updated, the `AlertCard` will highlight for a couple of seconds. On the `AlertDetails` page, `PlaceRow`s will highlight as they appear on the list. No animation should occur on page load, only when the page is already loaded and data changes.

### DIFF
--- a/assets/js/components/Dashboard/AlertCard.tsx
+++ b/assets/js/components/Dashboard/AlertCard.tsx
@@ -1,6 +1,6 @@
 import moment from "moment";
-import React from "react";
-import { Container } from "react-bootstrap";
+import React, { useEffect, useRef, useState } from "react";
+import { Container, Fade } from "react-bootstrap";
 import { ChevronRight } from "react-bootstrap-icons";
 import { ActivePeriod, Alert } from "../../models/alert";
 import classNames from "classnames";
@@ -16,6 +16,23 @@ interface AlertCardProps {
 
 const AlertCard = (props: AlertCardProps): JSX.Element => {
   const { alert, numberOfPlaces, numberOfScreens } = props;
+  const [showAnimation, setShowAnimation] = useState(false);
+
+  const firstUpdate = useRef(true);
+  useEffect(() => {
+    if (firstUpdate.current) {
+      firstUpdate.current = false;
+      return;
+    }
+
+    setShowAnimation(true);
+    setTimeout(() => setShowAnimation(false), 2000);
+  }, [
+    alert.header,
+    JSON.stringify(alert.active_period),
+    numberOfPlaces,
+    numberOfScreens,
+  ]);
 
   const renderEffect = (effect: string, severity: string) => {
     const formattedEffect = formatEffect(effect);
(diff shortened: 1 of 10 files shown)


assets/js/hooks/useUpdateAnimation.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+    setShowAnimation(true);
+    const newTimer = setTimeout(() => {
+      setTimer(undefined);
+      setShowAnimation(false);
+    }, 2000);
+    setTimer(newTimer);
+
+    // Cancel timer is component unmounts.
+    return () => clearTimeout(newTimer);
```
jzimbel-mbta, line 24: Unsure if this is what we want. Here's a case I'm imagining:

1. `deps` change. -> We start a timer and save its ID in the `timer` state.
2. `deps` change again within 2 seconds. -> React runs the cleanup function immediately before running the effect again, clearing the timer before it triggers. The effect runs and sees that we still have a timer ID stored in the `timer` state, so it does nothing.
3. Now, we have a timer ID stuck in the `timer` state with no logic to ever remove it. Animations on this component stop happening for the remainder of the component's existence on the page, because every time the effect runs, it stops short at `if (timer || ...) {return;}`.

Could you double check whether the hook actually behaves as I described if 2 updates happen within 2 seconds? It's very tricky to visualize timing stuff, so I might be wrong.

If it _does_ happen, I think you will need to have the timer-clearing logic run only when the component unmounts, and not every time it re-renders. The Internet:tm: recommends splitting the logic into two effects. The first effect will be the one you already have, but without a cleanup callback. The second effect will do nothing and have no dependencies—it will simply return the cleanup callback to be run on unmount.

```js
useEffect(() => {
  const cleanup = () => clearTimeout(timer);
  return cleanup;
}, [] );
```

or more succinctly:
```js
useEffect(() => () => clearTimeout(timer), []);
```

> **When exactly does React clean up an effect?** React performs the cleanup when the component unmounts. However, as we learned earlier, effects run for every render and not just once. This is why React _also_ cleans up effects from the previous render before running the effects next time.

assets/js/hooks/usePrevious.tsx
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from "react";
+
+export function usePrevious<T>(value: T) {
```
jzimbel-mbta, line 3: If you want to do the usual arrow function style, this can be written like:
```ts
const usePrevious = <T,>(value: T) => {
  ...
};
```

(Note the comma, which lets the parser rule out the `<T>` as a JSX expression)

assets/js/hooks/usePrevious.tsx
```diff
@@ -0,0 +1,9 @@
...
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current as T;
```
jzimbel-mbta, line 8: I love/hate that this works

assets/js/components/Dashboard/AlertCard.tsx
```diff
@@ -66,58 +82,65 @@ const AlertCard = (props: AlertCardProps): JSX.Element => {
...
-            />
-          ))}
-        </div>
-        <div className="alert-card__alert-details__main-body">
-          <div className="alert-card__alert-details__effect">
-            {renderEffect(alert.effect, alert.severity_string)}
-          </div>
-          <div className="alert-card__alert-details__header">
-            {alert.header}
+    <div style={{ position: "relative" }}>
```
jzimbel-mbta, line 85: Could this be done in CSS rather than inline?

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -63,6 +64,7 @@ const AlertsList: ComponentType<AlertsListProps> = ({
   } = useAlertsPageContext();
   const dispatch = useAlertsPageDispatchContext();
   const navigate = useNavigate();
+  const prevAlertIds = usePrevious(alerts)?.map((alert) => alert.id);
```
jzimbel-mbta, line 67: I believe you can ensure that `prevAlertIds` is always an array by using the nullish-coalesce operator `??`:
```suggestion
  const prevAlertIds = usePrevious(alerts)?.map((alert) => alert.id) ?? [];
```

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -260,6 +262,11 @@ const AlertsList: ComponentType<AlertsListProps> = ({
             ).length;
           }
 
+          let showAnimationOnMount = false;
+          if (prevAlertIds?.length) {
+            showAnimationOnMount = !prevAlertIds.includes(alert.id);
+          }
```
jzimbel-mbta, line 268: If you make the above change to have `prevAlertIds` always be an array, this can be:
```suggestion
          const showAnimationOnMount = !prevAlertIds.includes(alert.id);
```

assets/js/components/Dashboard/PlaceRow.tsx
```diff
@@ -35,6 +38,7 @@ const PlaceRow = (props: PlaceRowProps): JSX.Element => {
   const rowOnClick = useAccordionButton(props.eventKey, () =>
     props.onClick(props.eventKey)
   );
+  const { showAnimation } = useUpdateAnimation([], null, props.showAnimation);
```
jzimbel-mbta, line 41: Neat that this can be tweaked to show the animation once on mount like this!

assets/js/components/Dashboard/PlacesPage.tsx
```diff
@@ -68,6 +71,7 @@ const PlacesList: ComponentType<PlacesListProps> = ({
     activeEventKeys,
   } = usePlacesPageContext();
   const dispatch = usePlacesPageDispatchContext();
+  const prevPlaceIds = usePrevious(places)?.map((place) => place.id);
```
jzimbel-mbta, line 74: Same here, you can nullish-coalesce this into an empty array if previous places is undefined.

assets/js/components/Dashboard/PlacesPage.tsx
```diff
@@ -289,17 +293,24 @@ const PlacesList: ComponentType<PlacesListProps> = ({
...
+              key={place.id}
+              place={place}
+              eventKey={place.id}
+              onClick={handleClickAccordion}
+              classNames={isFiltered || isAlertPlacesList ? "filtered" : ""}
+              filteredLine={isOnlyFilteredByRoute ? getFilteredLine() : null}
+              defaultSort={sortDirection === 0}
+              showAnimation={
+                showAnimationForNewPlaces &&
+                prevPlaceIds !== undefined &&
```
jzimbel-mbta, line 308: Can remove this check if `prevPlaceIds` is always an array.

jzimbel-mbta (commented):
Looks good! I have one concern about when the timer cleanup runs, mainly. Timing + effect stuff is always deceptively tricky, so thanks for bearing with me on the review.

assets/js/hooks/useUpdateAnimation.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const useUpdateAnimation = (
+  deps: any[],
+  prevValue: any,
+  showAnimationOnMount?: boolean
+) => {
+  const [showAnimation, setShowAnimation] = useState(false);
+  const [timer, setTimer] = useState<NodeJS.Timeout>();
+  useEffect(() => {
+    // Prevents animation when already showing and on page load.
+    if (timer || (!showAnimationOnMount && !prevValue)) {
```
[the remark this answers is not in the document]
cmaddox5: No, good call. `undefined` is what I was looking for here.

assets/js/hooks/useUpdateAnimation.tsx
cmaddox5: Would our decision to fetch new alerts every 4 seconds prevent multiple updates in 2 seconds? I think it would because the only `deps` that can trigger an animation come from the same `fetch` that only happens every 4 seconds. The Screenplay client doesn't directly manipulate any of the `deps` (except maybe the places and screens counts, but even those numbers are based on results from the `fetch`).

jzimbel-mbta: I think it would, but I'm sliiiightly worried about implicitly baking that assumption into the code. If you think the proposed fix would decrease readability or something, I'd be willing to let it go, but if at all possible I'd prefer the hook to work the same regardless of how frequently a component re-renders.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -260,6 +262,8 @@ const AlertsList: ComponentType<AlertsListProps> = ({
             ).length;
           }
 
+          const showAnimationOnMount = !prevAlertIds.includes(alert.id);
```
jzimbel-mbta, line 265: Ah! My bad. I think the change I proposed would cause different behavior from your original code.

Original
```ts
const previousAlerts = usePrevious(alerts);
// previousAlerts == undefined

const prevAlertIds = previousAlerts?.map((alert) => alert.id)
// prevAlertIds == undefined

let showAnimationOnMount = false;
if (prevAlertIds?.length /*  false */) {
  // does not run
  showAnimationOnMount = !prevAlertIds.includes(alert.id);
}

// showAnimationOnMount == false
```

New
```ts
const previousAlerts = usePrevious(alerts);
// previousAlerts == undefined

const prevAlertIds = previousAlerts?.map((alert) => alert.id) ?? [];
// prevAlertIds == []

const showAnimationOnMount = !prevAlertIds.includes(alert.id);
// showAnimationOnMount == true
```

Apologies, please revert!

cmaddox5: Good catch. I didn't notice that for some reason.

assets/js/hooks/useUpdateAnimation.tsx
cmaddox5: Makes total sense, and doesn't decrease readability. Just took me a minute to wrap my head around the problem. I think I have it. Needed to make a small tweak though. In a separate comment, I said that we only need to check for `prevValue !== undefined`, but that isn't exactly true. We need to also make sure it isn't `null`. With our strict typing, those would need to be two separate checks, or we could check for a falsy value.

jzimbel-mbta: I think that's one of the few times when the less-strict `!=` (or `==`) are recommended—`x != undefined`, or `x != null`, returns true when `x` is either `undefined` or `null`. But no worries, the separate checks are fine.

jzimbel-mbta (approved):
Good to go!

assets/js/hooks/useUpdateAnimation.tsx
cmaddox5: Somehow my brain forgot that you didn't have to use strict equality 😬 . That's a much better idea. Changed to that just in case `prevValue` is a boolean in the future.

PR #174, opened 2022-12-08
tweak: Report A Problem URL

cmaddox5:
**Asana task**: [[Screenplay] For PIOs, "report a problem" button goes to new private slack channel](https://app.asana.com/0/1185117109217413/1203365549546187/f)

PIOs now have a new private Slack channel to use when communicating with our team. This PR changes the Report a Problem button to redirect to it if the logged in user is an admin. The only Screenplay admins are PIOs (and maybe a couple of our team members) so it is safe to assume that an admin has access to the private channel. If you are not an admin, the button continues to redirect you to #screens.


hannahpurcell (approved):
Huh, I think metadata was extracted into a separate pipeline previously, so I guess we had a bad merge that brought it back? Regardless, this looks good now. And our previous testing ✅-ed functionality.

PR #178, opened 2022-12-20
polish: Alert banner timing

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/The-banner-should-only-persist-40sec-not-2min-616e47fa44334366b728d47f99082c50)

Now that we know that alerts will take no more than 40 seconds to start showing on screens, we no longer need the banner to be visible for 2 minutes.


hannahpurcell (approved):
👍

PR #179, opened 2022-12-21
polish: Use local state on AlertDetails

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Expanded-station-rows-should-not-persist-between-different-alert-details-pages-f89d161f1df64393bef4ded06473f087)

Because `AlertDetails` and `PlacesPage` both use `PlacesList`, we are inadvertently dispatching state change for `PlacesPage` from `AlertDetails`. `AlertDetails` state does not need to persist at all, so I refactored `PlacesList` to take in state related functions/values as props. This allows a component to control where state is stored (in the Screenplay context or locally). 

I tested this and everything works, but it feels very messy. Any suggestions for better ways to do this are encouraged.

### DIFF
--- a/assets/js/components/Dashboard/AlertDetails.tsx
+++ b/assets/js/components/Dashboard/AlertDetails.tsx
@@ -10,7 +10,15 @@ import {
   SlashCircleFill,
 } from "react-bootstrap-icons";
 import { formatEffect, placesWithSelectedAlert } from "../../util";
-import { useScreenplayContext } from "../../hooks/useScreenplayContext";
+import {
+  DirectionID,
+  useScreenplayContext,
+} from "../../hooks/useScreenplayContext";
+import {
+  MODES_AND_LINES,
+  SCREEN_TYPES,
+  STATUSES,
+} from "../../constants/constants";
 
 const AlertDetails: ComponentType = () => {
   const screenplayContext = useScreenplayContext();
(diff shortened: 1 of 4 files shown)


assets/js/components/Dashboard/AlertDetails.tsx
```diff
@@ -26,6 +34,44 @@ const AlertDetails: ComponentType = () => {
...
   const [showModal, setShowModal] = useState(false);
+  const [sortDirection, setSortDirection] = useState(0 as DirectionID);
+  const [modeLineFilterValue, setModeLineFilterValue] = useState(
+    MODES_AND_LINES[0]
+  );
+  const [screenTypeFilterValue, setScreenTypeFilterValue] = useState(
+    SCREEN_TYPES[0]
+  );
+  const [statusFilterValue, setStatusFilterValue] = useState(STATUSES[0]);
+  const [activeEventKeys, setActiveEventKeys] = useState([]);
```
jzimbel-mbta, line 45: I'm honestly not sure if this would be an improvement, but one option that would let you avoid duplicating all these separate state instantiations would be to create the context again, reusing one or more providers from `useScreenplayContext.tsx`—this new context would take priority over the existing one(s) provided higher up in the component tree.

https://reactjs.org/docs/hooks-reference.html#usecontext

> The current context value is determined by the `value` prop of the nearest `<MyContext.Provider>` above the calling component in the tree.

Then, the component that consumes the context values could continue doing so, not caring about whether they're the ones from the top-level provider, or the lower-level AlertDetails-specific one.

jzimbel-mbta (commented):
I took a look around in the docs, and it looks like you can simply define a new set of context values, which will take priority over the ones defined higher up.

Do you think that would be an improvement over the current changes? It feels a little "magical" / smells a little bit like inheritance to me, but it _would_ allow the original component to continue using `usePlacesPageDispatchContext` to get the appropriate dispatcher function.

assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -176,7 +172,7 @@ const AlertsList: ComponentType<AlertsListProps> = ({
 
       return screensWithAlert
         ? screensWithAlert.find((screen_id) =>
-            ids.includes(screenMetaData[screen_id].type)
+            ids.includes(screenMetaData[screen_id]?.type)
```
jzimbel-mbta, line 175: Was this in the original diff? Wondering why the `?` needed to be added.

assets/js/hooks/useScreenplayContext.tsx
```diff
@@ -255,9 +255,15 @@ const ScreenplayProvider = ({ children }: Props) => {
...
-  usePlacesPageContext,
-  usePlacesPageDispatchContext,
-  useAlertsPageContext,
-  useAlertsPageDispatchContext,
+  usePlacesListContext,
+  usePlacesListDispatchContext,
+  useAlertsListContext,
+  useAlertsListDispatchContext,
   ScreenplayProvider,
+  FilterValue,
```
jzimbel-mbta, line 263: I know all of these exports are necessary, but since you mentioned not liking how many there are, maybe it would help to split it into two `export` statements—one for values, one for types?

```tsx
// types
export {
  FilterValue,
  DirectionID,
  PlacesListReducerAction,
  ...
};

// values
export {
  usePlacesListContext,
  usePlacesListDispatchContext,
  ...
};
```

Imports in other modules could similarly be split into two lines, or not.

jzimbel-mbta (approved):
Looks great!! Just 2 small, optional Q's/suggestions.

assets/js/components/Dashboard/AlertsPage.tsx
cmaddox5: So I think this may have been a bandaid for the `screens_by_alert` bug that was fixed recently. Screens that were hidden from Screenplay were still logging alerts to the cache. When filtering by screen type, Screenplay would crash because of a null reference. Now that we aren't caching hidden screens, this is no longer necessary.

assets/js/hooks/useScreenplayContext.tsx
cmaddox5: Love this idea.

PR #182, opened 2022-12-22
polish: AlertCard animation

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Don-t-show-highlight-animation-when-alert-duration-changes-ef478a2f0045443092be5ee8cbb40b47)

When alerts are created with an end date of `Until later today`, the end date is changed by the API once a minute of two. This is causing animations to happen too frequently for those alerts. Because we have no way of determining if an alert was created using `Until later today`, the best thing to do is only consider start date in the animation hook.


jzimbel-mbta (approved):
What strange behavior on the API's part. The workaround seems reasonable, though.

PR #186, opened 2022-12-23
polish: AlertBanner visibility logic change

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Logic-for-when-banner-is-shown-4a02be5c16d94cb68137b689fdc7f0d4)

See table in task for new logic.

### DIFF
--- a/assets/js/util.ts
+++ b/assets/js/util.ts
@@ -84,7 +84,7 @@ export const getModeFromAffectedList = (affectedList: string[]) => {
       "green-c",
       "green-d",
       "green-e",
-    ].filter((value) => affectedList.includes(value))
+    ].filter((value) => affectedList.includes(value)).length
   ) {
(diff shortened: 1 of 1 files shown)


assets/js/util.ts
```diff
@@ -84,7 +84,7 @@ export const getModeFromAffectedList = (affectedList: string[]) => {
       "green-c",
       "green-d",
       "green-e",
-    ].filter((value) => affectedList.includes(value))
+    ].filter((value) => affectedList.includes(value)).length
```
hannahpurcell, line 87: A little more precise than filtering (since the filtered list isn't getting saved for later use)

```suggestion
    ].some(line => affectedList.includes(line))
```

hannahpurcell (approved):
Looks good to me! One suggestion

PR #187, opened 2022-12-23
polish: Sticky AlertBanner

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Banner-doesn-t-stick-when-page-is-scrolled-f93f13fd55124e9980d263dbd2ea840c)

Made the `AlertBanner` always stick to the top of the screen so it is visible after scrolling.

### DIFF
--- a/assets/css/alert-banner.scss
+++ b/assets/css/alert-banner.scss
@@ -6,7 +6,7 @@
   padding: 16px 221px 16px 50px;
   position: sticky;
   top: 0;
-  z-index: 2;
+  z-index: 999;
 
   &__icon-container {
     height: 24px;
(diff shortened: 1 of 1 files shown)


assets/css/alert-banner.scss
```diff
@@ -4,6 +4,9 @@
   background-color: #095292;
   align-items: center;
   padding: 16px 221px 16px 50px;
+  position: sticky;
+  top: 0;
+  z-index: 2;
```
jzimbel-mbta, line 9: Unless there's something else that might need to cover up this banner (which I doubt since it's sticky), you don't need to be shy about setting the z-index to something big, like `99`, `999`, something like that. It will ensure that nothing appears on top of the banner, and we can always adjust the number later if needed.

jzimbel-mbta (approved):
One optional suggestion, otherwise looks good!

assets/css/alert-banner.scss
cmaddox5: Ahh, good idea.

PR #188, opened 2022-12-23
polish: AlertsPage sorting

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Unexpected-end-date-sort-ordering-9ab2e5ec271140a3810e220290b15505)

The `AlertsPage` logic we had before wasn't quite right when comparing a recurring alert. Switching to `moment` makes comparison much easier with the help of the many helper functions.

### DIFF
--- a/assets/js/components/Dashboard/AlertsPage.tsx
+++ b/assets/js/components/Dashboard/AlertsPage.tsx
@@ -23,6 +23,7 @@ import {
   useScreenplayContext,
 } from "../../hooks/useScreenplayContext";
 import { usePrevious } from "../../hooks/usePrevious";
+import moment from "moment";
 
 const AlertsPage: ComponentType = () => {
   const { places, alerts, screensByAlertMap } = useScreenplayContext();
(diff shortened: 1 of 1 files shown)


assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -186,21 +187,22 @@ const AlertsList: ComponentType<AlertsListProps> = ({
...
     // Get the soonest start time
-    const { start: start1 } = active_period_1[0];
-    const { start: start2 } = active_period_2[0];
+    const start1 = moment(active_period_1[0].start);
+    const start2 = moment(active_period_2[0].start);
     // Get the latest end time
-    const { end: end1 } = active_period_1[active_period_1.length - 1];
-    const { end: end2 } = active_period_2[active_period_2.length - 1];
+    const end1 = moment(active_period_1[0].end);
+    const end2 = moment(active_period_2[0].end);
```
PaulJKim, line 194: Hmm so this means that we'd end up comparing against the soonest active period for a recurring alert instead of the last active period right? Is that that desired behavior?

cmaddox5: I believe so because we only show the soonest active period on the `AlertCard`. Sorting by the last would potentially cause a recurring alert that is currently active to show under upcoming alerts. Luckily the API tosses active periods after they are over, so we don't have to worry about past dates showing.

PaulJKim: Ohh okay I think that makes sense. Yeah if we're showing only the soonest active period on the Alert Card, then it would be odd to sort by end date and have the recurring alert jump further down the list given how it was being compared based on end of the _very_ last active period.

cmaddox5: Ideally, we would show what the recurring schedule is on the `AlertCard`. We just unfortunately don't have the data to do that without making some assumptions. I think the original designs had that but we had to pivot given the data we receive.

PaulJKim: But just to be sure, I think this is a departure from what was discussed in [this old thread](https://mbta.slack.com/archives/C039ARUM48H/p1663261979867459?thread_ts=1663261833.303339&cid=C039ARUM48H). Does that check out with the aim of the PR?

cmaddox5: Interesting! Great find. With PaulH being the one to tell you yes to your question, I think we definitely wait on this. I don't want to accidentally make any product changes off a hunch.

PaulJKim: Sounds good to me! Yeah, I could see both approaches making sense from a UX perspective so would be curious to see what Paul and Mindy think

cmaddox5: Based on the Slack huddle earlier today, we should be good to go with this solution. There will need to be other changes to `AlertCard`s, but those will need to go through design first.

PaulJKim (approved):

PR #190, opened 2022-12-29
Polish: Banner actively counts down based on latest alert change

hannahpurcell:
**Notion [task](https://www.notion.so/mbta-downtown-crossing/Screenplay-Alerts-Polish-d3f44e3eec4346a19f31f24f75fa1ae6?p=09b494315f4143f8b916673ae8eb2a31&pm=s)**

The banner should be visible and counting down until 40seconds after the alert change (closed/updated/posted). (It's likely that the countdown will never start at 40s because it takes several seconds for posts to appear in Screenplay.)

If a new change overrides the banner, the countdown is also updated.

### DIFF
--- a/assets/js/components/Dashboard/AlertBanner.tsx
+++ b/assets/js/components/Dashboard/AlertBanner.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentType } from "react";
+import React, { ComponentType, useEffect, useState } from "react";
 import { ArrowRepeat } from "react-bootstrap-icons";
 import { formatEffect, translateRouteID } from "../../util";
 import { useParams } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


assets/js/components/Dashboard/AlertBanner.tsx
```diff
@@ -54,7 +82,8 @@ const AlertBanner: ComponentType = () => {
               alert was just closed.`}
             </span>{" "}
             Closing an alert may cause others to appear on more screens. It
-            could take up to 40 seconds for any impacted screens to update.
+            could take up to {timeLeft} seconds for any impacted screens to
```
cmaddox5, line 85: Super nit-picky, but should there be logic here to drop the "s" when `timeLeft === 1`?

cmaddox5 (approved):
Looks great! Just one question.

PR #195, opened 2023-01-10
Alert Details Page: filter out screens that don't have alert

hannahpurcell:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Screenplay-Alerts-Polish-d3f44e3eec4346a19f31f24f75fa1ae6?p=0c3aa543241f4961a3729502fa5636a4&pm=s).

On the alert details page, filter out screens that don't show the alert. This automatically updates the screens count for the page.


jzimbel-mbta (approved):
Looks good to me! Just need to run your changes through Prettier.

The filter function also lost its `screen: Screen` argument type annotation, which is making eslint complain. Up to you if you want to put that back, or just remove the now-unused import.

PR #197, opened 2023-01-18
polish: systemwide alert icons

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Systemwide-alert-shows-up-with-an-Access-pill-eb5a5852e0d74c318f6d17b86d9f9a75)

For systemwide alerts, we were incorrectly showing the access icon. This is because it was assumed that any alert with no `relationships` was an access alert of some kind. Systemwide alerts also have no `relationships`. I added a function pattern that matches `ELEVATOR_CLOSURE` to `:access` and fallbacks to `:alert` for any other effect with no `relationships`.

This does cause other access alerts to display the `:alert` icon, but with this [task](https://www.notion.so/mbta-downtown-crossing/Filter-out-alerts-that-would-never-show-up-on-screens-369ad44cc004411b8d9a173ae80b07ff), I _think_ that's ok. Elevator alerts are the only alerts we show on screens, and I don't foresee us needing others any time soon. Feel free to push back on this though.


jzimbel-mbta (approved):
👍

PR #198, opened 2023-01-18
polish: add favicon

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Add-favicon-for-easier-navigation-between-tabs-a671059e66c04817935ee38036802845)

Added a favicon for browser tabs. I have found that this is a cached image so if you want to test change during review, clear cache or use incognito.

Screenshot 2023-01-18 at 8 39 27 AM


hannahpurcell (approved):
Mindy later realized the icon was grainier than she'd prefer, but we're keeping it for now

PR #199, opened 2023-01-18
feat: Replace MS Clarity with Fullstory

cmaddox5:
**Asana task**: [[Screenplay] Set up Fullstory](https://app.asana.com/0/1185117109217413/1203570469014618/f)

Fullstory has a npm package we can use, so setup is much easier than Clarity was. It does require terraform changes (see https://github.com/mbta/devops/pull/758). Also, the `!FullStory.isInitialized()` in the conditional may be overkill but I thought better safe than sorry. 

Once this is merged in, we can then go into terraform and remove `CLARITY_TAG` from the screenplay modules.


jzimbel-mbta (approved):
🙅🪟

🙆📖

PR #200, opened 2023-01-19
Polish: the alerts for banner & list are filtered on the backend

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/Screenplay-Alerts-Polish-d3f44e3eec4346a19f31f24f75fa1ae6?p=369ad44cc004411b8d9a173ae80b07ff&pm=s)

Both the banner and alert list should be filtered down by alert effect (only the effects that appear as alerts on screens) and whether it is happening now.

Additional changes: banner should not count down. Also, the banner should correctly identify SL alerts.

### DIFF
--- a/lib/screenplay/util.ex
+++ b/lib/screenplay/util.ex
@@ -10,4 +10,20 @@ defmodule Screenplay.Util do
     username
     |> String.replace("ActiveDirectory_MBTA\\", "")
   end
+
+  def happening_now?(%{active_period: aps}, now \\ DateTime.utc_now()) do
+    Enum.any?(aps, &in_active_period(&1, now))
+  end
+
+  def in_active_period({nil, end_t}, t) do
+    DateTime.compare(t, end_t) in [:lt, :eq]
+  end
+
+  def in_active_period({start_t, nil}, t) do
+    DateTime.compare(t, start_t) in [:gt, :eq]
+  end
+
+  def in_active_period({start_t, end_t}, t) do
+    DateTime.compare(t, start_t) in [:gt, :eq] && DateTime.compare(t, end_t) in [:lt, :eq]
+  end
 end
--- a/lib/screenplay_web/controllers/alerts_api_controller.ex
+++ b/lib/screenplay_web/controllers/alerts_api_controller.ex
@@ -48,7 +48,7 @@ defmodule ScreenplayWeb.AlertsApiController do
     })
   end
 
-  @spec is_significant_alert?(Alert) :: boolean()
+  @spec is_significant_alert?(Alert.t()) :: boolean()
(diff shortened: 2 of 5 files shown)


lib/screenplay/util.ex
```diff
@@ -10,4 +10,20 @@ defmodule Screenplay.Util do
...
+    DateTime.compare(t, end_t) in [:lt, :eq]
+  end
+
+  def in_active_period({start_t, nil}, t) do
+    DateTime.compare(t, start_t) in [:gt, :eq]
+  end
+
+  def in_active_period({start_t, end_t}, t) do
+    DateTime.compare(t, start_t) in [:gt, :eq] && DateTime.compare(t, end_t) in [:lt, :eq]
+  end
```
cmaddox5, line 28: The logic here looks good, but I think adding these to `alert.ex` makes a little more sense.

lib/screenplay_web/controllers/alerts_api_controller.ex
```diff
@@ -3,21 +3,71 @@ defmodule ScreenplayWeb.AlertsApiController do
...
+  @spec is_significant_alert?(Alert.t()) :: boolean()
+  def is_significant_alert?(alert = %{affected_list: affected_list}) do
+    mode = primary_affected_mode(affected_list)
+
+    cond do
+      is_nil(mode) -> false
+      alert.effect === "DELAY" and mode === :subway -> alert.severity > 3
+      true -> alert.effect in @significant_alert_effects[mode]
+    end
+  end
```
cmaddox5, line 60: Same here. Because this always takes an `Alert.t()`, can it just live in `alert.ex`?

lib/screenplay_web/controllers/alerts_api_controller.ex
```diff
@@ -3,21 +3,71 @@ defmodule ScreenplayWeb.AlertsApiController do
...
+
+  defp primary_affected_mode(affected_list) do
```
cmaddox5, line 62: Similarly, this could be useful to code outside of this controller down the road. I'm thinking `util.ex` for this though.

cmaddox5 (approved):
Code looks good! Just have some suggestions on where the code lives.

lib/screenplay_web/controllers/alerts_api_controller.ex
hannahpurcell: Is `affected_list` ever a thing outside of alerts? I'm thinking it's mainly an alert thing, so might put it in `alert.ex`, too

hannahpurcell: Great call on this re-organization, btw. Mine was all over the place

PR #202, opened 2023-01-24
Polish: Alert not found page

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/Navigating-to-a-non-existent-alert-detail-page-abc932a5310b456d902a706ffa46d009?d=0eb8f9812ad44d1eafb7dccb89a0cf08#4efecff8caf44252bd5db1b55b036c63)

The Alert Details page is doing a lot now! 

If the selected alert is active and relevant to Screenplay -> show the details page
If the selected alert was active but is now closed -> show a modal saying that the alert was closed, overlaying the details page
If the selected alert is active but not relevant to Screenplay -> show an "Alert not found" page suggesting edits to the alert
If the selected alert is non-existent -> show an "Alert not found" page saying "sorry bud, that's probably a typo"


assets/js/components/Dashboard/Dashboard.tsx
```diff
@@ -17,11 +17,12 @@ const Dashboard: ComponentType = () => {
   const [bannerDone, setBannerDone] = useState(false);
 
   useEffect(() => {
-    fetchAlerts((newAlerts, screens_by_alert) => {
+    fetchAlerts((all_alert_ids, newAlerts, screens_by_alert) => {
```
cmaddox5, line 20: Can you change `newAlerts` to snake case to match the other two parameters? I think I'm the one to introduce the inconsistency, so sorry about that 😬

assets/js/components/Dashboard/AlertNotFoundPage.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <div className="alert-not-found_title">Alert not found</div>
+        <p className="alert-not-found_detail">
+          {props.validAlertId
+            ? `We couldn’t find the alert you’re looking for in Screenplay. This alert \
+            (ID #${props.validAlertId}) is open, but it won’t show up on any screens due to \
+            its Service Type, Effect Type, or Effect Period.`
+            : `We couldn’t find the alert you’re looking for in Screenplay. This could be \
+            because the alert has been closed, or because an alert with that ID \
+            (#${
+              window.location.pathname.split("/").slice(-1)[0]
```
cmaddox5, line 22: react-router provides a hook that can be used here instead of looking at `window`: `useLocation`. The basic string splitting logic would still be the same so you wouldn't gain much, so just a suggestion.

assets/js/components/Dashboard/AlertDetails.tsx
```diff
@@ -50,6 +57,10 @@ const AlertDetails: ComponentType = () => {
     initialPlacesListState
   );
 
+  const validAlertId = contextState.allAPIAlertIds.length
```
cmaddox5, line 60: You could change this to a one-liner (well kinda, formatting may make it two lines).

`const validAlertId = contextState.allAPIAlertIds.find((alert) => alert !== undefined && alert === id);`

cmaddox5 (approved):
Looks good! Just a couple of suggestions.

assets/js/components/Dashboard/AlertDetails.tsx
hannahpurcell: Ope! I didn't realize `find` would work on an empty array and just return `undefined`. In that case, can condense even further: 
`const validAlertId = contextState.allAPIAlertIds.find(alert => alert === id);`

assets/js/components/Dashboard/Dashboard.tsx
hannahpurcell: Actually going to use camelCase because it's a tsx file. Thanks for the catch!

PR #203, opened 2023-01-26
polish: group alerts by screen count

cmaddox5:
**Asana task**: [ad-hoc-ish](https://mbta.slack.com/archives/C02SXM166N8/p1674761551378159)

For the `AlertsPage`, we now group alerts in two "sections": one has alerts on screens, the other has alerts not on screens. Within the sections, we are still sorting alerts by end date. With this change, we also are removing custom sorting on this page. For now, alerts will always be sorted in the specific way described above.


hannahpurcell (approved):
Looks clean to me!

PR #206, opened 2023-02-02
doc: Add required env vars to README; remove unused config

jzimbel-mbta:
**Asana task**: ad hoc

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,9 @@ To start your Phoenix server:
 - Set required env variables with
   - `export SECRET_KEY_BASE=`. You can use the value in `config.exs`.
   - `export GUARDIAN_SECRET_KEY=test_auth_secret`
+  - URLs for other MBTA APIs that Screenplay communicates with. **Make sure to use the same environment for all!**
+    - `export API_V3_URL="https://api-v3.mbta.com/"` (or another environment's URL if you want to fetch data from there instead)
+    - `export SCREENS_URL="https://screens.mbta.com/"` (or another environment's URL if you want to fetch data from there instead)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -9,6 +9,9 @@ To start your Phoenix server:
 - Set required env variables with
   - `export SECRET_KEY_BASE=`. You can use the value in `config.exs`.
   - `export GUARDIAN_SECRET_KEY=test_auth_secret`
+  - URLs for other MBTA APIs that Screenplay communicates with. **Make sure to use the same environment for all!**
+    - `export API_V3_URL="https://api-v3.mbta.com/"` (or another environment's URL if you want to fetch data from there instead)
+    - `export SCREENS_URL="https://screens.mbta.com/"` (or another environment's URL if you want to fetch data from there instead)
```
jzimbel-mbta, line 14: Not 100% sure about this. Do we want to hit the prod environments of these applications from local dev? I think we've always done that from Screens, but that certainly doesn't mean it's a best practice...

If Screenplay ever starts writing data to Screens, having local dev talk to prod by default could cause big problems.

hannahpurcell: Hm yeah. I think for now, while we are not writing to Screens, letting dev read from prod is fine. It's useful to let Screenplay look at prod screens from time to time (so the places/counts can stabilize on the alerts list, for example). When the time comes where we're writing data to Screens, we'll have to reconsider. Does that sound ok?

hannahpurcell (approved):
Thanks! Great catch on the need for doc updates

README.md
jzimbel-mbta: I think so! Just need to be very sure we change these values before adding any writing behavior.

PR #211, opened 2023-03-15
fix?: Fullstory init conditional

cmaddox5:
**Asana task**: ad-hoc

I initially added this conditional because it _seemed_ like a good thing to do. I think it might be causing Fullstory to never initialize. Unfortunately, this is not possible to test anywhere but prod.


hannahpurcell (approved):
/shrug! We can try it!

PR #220, opened 2023-08-01
RFC: Selecting an approach for sharing code between Screens and Screenplay

PaulJKim:
**Asana task**: [[Screenplay] Brainstorm: configuring screens](https://app.asana.com/0/1185117109217413/1205058023431637/f)

Description
This RFC is for outlining at the high-level an approach for allowing for Screenplay to reference code from Screens.

### DIFF
--- a/docs/tech_specs/configuration_backend_approach.md
+++ b/docs/tech_specs/configuration_backend_approach.md
@@ -105,3 +105,7 @@ The approach I recommend is number 2, the library approach. This approach requir
 While the git dependency would be the lowest hanging fruit in achieving our goal, we will lack the ability to maintain any boundaries between Screenplay and Screens which isn’t what we want since the majority of Screens code will not be needed by Screenplay.
 
 The library approach is a familiar and well supported method and achieves our needs in a way that lets us mark clear boundaries between the applications. While this approach will introduce some level of overhead in maintaining dependencies, it also gives us flexibility if we need it.
+
+# Selected Approach
+
+After discussion, we've decided to go with a mixture of two approaches. We will separate out the config code into its own Github repository but use the git dependency approach to import the code to Screens and Screenplay. This achieves the benefits of the library approach while allowing us to skip publishing a package to Hex.
--- a/docs/tech_specs/configuration_backend_approach.md
+++ b/docs/tech_specs/configuration_backend_approach.md
@@ -1,0 +1,107 @@
+- Feature Name: `select_backend_approach_for_screenplay_configuration`
+- Start Date: 2023-07-27
+- RFC PR
+- Asana task: [[Screenplay] Brainstorm: configuring screens](https://app.asana.com/0/1185117109217413/1205058023431637/f)
+- Status: Proposed
+
+# Summary/Motivation
+
+The Screens team is working to build out a new tool within Screenplay for managing screen configurations, effectively replacing the current Screens Admin Tool.
+
+The Screenplay Configuration Tool will need knowledge of the shape of the configuration for each screen type in order to be able to validate and write the configuration to the data store, currently a JSON file in S3. The challenge here is that the code that defines these shapes lives in and is used by the Screens application and now needs to also be referenced by Screenplay.
+
+The expected outcome of this document is to outline a few reasonable solutions for overcoming this hurdle and having the engineering team come to a consensus on an approach.
+
+Note: This document will not go into any lower-level implementation details for the Screenplay Configuration tool and will only focus on selecting an architectural approach for enabling Screenplay to reference code from Screens.
+
+# Possible approaches and rationale
+
+## 1. Umbrella Project
+
+### Description
+
+An Umbrella Project would allow us to group Screens and Screenplay in a single, unified codebase while still maintaining and deploying them as distinct applications. A dependency on Screens from Screenplay can then be defined and the code can be shared. Or the shared code can be separated into its own app within the umbrella and both applications can define a dependency on it.
+
+### High-level steps to get there
+
+- Create new umbrella project using `mix new` with `--umbrella` flag
+- Move Screens and Screenplay to `apps` folder in new umbrella project
+- Separate shared code into a new application within `apps` folder
+- Update `mix.exs` files to match umbrella structure and add dependency on new app with shared code to Screens and Screenplay's `mix.exs` files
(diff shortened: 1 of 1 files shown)


jzimbel-mbta:
(I assigned myself just so that I don't lose track of this PR—I don't mean to say I'm the primary reviewer. Others should feel free to do the same if they like.)

docs/tech_specs/configuration_backend_approach.md
```diff
@@ -0,0 +1,107 @@
...
+Elixir also supports adding a GitHub repository directly as a dependency to your project. You can also pin the dependency to specific tags/commits/branches as well.
+
+### High-level steps to get there
+
+- Add the git dependency to the deps in Screenplay’s `mix.exs` file
+- Fetch dependencies and start referencing Screens code
+
+### Benefits
+
+This method is the least intrusive and easiest as it requires a trivial amount of upfront work to enable Screenplay to reference the code from Screens. It also requires no changes to the Screens application since we don’t need to move the shared code anywhere. We also have some of the same flexibility as the library approach since we can pin Screenplay to a specific version of Screens.
```
jzimbel-mbta, line 91: I think we might run into the same problem here as I mentioned above—if the only way to access some Screens code is to get the entire application as a dependency, then we risk running into problems with having two different Phoenix web apps.

The git dependency approach could be useful for option 2 though—we could still split out the config code into a separate library, and skip the step of publishing it as an official package on hex, instead referencing the git repo in Screens/Screenplay deps.

jzimbel-mbta (approved):
Added some thoughts but LGTM and I agree with your recommendation.

I think we will need to be careful about where the boundary is for the new library, though. Since both apps will have code that expects the config to be structured in a very specific way, we should come up with a plan for how to handle changes (I think the common term is "migrations") in that structure before we start pulling the code into a library.

docs/tech_specs/configuration_backend_approach.md
```diff
@@ -0,0 +1,107 @@
...
+
+### References
+
+https://hexdocs.pm/mix/1.12.3/Mix.Tasks.Deps.html
+
+# Recommendation
+
+The approach I recommend is number 2, the library approach. This approach requires some upfront work to set up like the umbrella approach, but is less intrusive to our team’s workflow. Umbrella projects are more suited for cases where there are many, tightly integrated applications, which is not the case with Screens and Screenplay. In addition, setting up an umbrella project may backfire on us as it appears that it’s not a widely supported approach in the larger Elixir community and other teams within CTD have faced frustrations when using an umbrella, such as handy tools like Lcov being optimized for standard Elixir apps and not playing well with an umbrella setup.
+
+While the git dependency would be the lowest hanging fruit in achieving our goal, we will lack the ability to maintain any boundaries between Screenplay and Screens which isn’t what we want since the majority of Screens code will not be needed by Screenplay.
```
hannahpurcell, line 105: Great write-up! My first look-over, I didn't have strong opinions. All options were well discussed, and it was hard to choose. This time, though, the git option catches my eye as a particular fave.

Is keeping boundaries here critical? I don't see an obvious downside to allowing Screenplay to be able to access the full Screens repo. The developers of the two repos have the same permission levels, so it isn't a concern of security. And while much of the Screens repo will not be useful to Screenplay, whatever isn't used should cause performance issues or dead weight, I think?

hannahpurcell: Whoops, reading Jon's thoughts now!

hannahpurcell: After reading Jon's notes and [looking around](https://elixirforum.com/t/importing-a-phoenix-app-as-a-dependency-in-another-phoenix-app-in-an-umbrella-project-please-any-advice/33144/3), I see why the first and third option would cause Screenplay to always fully spin up Screens, which wouldn't be ideal. So I like Jon's suggestion on recommendation 2: isolate the shared code, make a library, make it a GH dependency instead of publishing to hex.

docs/tech_specs/configuration_backend_approach.md
jzimbel-mbta: One more thing supporting the git-based dep idea: the [getting started guide](https://elixir-lang.org/getting-started/mix-otp/dependencies-and-umbrella-projects.html#internal-dependencies) differentiates "external" and "internal" dependencies.

External deps are for domain-agnostic libraries and such that your application essentially uses as tools.

Internal deps are for things that are specific to your project and just need to be in a separate package for some reason—pretty much exactly our use case. The git-based approach is recommended for internal deps.

docs/tech_specs/configuration_backend_approach.md
PaulJKim: Functionally, I didn't think there were any obvious downsides either although some of @jzimbel-mbta 's comments I think point out that there could be some risks.

I think with the current team and everyone having a familiarity with both applications and the current work being done, the risk is low to none for doing a full git dependency. But if down the line a different org/team/individual ends up maintaining these applications without as much context, there's nothing to suggest to them that they shouldn't start using modules from `Screens` in `Screenplay` and vice versa.

docs/tech_specs/configuration_backend_approach.md
PaulJKim: Ohh hm yeah that mixed approach is interesting! Personally, I would be down to skip publishing to hex. In practice, this approach seems like it would be the same as the hex package besides the fact that version management will be slightly different.

docs/tech_specs/configuration_backend_approach.md
```diff
@@ -0,0 +1,107 @@
...
+
+Note: This document will not go into any lower-level implementation details for the Screenplay Configuration tool and will only focus on selecting an architectural approach for enabling Screenplay to reference code from Screens.
+
+# Possible approaches and rationale
+
+## 1. Umbrella Project
+
+### Description
+
+An Umbrella Project would allow us to group Screens and Screenplay in a single, unified codebase while still maintaining and deploying them as distinct applications. A dependency on Screens from Screenplay can then be defined and the code can be shared. Or the shared code can be separated into its own app within the umbrella and both applications can define a dependency on it.
```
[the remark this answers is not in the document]
PaulJKim: Yeah, I think this would be the way to go if we went with an umbrella

PR #224, opened 2023-09-15
tweak: Script update

cmaddox5:
**Asana task**: ad-hoc

Updated script so only one set of triptychs per platform is displayed.


scripts/fetch_places_and_screens.exs
```diff
@@ -272,6 +272,14 @@ live_screens =
         %{id: id, type: app_id, disabled: disabled, direction_id: direction_id}
     end
   )
+  |> Enum.reduce(%{}, fn {id, screen_list}, acc ->
```
jzimbel-mbta, line 275: I think this can just be a `Map.new` or `Enum.into` since the number of pairs in the map doesn't change. You're just mapping the value of each pair to a new one, so the mapper would look like
```ex
fn {id, screen_list} ->
  all_but_one_triptych = ...
  {id, all_but_one_triptych}
end
```

jzimbel-mbta (approved):
Looks good!

It should work fine as-is, just some suggestions for clarity.

scripts/fetch_places_and_screens.exs
```diff
@@ -405,11 +413,17 @@ stop_ids =
     %{"source_config" => %{"sources" => sources}} ->
       sources
 
-    %{"source_config" => [%{"sources" => top_sources}, %{"sources" => bottom_sources}]} ->
-      top_sources ++ bottom_sources
+    %{"source_config" => sources} ->
```
cmaddox5, line 416: @PaulJKim I had to make some changes to the paess logic to account for a new config shape. Can you look this over and let me know if I'm missing any or if you see some that aren't needed anymore?

PaulJKim: Sorry I missed this earlier! There's another field I think we'll need to account for which is `top_configs` which would be present if there's a `bottom_configs` field. You can copy over what I had actually ended up putting up a PR for a similar fix in #239 earlier this week 😅

PaulJKim: If you want, you can also bundle in the rest of those changes in that PR as well. One of them removes old logic that was resulting in duplicate PA/ESS signs being added to Screenplay, and the other one adds some missing/new custom labeling we need to apply to some sign zones.

PR #228, opened 2023-09-20
[tweak] Log issues with making SFTP connection

hannahpurcell:
Log if outfront ssh key is NOT a valid string; log connection error


lib/screenplay/outfront/sftp.ex
```diff
@@ -110,13 +110,21 @@ defmodule Screenplay.Outfront.SFTP do
     user = Application.get_env(:screenplay, :outfront_sftp_user)
     key = Application.get_env(:screenplay, :outfront_ssh_key)
 
+    if is_binary?(key) and String.length(key) > 0 do
```
hannahpurcell, line 113:
```suggestion
    if is_binary(key) and String.length(key) > 0 do
```

jzimbel-mbta (approved):

PR #230, opened 2023-10-20
Chore: Delete no longer used workflow

krisrjohnson21:
**Asana task**: [Update application repos to use the new version of the GitHub action(s) (v2 tag)](https://app.asana.com/0/1113179098808463/1205367032564122/f)

Description
This PR deletes the `deploy-prod-ecs` workflow, which is no longer used per discussion below


ianwestcott:
@hannahpurcell does this workflow even get used anymore, or does the team exclusively use the [Deploy to ECS](https://github.com/mbta/screenplay/blob/master/.github/workflows/deploy-ecs.yml) one now?

hannahpurcell:
> does this workflow even get used anymore, or does the team exclusively use the [Deploy to ECS](https://github.com/mbta/screenplay/blob/master/.github/workflows/deploy-ecs.yml) one now?

Thanks for asking / catching that! Correct, we don't use this workflow at all anymore and never knocked out the chore of deleting the workflow

krisrjohnson21:
> > does this workflow even get used anymore, or does the team exclusively use the [Deploy to ECS](https://github.com/mbta/screenplay/blob/master/.github/workflows/deploy-ecs.yml) one now?
> 
> @ianwestcott Thanks for asking / catching that! Correct, we don't use this workflow at all anymore and never knocked out the chore of deleting the workflow

Got it, thanks! I'll update this PR to just delete this workflow

ianwestcott (approved):

PR #232, opened 2023-10-31
Stub out permanent config operations

PaulJKim:
**Asana task**: [Stub out functions for Permanent Config operations](https://app.asana.com/0/1185117109217413/1205700210661285/f)

Description
Sets up routes and controller actions to be used for permanent config. Also adds a new file that will contain the business logic for each action.

### DIFF
--- a/lib/screenplay/config/config.ex
+++ b/lib/screenplay/config/config.ex
@@ -1,0 +1,10 @@
+defmodule Screenplay.Config.Config do
+  def add_new_screen() do
+  end
+
+  def edit_screen() do
+  end
+
+  def delete_screen() do
+  end
+end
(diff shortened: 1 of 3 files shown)


cmaddox5 (commented):
I'm wondering if the name `Config` is not clear as to what config we are working with given that this particular config is for a different app. Should we call the root directory `PermanentConfig` or `ScreensPermanentConfig` or something similar instead? What do you think?

PaulJKim:
@cmaddox5 yeah I had a similar thought. I went with `Config` initially since down the line there may be some "impermanent" config values. But considering that's not well defined, I think it's probably fine and clearer to go with something like `PermanentConfig` for the time being

cmaddox5 (approved):

PR #233, opened 2023-11-02
Add new group for permanent config

PaulJKim:
**Asana task**: [Create Cognito group for Permanent Configuration Tool](https://app.asana.com/0/1185117109217413/1205700210661282/f)

Description

This PR adds a new cognito group for users that will be allowed to configure new screens through Screenplay.

Related Terraform changes: https://github.com/mbta/devops/pull/1344

### DIFF
--- a/lib/screenplay_web/auth_manager/auth_manager.ex
+++ b/lib/screenplay_web/auth_manager/auth_manager.ex
@@ -3,9 +3,10 @@ defmodule ScreenplayWeb.AuthManager do
 
   use Guardian, otp_app: :screenplay
 
-  @type access_level :: :none | :read_only | :admin
+  @type access_level :: :none | :read_only | :admin | :configurer
 
-  @screenplay_admin_group "screenplay"
+  @screenplay_admin_group "screenplay-emergency-admin"
+  @screenplay_configurer_group "screenplay-screen-configurer"
 
   @spec subject_for_token(
           resource :: Guardian.Token.resource(),
@@ -24,11 +25,16 @@ defmodule ScreenplayWeb.AuthManager do
   def resource_from_claims(_), do: {:error, :invalid_claims}
 
   @spec claims_access_level(Guardian.Token.claims()) :: access_level()
-  def claims_access_level(%{"groups" => groups}) do
-    if not is_nil(groups) and @screenplay_admin_group in groups do
-      :admin
-    else
-      :read_only
+  def claims_access_level(%{"groups" => groups}) when not is_nil(groups) do
+    cond do
+      @screenplay_admin_group in groups ->
+        :admin
+
+      @screenplay_configurer_group in groups ->
+        :configurer
+
+      true ->
+        :read_only
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -3,9 +3,10 @@ defmodule ScreenplayWeb.AuthManager do
 
   use Guardian, otp_app: :screenplay
 
-  @type access_level :: :none | :read_only | :admin
+  @type access_level :: :none | :read_only | :admin | :configurer
 
-  @screenplay_admin_group "screenplay"
+  @screenplay_admin_group "screenplay-emergency-admin"
+  @screenplay_configurer_group "screenplay-screen-configurer"
```
hannahpurcell, line 9: Ohhhh! I just realized, we didn't need to add a new group but could just use the existing group called `screens-admin`. I'm not sure `screenplay-screen-configurer` is clearer than the name of the existing group. What do you think about nixing that new group and keeping the old one, especially since you're essentially renaming the group here to `screenplay_configurer_group` (or maybe `screens_admin_group`)

hannahpurcell (commented):
One question on this code below, and a separate question on the merged devops PR: the description of the `screenplay-emergency-admin` is actually no longer true. Now that Screenplay is more than just the OFM takeover tool, the description of that group should be something like "Access to the Emergency Takeover tool in Screenplay". Not a big deal, but I think it will need to be changed before we add users to the group. (If users are added, and we change the group description, would it wipe out the users?)

lib/screenplay_web/auth_manager/auth_manager.ex
PaulJKim: When you refer to 'group description' are you just referring the name of the group (e.g. `screenplay-emergency-admin` or something else? If it's just the name, I feel like `screenplay-emergency-admin` does a pretty decent job of describing a group that represents users who have access to the OFM takeover tool in Screenplay, but not sure if the 'description' of a group is a specific attribute that belongs to the cognito group?

And yeah that's a good point and agreed, we should just re-use the old `screens-admin` group rather than setting up a new one for Screenplay that will contain the same users.

hannahpurcell: [Group description](https://github.com/mbta/devops/pull/1344/files#diff-7b658ee617733625d8c755ac321b1462018a45c08a6e24f1e7b63533dae7dac7R92). I agree the name `screenplay-emergency-admin` is better than `screenplay`, but the description of "Access to the Screenplay tool" is no longer quite accurate

PaulJKim: Ah, that description. My gut tells me that changing the description wouldn't do anything to change what users belong to that group, but I don't know for sure. I can ask infra about that one to be sure though.

hannahpurcell (approved):
👍 I also see Kris' reply in the infra channel. Glad we can update the description without much bother!

PR #234, opened 2023-11-08
[WIP] POC for using ETag to verify screens config version

PaulJKim:
**Asana task**: [S3: Use ETag to verify latest version is being edited ](https://app.asana.com/0/1185117109217413/1205700210661288/f)

Description
This is a basic POC for how we can use the `ETag` field returned from calls to S3 to understand if the config has changed since a user began preparing to add (or remove) a screen's config.

Currently, the logic simply allows or disallows an update based on the given and current value of the ETag but I'm not 100% sure if that's the direction we want to take things. 

The current approach also still allows for a race condition if two clients tried to make updates to the screens config at the same time. Not sure how much we need to worry about that at this point given our limited user pool, but we could ensure requests are handled serially by passing the operations to a GenServer and taking advantage of the nature of its message queue.

### DIFF
--- a/lib/screenplay/config/config.ex
+++ b/lib/screenplay/config/config.ex
@@ -1,12 +1,43 @@
 defmodule Screenplay.Config.PermanentConfig do
   @moduledoc false
+  alias Screenplay.Config.S3Fetch
 
-  def add_new_screen do
+  @spec add_new_screen(binary(), map(), binary()) :: :error | :ok
+  def add_new_screen(screen_id, screen, etag) do
+    case get_current_config(etag) do
+      {:ok, config} ->
+        config
+        |> Map.put(screen_id, Jason.decode!(screen))
+        |> S3Fetch.put_screens_config()
+
+      _ ->
+        :error
+    end
   end
 
   def edit_screen do
   end
 
-  def delete_screen do
+  @spec delete_screen(binary(), binary()) :: :error | :ok
+  def delete_screen(screen_id, etag) do
+    case get_current_config(etag) do
+      {:ok, config} ->
+        config
+        |> Map.delete(screen_id)
+        |> S3Fetch.put_screens_config()
+
+      _ ->
+        :error
+    end
+  end
+
+  defp get_current_config(etag) do
+    with {:ok, config, current_etag} <- S3Fetch.get_screens_config() do
+      if etag == current_etag,
+        do: {:ok, config},
+        else: :error
+    else
+      _ -> :error
+    end
   end
 end
--- a/lib/screenplay/config/config.ex
+++ b/lib/screenplay/config/config.ex
@@ -2,7 +2,8 @@ defmodule Screenplay.Config.PermanentConfig do
   @moduledoc false
   alias Screenplay.Config.S3Fetch
 
-  @spec add_new_screen(binary(), map(), binary()) :: :error | :ok
+  @spec add_new_screen(binary(), map(), binary()) ::
+          {:error, :etag_mismatch | :config_not_fetched | :config_not_written} | :ok
   def add_new_screen(screen_id, screen, etag) do
     case get_current_config(etag) do
       {:ok, config} ->
@@ -10,12 +11,13 @@ defmodule Screenplay.Config.PermanentConfig do
         |> Map.put(screen_id, Jason.decode!(screen))
         |> S3Fetch.put_screens_config()
 
-      _ ->
-        :error
+      error ->
+        error
     end
   end
 
-  @spec delete_screen(binary(), binary()) :: :error | :ok
+  @spec delete_screen(binary(), binary()) ::
+          {:error, :etag_mismatch | :config_not_fetched | :config_not_written} | :ok
   def delete_screen(screen_id, etag) do
     case get_current_config(etag) do
       {:ok, config} ->
@@ -23,8 +25,8 @@ defmodule Screenplay.Config.PermanentConfig do
         |> Map.delete(screen_id)
         |> S3Fetch.put_screens_config()
 
-      _ ->
-        :error
+      error ->
+        error
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/screenplay/config/s3_fetch.ex
```diff
@@ -35,12 +49,35 @@ defmodule Screenplay.Config.S3Fetch do
   end
 
   defp config_path_for_environment(file_spec) do
-    base_path = "screenplay/#{Application.get_env(:screenplay, :environment_name, "dev")}"
```
PaulJKim, line 38: This default actually wasn't working because `Application.get_env/2` only returns the default when the key doesn't exist. Because the key was being set to `nil` in the case where the env var wasn't set, the config key technically still exists at runtime, just with a `nil` value. I made a change to set the default within `runtime.exs` instead

lib/screenplay/config/config.ex
```diff
@@ -1,12 +1,45 @@
...
+        |> Map.put(screen_id, Jason.decode!(screen))
+        |> S3Fetch.put_screens_config()
+
+      _ ->
+        :error
+    end
   end
 
   def edit_screen do
   end
```
PaulJKim, line 19: Wonder if we really need a proper `edit` action. It seems like we may be just fine with add and delete. Even when an existing config needs to be edited, I think it'd be easier to just have the client pass along the full screen config object and have the server just overwrite the existing key with an `add`

lib/screenplay/config/config.ex
```diff
@@ -1,12 +1,45 @@
...
+        :error
+    end
+  end
+
+  defp get_current_config(etag) do
+    case S3Fetch.get_screens_config() do
+      {:ok, config, current_etag} ->
+        if etag == current_etag,
+          do: {:ok, config},
+          else: :error
```
cmaddox5, line 39: Should be able to reduce this a little bit with more strict pattern matching in the `case`. 

```suggestion
      {:ok, config, ^etag} -> {:ok, config}
```

lib/screenplay_web/controllers/config_controller.ex
```diff
@@ -3,18 +3,23 @@ defmodule ScreenplayWeb.ConfigController do
...
+  def add(conn, %{"screen_id" => screen_id, "screen" => screen, "etag" => etag}) do
+    case PermanentConfig.add_new_screen(screen_id, screen, etag) do
+      :ok ->
+        send_resp(conn, 200, "OK")
 
-  def edit(conn, _params) do
-    PermanentConfig.edit_screen()
-    conn
+      :error ->
+        send_resp(conn, 400, "Unable to add new screen")
```
cmaddox5, line 12: Only issue with removing the `/edit` endpoint is that this error message is a little confusing. Should this be changed to something more general like `Config is outdated`? Then again, this will mostly be used as a comment of sorts because I'm sure the client will ignore whatever goes there.

PaulJKim: Hmm well technically the action can also return an error if either the get or put S3 operations fail themselves. I thought about adding more specific error reasons to those cases but it seemed a bit like overkill, especially since like you said, it's likely the client won't make sure of those more specified error messages, so I just left it as a more general message communicating "that didn't work". But I could certainly still see it being helpful to be more specific here, if even for local dev. What do you think about returning error reasons and specifying the error message returned with the response here accordingly?

cmaddox5: I think that's a good idea. And it still keeps the code conventional by using the error tuple pattern.

PaulJKim: Just added some more specific error handling. Let me know if you have additional thoughts!

cmaddox5 (approved):
I'm a fan of this approach. It definitely still isn't perfect like you mentioned, but the race condition should be super rare. The odds of two people saving a config at the exact same time is something we would run into with the admin tool and that hasn't happened to me since I started working with it.

lib/screenplay/config/config.ex
hannahpurcell: 🤔 This might be false logic, so lemme know if I'm missing something. Edit might be needed in this case: a user is adding a handful of new screens. Configures them, goes to the preview page (which involves a save to the `pending screens` file in s3, so that we can build the screen previews on the next page), then they realize one of the screen ids has a typo? Then they go to the previous page to fix it. That screen id is the key for the pending screens file, so a simple `add` action might not be enough at this point?

cmaddox5: Yikes, screen ID is the one field I didn't think of in this scenario. That might be a UX question for me. I feel like in most workflows where you can edit, the key of the item you are editing is not something you can change. That would normally requiring deleting the item and adding a "new" item with the correct key. I know pending screens are a little different considering they are meant to be WIP screens, but should we allow for editing a saved screen ID?

hannahpurcell: Actually! After reviewing the [footage](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc/Permanent-Configuration?type=design&node-id=1520-98508&mode=design&t=KixhwTGweilaGr0o-4), I'm thinking that the Edit button on the Preview page does kinda nullify my expectations for this page. I thought there would be a "back" button or a button that allows the user to edit a particular screen, but it seems like it's a bulk edit, and I'm not sure a screen ID should be editable in that case...

PaulJKim: It seems like the discussion is starting to veer into full implementation territory and discussing how the UX will flow. With having in mind that this ticket was more to do a POC of the ETag approach (which it seems like we like and will serve our purposes) would we be okay with merge in what's here and saving the rest for a task for the next iteration of the implementation?

hannahpurcell: Yeah, totally agree. I put my name on it because I'm still reviewing, but will take my name off when I'm done catching up

hannahpurcell (approved):
Ok tracking! This looks great to me 👍

PR #235, opened 2023-11-09
feat: Select screen type page

cmaddox5:
**Asana task**: [[Perm config] Build select screen type](https://app.asana.com/0/1185117109217413/1205450031351981/f)

New page to be used as a starting point for Permanent Configuration. Includes the addition of a custom component `ButtonImage`. I tried to make the component as general as possible so it can be used outside of this page. This page is only configured to navigate when GL E-Ink is clicked. Other buttons will set state but will not change the layout.

**Because routing is covered in another [task](https://app.asana.com/0/1185117109217413/1205709923095884/f), there is no way to see the page without setting up some temporary routing in your project.

Note: It looks like I misread the other [task](https://app.asana.com/0/1185117109217413/1205724153034243/f) and built the layout for that task in this branch. I feel like even with that, this PR seems manageable from a reviewers perspective. If you disagree, please let me know and I will rearrange some commits.


assets/css/dashboard/configure-screens-page.scss
```diff
@@ -0,0 +1,29 @@
...
+
+    .row {
+      justify-content: center;
+      gap: 24px;
+      margin-bottom: 32px;
+    }
+
+    .col {
+      width: fit-content;
+      max-width: 306px;
```
hannahpurcell, line 26: This max-width on the column is not getting applied on the button-image itself. The effective max-width on the button is 282px.

I'm thinking the column / row approach is not the most flexible way of building this. Instead, maybe use flex, so that smaller viewports can shrink to have only 2 columns per row, [like this](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc/Permanent-Configuration?type=design&node-id=1764-52754&mode=dev) (I know I never did my lunch chat on flex... 🙃 maybe time to do it?? lemme know if that'd still be helpful)

assets/js/components/Dashboard/PermanentConfiguration/AppBar.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const Appbar: ComponentType<AppBarProps> = (props: AppBarProps) => {
+  const { title } = props;
+  const navigate = useNavigate();
+
+  return (
+    <Navbar variant="dark" data-bs-theme="dark">
+      <Container>
+        <Navbar.Brand>{title}</Navbar.Brand>
+        <CloseButton variant="white" onClick={() => navigate(-1)} />
```
hannahpurcell, line 17: In the design, the color of the x looks the same as the "Configure screens" label, but not in implementation. Then I realized the implemented button has a hover color, which might be something do want to keep, but I'm not sure what the UI should be here, so maybe a q for Mary.

assets/js/components/Dashboard/PermanentConfiguration/ButtonImage.tsx
```diff
@@ -0,0 +1,26 @@
...
+  label: string;
+  onClick: () => unknown;
+}
+
+const ButtonImage: ComponentType<ButtonImageProps> = (
+  props: ButtonImageProps
+) => {
+  const { fileName, label, onClick } = props;
+  return (
+    <div className="button-image-container">
```
hannahpurcell, line 15: The content of this container should be horizontally centered, but isn't

assets/css/dashboard/configure-screens-page.scss
cmaddox5: The main reason for using the row/column approach is because it uses flex under the hood. With that said, I totally glossed over that image somehow. I thought that button sizes needed to shrink as the screen width decreased, but it actually just needs to reduce the number of columns. Bootstrap has what they call "Row Columns" to help with this. We can dictate the number of columns we want for each breakpoint. The `md` breakpoint is for widths `≥768px`. The `lg` is for `≥992px`. So if we say `lg={3}` and `md={2}`, we should see the behavior Mary designed. I will work on this and see what I come up with.

assets/js/components/Dashboard/PermanentConfiguration/ButtonImage.tsx
cmaddox5: Ugh, another case of me only testing with FF. For some reason, Chrome doesn't center them automatically like FF does. Will add explicit centering.

assets/css/dashboard/configure-screens-page.scss
```diff
@@ -0,0 +1,29 @@
...
+  min-height: 100vh;
+
+  .select-screen-type {
+    max-width: 1320px;
+
+    &__title {
+      white-space: nowrap;
+      width: fit-content;
+      font-weight: 700;
+      font-size: 28px;
```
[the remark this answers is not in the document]
cmaddox5: Thanks for the link. Something weird I'm seeing after changing that is the letter spacing doesn't look right when implementing. Gonna need to play around with it a little to see if that can be fixed.

Screenshot 2023-11-15 at 5 46 32 PM

assets/js/components/Dashboard/PermanentConfiguration/AppBar.tsx
cmaddox5: Asked Mary what she thought [here](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc?node-id=1760:58899&mode=design#618678226). After explaining to her how unusually complicated it is to change the color of the Bootstrap close button, she is ok with using its defaults.

assets/css/dashboard/configure-screens-page.scss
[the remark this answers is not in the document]
cmaddox5: Went with the global CSS class approach. The actual `h3` tag just doesn't have the right letter spacing. The class is placed where all components can see it so when Mary designs something with the h3 typography, we can add `h3` to `classNames`.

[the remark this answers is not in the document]
hannahpurcell: Oh no... I'm looking at the designs more closely and finding inconsistencies. Not with your implementation, but the design itself. Two identical fonts are labeled H2 & H3. I'm asking Mary a question on it

assets/css/screenplay.scss
```diff
@@ -1,3 +1,7 @@
+$theme-colors: (
+  "primary": #384956,
+);
```
hannahpurcell, line 3: I don't think this is being used anymore? Not sure

cmaddox5: While it isn't being explicitly used by us, bootstrap uses it to calculate colors of element hover, focus, borders, etc. So I think it's good to keep.

assets/js/components/Dashboard/PermanentConfiguration/ConfigureScreensPage.tsx
hannahpurcell, line 1: With this wizard, I'm not sure that all the pages/steps should be preserved in a single component. Do you think this would be better-organized if the select-screen-type piece was isolated to a new component, then the stations list page would be a new component, and then the control to step between the components would be managed here in this file?

cmaddox5: For some reason, I did decide to go that direction but never made the change here. In #241, I moved the component into its own file. I agree that having them all in one file would be a nightmare. Do you want me to go ahead and make that change here? 

You also make a good point about the stations list being its own component, and I did not do that. I'll make sure to do that before merging the other PR.

assets/css/screenplay.scss
hannahpurcell: Gotcha. Can the color itself go into the colors file?

cmaddox5: I was taking the whole "override colors before imports" a little too literally. These overrides just need to be above the bootstrap import. If I import `variables.scss` first, we can use it like we do everywhere else.

assets/js/components/Dashboard/PermanentConfiguration/ConfigureScreensPage.tsx
hannahpurcell: Ok! That's ok, no need to move those changes here. Thanks 👍

hannahpurcell (approved):

hannahpurcell (changes requested):
Oops, I was trigger happy with the approval. I missed a comment I had drafted:

Playing spot-the-difference between the mocks and the app here. Two things I noticed on the [button-image](https://www.figma.com/file/OI3ir199riWIlDMpNn9vtz/Screenplay-Design-System?node-id=817%3A1068&mode=dev). The font-weight for the screen type should be 700. And I think the bottom text should have an 18px bottom padding/margin instead of 6px.

And then on the page as a whole, if the vertical scroll bar is available and you scroll down, the background should extend all the way to the bottom. Not sure if that's fixed in a later branch yet, but would be good to have that fixed at some point.
Screenshot 2023-12-13 at 2 15 30 PM

assets/css/dashboard/configure-screens-page.scss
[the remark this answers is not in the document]
hannahpurcell: Ok I got a [reply](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc?node-id=1760:58899&mode=dev#645365932) from Mary on the font confusion. Not sure what happened, but using H3 in this PR is correct, and the font size you use is correct.

cmaddox5:
@hannahpurcell Geez, you got some eagle eyes. Great finds. I think I addressed the issues you mentioned.

hannahpurcell (approved):
Thanks, looks good 👍

PR #238, opened 2023-11-28
refactor: PlaceRow component

cmaddox5:
**Asana task**: [Update place row component](https://app.asana.com/0/1185117109217413/1205552592584843/f)

There are 4 primary goals for these updates

1. Change `disabled` logic so it is read from `props`
2. Add a `select-box` variant
3. Move as much `Accordion` logic out as possible so the component is more reusable
4. Maintain all existing functionality 

The original `PlaceRow` component made the assumption that it would always be used in an `Accordion`. That is no longer the case in Permanent Configuration. We now need the ability to use this component without any accordion logic. This was a lot more work than I anticipated, mostly due to issues with hooks and deciding where accordion logic needed to live.

The solution I went with is moving all accordion logic into the `PlacesList`. This component is where the top level `Accordion` component lives, so it makes sense that all the other logic lives in the same place. `PlacesList` now displays a list of `PlaceRowAccordion`s, which is a wrapper of the main `PlaceRow`. This wrapper handles all of the accordion logic that `PlaceRow` used to be responsible for. 

~~**I know tests are failing, but I wanted to get the first round of review started while I worked on fixing them.~~
Got tests sorted out. Should be good to go now.

### DIFF
--- a/assets/js/components/Dashboard/PlacesPage.tsx
+++ b/assets/js/components/Dashboard/PlacesPage.tsx
@@ -1,18 +1,10 @@
-import React, { ComponentType, useContext } from "react";
-import PlaceRow from "./PlaceRow";
+import React, { ComponentType } from "react";
 import PlacesActionBar from "./PlacesActionBar";
 import FilterDropdown from "./FilterDropdown";
-import {
-  Accordion,
-  AccordionContext,
-  Col,
-  Container,
-  Row,
-  useAccordionButton,
-} from "react-bootstrap";
+import PlaceRowAccordion from "./PlaceRowAccordion";
+import { Accordion, Col, Container, Row } from "react-bootstrap";
 import { ArrowDown, ArrowUp } from "react-bootstrap-icons";
 import { Place } from "../../models/place";
-import { Screen } from "../../models/screen";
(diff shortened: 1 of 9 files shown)


assets/js/components/Dashboard/PlacesPage.tsx
```diff
@@ -301,20 +281,27 @@ const PlacesList: ComponentType<PlacesListProps> = ({
...
-              showAnimation={
+              canShowAnimation={
                 showAnimationForNewPlaces &&
                 prevPlaceIds !== undefined &&
                 !prevPlaceIds.includes(place.id)
               }
+              dispatch={dispatch}
+              activeEventKeys={activeEventKeys}
+              sortDirection={sortDirection}
+              isFiltered={isFiltered}
```
PaulJKim, line 302: This prop gets passed to the accordion but it doesn't seem to be used anywhere within the component

PaulJKim (approved):
One tiny comment, but tested locally and things look good in terms of maintaining functionality. And the overall approach makes sense to me!

assets/js/components/Dashboard/PlacesPage.tsx
cmaddox5: Good catch! Got that removed.

PR #240, opened 2023-11-29
feat: Fetch station list for config pages

cmaddox5:
**Asana task**: [Fetch stations logic](https://app.asana.com/0/1185117109217413/1205973195471517/f)

Added a function for fetching station list need on future config pages. Subway will be easy because we hold all subway stations in app state. We simply need to get that list and filter it down.


jzimbel-mbta (approved):
Good to go!

PR #241, opened 2023-12-05
feat: GL E-Ink Config Workflow

cmaddox5:
**Asana task**: [Stub out page layout, with station selection list](https://app.asana.com/0/1185117109217413/1205855245091613/f)

This PR adds stubs out the first page of the GL E-Ink workflow. Includes the places list and state management to keep track of selected rows. Does not include the search box or the bottom action bar (both will be implemented in a separate task).


assets/js/util.ts
```diff
@@ -196,3 +197,23 @@ export const sortScreens = (screenList: Screen[]) => {
...
+
+  places.sort((placeA, placeB) => {
+    const indexA = stationOrder.findIndex((station) => {
+      return station.name.toLowerCase() === placeA.name.toLowerCase();
+    });
+    const indexB = stationOrder.findIndex((station) => {
+      return station.name.toLowerCase() === placeB.name.toLowerCase();
+    });
+    return indexA - indexB;
+  });
```
jzimbel-mbta, line 216: This is pretty inefficient since every time we need to compare two elements of `places` (potentially once for every unique pair of elements—`O(n^2)`), we need to iterate through `stationOrder` twice. That's something like cubic time—`O(m*n^2)` where m is `stationOrder.length` and n is `places.length`—in the worst case.

It should be faster to compute the indices once for each place, up front. You can also create a map from `stationOrder` to make lookups faster. What we really want is something like `Enum.sort_by/2`, but that method isn't available in standard JS so we can do it ourselves. _Or_, add lodash to the project and use [its `sortBy` function](https://lodash.com/docs/4.17.15#sortBy).

```ts
const stationOrderToIndex = Object.from_entries(
  stationOrder.map((station, i) => [station.name.toLowerCase(), i])
);

const placesByStationOrder = places.map((place) => ({
  place,
  index: stationOrderToIndex[place.name.toLowerCase()]
}));

const sortedPlaces = placesByStationOrder
                       .sort(({index: i1}, {index: i2}) => i1 - i2) 
                       .map(({place}) => place);

return reverse? sortedPlaces.reverse() : sortedPlaces;
```
(I didn't write this in vscode, so please double check that it works if you decide to use it!)

jzimbel-mbta (approved):
Looks good overall, just some minor things that should be quick to fix.

I'm approving now assuming you respond as you see fit, but feel free to assign it back to me after making changes if you want me to take another look.

assets/js/components/Dashboard/PlaceRow.tsx
```diff
@@ -40,34 +40,40 @@ const AccordionToggle = ({
...
 }: SelectBoxToggleProps): JSX.Element => {
   return (
     <div className="place-row__toggle">
-      <Form.Check disabled={disabled} onChange={onChange} />
+      <Form.Check
+        disabled={disabled}
+        checked={checked}
+        onChange={(e) => {
+          e.stopPropagation();
+        }}
```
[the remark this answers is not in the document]
cmaddox5: This actually isn't needed anymore. One of my earlier implementation had an issue with the default `onChange` so I added this to keep it from doing anything. Now the `onRowClick` handles all of the click logic for the row.

assets/js/components/Dashboard/PlaceRowAccordion.tsx
```diff
@@ -94,7 +94,7 @@ const PlaceRowAccordion: ComponentType<PlaceRowAccordionProps> = ({
     place.screens.filter((screen) => !screen.hidden).length > 0;
   const rowOnClick = hasScreens
     ? useAccordionButton(place.id, () => handleClickAccordion(place.id))
```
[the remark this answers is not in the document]
cmaddox5: Very interesting. I feel like you mentioning this is jogging my memory a bit on this particular hook rule. That change actually works the same way, so I'll change it to play it safe.

assets/js/util.ts
cmaddox5: Wow, great catch. This is working great as is so I'll stick with this instead of adding `lodash` at the moment.

assets/css/screenplay.scss
```diff
@@ -74,7 +65,24 @@ input[type="button"] {
   letter-spacing: 0.5px;
 }
 
+.h2 {
+  font-weight: 700;
+  font-size: 28px;
+}
```
hannahpurcell, line 71: just popping in to note that Mary says here H2 is [actually larger](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc?node-id=1760:58899&mode=dev#645365932)

PR #242, opened 2023-12-05
Bottom action bar

PaulJKim:
**Asana task**: [[Perm config] Create Bottom Action Bar Component + paging logic](https://app.asana.com/0/1185117109217413/1205450031351973/f)

Description
Branched off of #235 

This PR adds a reusable bottom action bar that takes in button labels and actions as props.

State should be preserved as a user pages through the configuration steps, and while there is currently no state to be saved, I think we should be able to initialize/maintain state at the top level component for a given screen type (`GlEinkConfigPage` in this case) and pass that down to any future child components that will be used for the actual configurable aspects.


hannahpurcell (approved):
Alright! Looks great to me

cmaddox5:
@PaulJKim Just FYI while you are out, I am going to resolve the merge conflicts and merge this into the feature branch. I know you were waiting on some other PRs to be merged before doing so, but I believe we are caught up on those.

PR #244, opened 2023-12-13
Update Erlang, Elixir, Phoenix

cmaddox5:
**Asana task**: ad-hoc

Screenplay needs to be on at least Elixir 1.14 before it can use `screens-config-lib`. I decided to go all out and upgrade to the most recent stable version of a few different dependencies. I did a lot of testing to make sure everything looks and acts like it did before. We now also have access to [verified routes](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html) (although I'm not sure we'll use them). Seemed like a cool feature and it was harmless it add.


cmaddox5:
Changed the base to the feature branch. After our convo yesterday, I agree that this doesn't need to be in `master` right now.

hannahpurcell (approved):
Looks good!

PR #251, opened 2024-02-07
Specify commit hash of screens_config dep to avoid needing to do `mix deps.update`

jzimbel-mbta:
**Asana task**: ad hoc

Did a little testing on this—specifying the commit hash forces mix to grab the code at that specific commit when you run `mix deps.get`.

Previously without a commit specified, it would not pull new code on main if the current commit it was using was already on main. You'd have to run `mix deps.update` to force it to pull down the latest code on main.


hannahpurcell:
So when we push new commits to the config library, we'd need remember to do a code change here?

jzimbel-mbta:
Yup. This lets us keep Screenplay locked to an explicitly stated version of the library until it's ready to use the new version.

hannahpurcell (approved):
Ok! Could you socialize this with the -eng channel if it hasn't been already, just so there's a Slack record of this decision?

PR #253, opened 2024-02-12
feat: Publish Pending Screens

cmaddox5:
**Asana task**: [[Perm config] Publish a station's screens](https://app.asana.com/0/1185117109217413/1205973195471512/f)

This PR adds the server logic for publishing a pending screens. The new endpoint takes a single place ID and moves any pending screens for the place to the main screens config. Then, it adds the screen ID to `places_and_screens.json` so it will show on the places page for Screenplay.


lib/screenplay_web/controllers/config_controller.ex
```diff
@@ -89,6 +89,23 @@ defmodule ScreenplayWeb.ConfigController do
...
+      }) do
+    app_id_atom = String.to_existing_atom(app_id)
+
+    case PermanentConfig.publish_pending_screens(
+           place_id,
+           app_id_atom,
+           hidden_from_screenplay_ids
+         ) do
+      :ok -> send_resp(conn, 200, "OK")
+      _ -> send_resp(conn, 400, "Could not publish screens")
```
PaulJKim, line 105: Should this technically be a 500 error? I thought 4XX errors are typically given when the client has provided a bad request, whereas this case seems to indicate a server-side issue

cmaddox5: Yeah, you're totally right. Changed that and a couple in `put` as well. I left `:version_mismatch` as a 400 error considering the version we check comes from the request. Seemed like a good use of 400, but let me know if you disagree.

PaulJKim: Yeah I think `:version_mismatch` does make sense as a 400

PaulJKim (approved):
Looks good to me!

PR #254, opened 2024-02-12
WIP: Pending screens page

cmaddox5:
**Asana task**: [title](url)

Description


assets/js/components/Dashboard/PendingScreenDetail.tsx
```diff
@@ -0,0 +1,32 @@
...
+  config: ScreenConfiguration;
+  // I'm still not super clear what the purpose of this is, or how/where to store it...
+  // We now have two different ways of hiding screens from the places list:
+  // - When hidden via the old admin tool, the screen is excluded from places_and_screens.json
+  // - When hidden this new way, the screen is known to Screenplay--and thus needs a full, valid
+  //   config + Screenplay-relevant metadata, since we show info about it here on
+  //   the pending page--but ignored when rendering the places list.
+  //
+  // This seems messy and I'm not sure how to explain the issue to design.
+  isHiddenOnPlacesPage: boolean;
```
cmaddox5, line 20: I'm not sure this is necessary. By default, screens in the pending config will be hidden from the places page. Maybe you can just look up the Screenplay-relevant metadata and pending screens separately instead of needing to keep pending screens in the place object? Looking at the designs, the only metadata you need from `places_and_screens` is the ID and name. The rest could be ignored.

cmaddox5: Actually, looks like I misread the design. There are live screens shown on this page. But that makes me wonder, should there be? It feels like only pending screens should live here so that way the user doesn't have to wonder what kind of screen they are looking at if they happen to overlook the status.

lib/screenplay_web/controllers/config_controller.ex
```diff
@@ -79,4 +110,13 @@ defmodule ScreenplayWeb.ConfigController do
   defp place_id_has_screen?(place_id, app_id, _),
     do:
       raise("place_id_has_screen/2 not implemented for app_id: #{app_id}, place_id: #{place_id}")
+
+  # TODO: This is yucky but maybe ok?
+  # Need a function that can take an arbitrary screen config and find the place it's associated with.
+  # Finding this via the places_and_screens config would be expensive if we need t do it for a bunch of screen configs
+  defp screen_to_place_id(%Screen{app_params: %GlEink{footer: %{stop_id: stop_id}}}),
```
cmaddox5, line 117: I think this is fine, and is similar to what I made in the backend piece of this work ([here](https://github.com/mbta/screenplay/blob/9f50857a1d104a9bd821c3c1848f4e14fa6584e6/lib/screenplay/config/config.ex#L263)). Since most of our apps hold the stop IDs in different places, this seems like the natural way to do it.

PR #255, opened 2024-02-21
tweak: Config fetch logic changes

cmaddox5:
**Asana task**: [Remove Cache module for Pending Screens Configuration](https://app.asana.com/0/1185117109217413/1206645609745207/f)

This PR includes removing pending config caching and switching from using S3 etags to version IDs. Using version IDs helps with reverting changes if something goes wrong while publishing screens.


lib/screenplay_web/controllers/config_controller.ex
```diff
@@ -38,6 +41,21 @@ defmodule ScreenplayWeb.ConfigController do
...
+          %PendingConfig{screens: pending_screens} =
+            config
+            |> Jason.decode!()
+            |> PendingConfig.from_json()
+
+          {pending_screens, version_id}
+
+        _ ->
+          Logger.error("Could not fetch pending screens config in existing_screens/2")
+          []
```
jzimbel-mbta, line 56: The empty list returned here will cause the server to give a 500 status since `[]` won't match the pattern `{pending_screens_config, version_id}`.

Do we want that? We don't really need to do any kind of graceful handling here since a failed S3 fetch would probably mean something is very broken, but the error that would get raised if this happened could be a bit confusing.

I guess I'm wondering if it would make sense to just `raise` a runtime error instead of / in addition to the log, rather than returning anything from the case expression.

lib/screenplay_web/controllers/config_controller.ex
```diff
@@ -50,16 +68,24 @@ defmodule ScreenplayWeb.ConfigController do
...
-        live_screens = ScreensConfigCache.screens(filter_fn)
-        pending_screens = PendingScreensConfigCache.screens(filter_fn)
+        live_screens =
+          ScreensConfigCache.screens(filter_fn)
+          |> Enum.map(fn {k, v} -> {k, Screen.to_json(v)} end)
+          |> Enum.into(%{})
+
+        pending_screens =
+          pending_screens_config
+          |> Enum.filter(&filter_fn.(&1))
```
jzimbel-mbta, line 78:
```suggestion
          |> Enum.filter(filter_fn)
```

jzimbel-mbta (approved):
Two super minor things. Good to go!

lib/screenplay_web/controllers/config_controller.ex
cmaddox5: Ahh, thank you. I was having an embarrassingly hard time figuring out how to pass an anonymous function variable 🤦

lib/screenplay_web/controllers/config_controller.ex
jzimbel-mbta: Raising works for now, and for the error handling task we could adjust this to send a response with a more informative message for the client to display.

I think this would fall in the bucket of failure modes where we'd tell the user to try again and contact us if it continues to not work.

PR #258, opened 2024-03-04
feat: Edit Pending Screens

cmaddox5:
**Asana task**: [[Perm config] Edit station's pending screens page](https://app.asana.com/0/1185117109217413/1206114010686515/f)

This PR includes most of the work needed for the "Edit" button to work on the Pending Screens Page. The actual button does not exist quite yet (will be added in this [task](https://app.asana.com/0/1185117109217413/1205973195471510/f)), but the logic would not change once the button is available.

### DIFF
--- a/assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/GlEinkWorkflow.tsx
+++ b/assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/GlEinkWorkflow.tsx
@@ -1,9 +1,14 @@
-import React, { ComponentType, useState } from "react";
+import React, {
+  ComponentType,
+  useEffect,
+  useLayoutEffect,
+  useState,
+} from "react";
 import ConfigureScreensWorkflowPage, {
   PlaceIdsAndNewScreens,
 } from "./ConfigureScreensPage";
 import BottomActionBar from "../../BottomActionBar";
-import { useNavigate } from "react-router-dom";
+import { useLocation, useNavigate } from "react-router-dom";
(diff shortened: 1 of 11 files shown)


assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/GlEinkWorkflow.tsx
```diff
@@ -1,25 +1,48 @@
...
 
-  const navigate = useNavigate();
-  const [configStep, setConfigStep] = useState<number>(0);
+  const getPlacesList = () => {
+    return places.filter((place) =>
+      place.routes.some((route) => route.startsWith("Green"))
+    );
+  };
+
+  useLayoutEffect(() => {
```
PaulJKim, line 35: Pardon my unfamiliarity with React, but why do we need to use `useLayoutEffect` over `useEffect` in this case? My basic understanding is that `useLayoutEffect` makes sure to run synchronously before the browser repaints, but what about the logic requires that exactly?

cmaddox5: It's actually a way to make the transition smoother when navigating from the Edit button on the Pending page (not present in this PR because task is still in progress). With `useEffect`, the page renders, the hook runs, the component rerenders from the state change. You can actually see the layout change when this happens, which looks odd. Using `useLayoutEffect` makes sure that the state is set correctly before the component renders for the first time.

PaulJKim (approved):

PR #262, opened 2024-03-14
fix: Pending Page integration 

cmaddox5:
**Asana task**: ad-hoc

While testing out Jon's Pending Page work, I found some integration issues for pieces of work that were done independently of the work that would use it.

1. Fixed issue where `AlertBanner` had no text if user is currently at `/pending`. Also prevented the banner from showing on `/configure-screens/*`.
2. Fixed an issue where deleting an existing pending screen didn't actually delete it.
3. Fixed navigation state object key name so state can be initialized correctly. Also initialized validation error state during navigation. This was normally done in the `StationSelectPage`, but that page is skipped during the edit workflow.
4. Moved inline types for `places_and_screens` configs to struct. This helps avoid a `to_existing_atom` error that was showing up in certain cases when screens are published.


assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/ConfigureScreensPage.tsx
```diff
@@ -217,6 +217,106 @@ const ConfigurePlaceCard: ComponentType<ConfigurePlaceCardProps> = ({
...
+
+      if (index === -1) {
+        newState.push({
+          screen_id: screenID,
+          is_deleted: true,
+          app_id: "gl_eink_v2",
+          app_params: screen.app_params,
+        });
+      } else {
+        newState[index].is_deleted = true;
```
jzimbel-mbta, line 244: `[...prevState]` creates only a shallow copy of `prevState`, so this statement will mutate the original.

For example:
```ts
const listOfObjs = [{is_deleted: false}, {is_deleted: false}, {is_deleted: false}];
const shallowCopy = [...listOfObjs];
shallowCopy[1].is_deleted = true;

console.log(shallowCopy);
// [{is_deleted: false}, {is_deleted: true}, {is_deleted: false}]
​
console.log(listOfObjs);
// [{is_deleted: false}, {is_deleted: true}, {is_deleted: false}]
```

https://react.dev/reference/react/useState#setstate
> If you pass a function as nextState, it will be treated as an updater function. **It must be pure**, should take the pending state as its only argument, and should return the next state.

This callback can be made pure in the following way:
```ts
(prevState) => {
  const index = newState.findIndex((screen) => screen.screen_id === screenID);

  if (index === -1) {
    const newDeletedScreen = {
      screen_id: screenID,
      is_deleted: true,
      app_id: "gl_eink_v2",
      app_params: screen.app_params,
    };

    return [...prevState, newDeletedScreen];
  } else {
    return [
      ...prevState.slice(0, i),
      {...prevState[index], is_deleted: true},
      ...prevState.slice(i + 1)
    ];
  }
}
```

Most of the other updaters passed to `set*` functions below are also impure—please adjust them to avoid mutating the original state values.

---

Side note: I realize it's ridiculously cumbersome to do immutable updates like this in JS. I bet there's a nice library out there to make the process easier, will ask around if other teams are using something like that.

jzimbel-mbta (changes requested):
Partial review--the JS changes have some impure state updater functions that need to be fixed.

Continuing on reviewing the elixir side while you address those.

assets/js/components/Dashboard/AlertBanner.tsx
```diff
@@ -20,8 +20,9 @@ const AlertBanner: ComponentType<AlertBannerProps> = ({
   isDone,
   queueExpiration,
 }: AlertBannerProps) => {
+  const pathname = useLocation().pathname.replace(/\//g, "");
```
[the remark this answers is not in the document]
cmaddox5: I remember having a reason when I originally put this in `Sidebar.tsx` during the initial implementation, but I can't for the life of me think of a good reason now. I know that `useLocation().pathname` is everything after the app URL (so like `/configure-screens/gl-eink`), but there's no reason to strip out `/` when I'm using `includes` anyway. I'll remove it before I fix the other issues here and make sure everything looks right.

assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/ConfigureScreensPage.tsx
cmaddox5: Wow, very interesting. I definitely need to pay more attention to things like this. Will address these now.

cmaddox5: I think I tackled them all. Did a lot of poking around to make sure nothing changed functionally, and everything is acting the same as before. Please do flag any I may have missed.

assets/js/components/Dashboard/AlertBanner.tsx
```diff
@@ -21,7 +21,8 @@ const AlertBanner: ComponentType<AlertBannerProps> = ({
   queueExpiration,
 }: AlertBannerProps) => {
   const { bannerAlert } = useScreenplayContext();
-  if (!bannerAlert) return null;
+  if (useLocation().pathname.includes("configure-screens") || !bannerAlert)
+    return null;
```
jzimbel-mbta, line 25: Since this early return results in all of the code past line 25 being conditionally evaluated, the `useParams` on line 30 violates the "don't call hooks conditionally" rule<sup>1</sup>. Could you either:
1. move `const params = useParams();` up before this `if` statement, or
2. move this conditional up to the parent component, so that this component is simply not mounted if it shouldn't render anything. E.g.
   ```ts
   // in Dashboard component
   const { bannerAlert } = useScreenplayContext();
   const pathname = useLocation().pathname;
   const showBanner = bannerAlert && !pathname.includes("configure-screens");

   // in JSX returned by component
   {showBanner && <AlertBanner ... />}
   ```

(2) is preferable, since (1) leaves the potential for future bugs if a new hook call is added to this component later on.

<sup>1</sup> [Can you early return with React hooks?](https://stackoverflow.com/questions/54938236/can-you-early-return-with-react-hooks)

assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/ConfigureScreensPage.tsx
```diff
@@ -190,25 +193,16 @@ const ConfigurePlaceCard: ComponentType<ConfigurePlaceCardProps> = ({
...
-        newState[place.id].existing_pending_screens = existingScreensToArray(
-          existingPendingScreens
-        );
-      } else {
-        newState[place.id] = {
+      return {
+        ...placesAndScreens,
+        [place.id]: {
           updated_pending_screens: updatedPendingScreens,
           new_pending_screens: newScreens,
```
jzimbel-mbta, line 200: Are there any other fields on `placesAndScreens[place.id]` that need to be preserved? In the updated state, this replaces `placesAndScreens[place.id]` with a new object with keys `updated_pending_screens`, `new_pending_screens`, `existing_pending_screens`.

If any existing fields besides those need to be preserved, then this should be
```suggestion
        [place.id]: {
          ...placesAndScreens[place.id],
          updated_pending_screens: updatedPendingScreens,
          new_pending_screens: newScreens,
```

lib/screenplay/config/places_and_screens.ex
```diff
@@ -1,15 +1,60 @@
 defmodule Screenplay.Config.PlaceAndScreens do
-  @moduledoc false
+  @moduledoc """
+  Module used to defined struct for screen configs stored in places_and_screens.json.
```
jzimbel-mbta, line 3:
```suggestion
  Module used to define struct for screen configs stored in places_and_screens.json.
```
Also in the 2 sub-modules' `@moduledoc`s

jzimbel-mbta (approved):
Good to go, just double checking on a couple things.

---

Separately: I found out from [this slack thread](https://mbta.slack.com/archives/GP5QJM98A/p1712001384948279) that lodash (which Screenplay already uses) provides a module with a bunch of pure utility functions, perfect for this. So sorry for making you fill your code with `...`s!

Feel free to leave this as-is, and in the future I'll keep `lodash/fp` in mind.

assets/js/components/Dashboard/AlertBanner.tsx
cmaddox5: Good catch. Totally missed that. The `Sidebar` wasn't violating the rule, but I moved that conditional to `Dashboard` so the conditional rendering is done at the same place.

PR #263, opened 2024-03-15
feat: Screens Admin auth

cmaddox5:
**Asana task**: ad-hoc

In this PR, we ensure users are in the new auth group, `screens-admin`, before accessing either `/pending` or `/configure-screens`. An unauthorized user will not see the buttons in the sidebar and will be redirected to `/dashboard` when trying to access the URLs directly.

### DIFF
--- a/lib/screenplay_web/plugs/metadata.ex
+++ b/lib/screenplay_web/plugs/metadata.ex
@@ -32,12 +32,12 @@ defmodule ScreenplayWeb.Plugs.Metadata do
     |> assign(:alerts_ui_url, Application.get_env(:screenplay, :alerts_ui_url))
     |> assign(:screens_url, Application.get_env(:screenplay, :screens_url))
     |> assign(:signs_ui_url, Application.get_env(:screenplay, :signs_ui_url))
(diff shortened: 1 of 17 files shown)


lib/screenplay_web/plugs/metadata.ex
```diff
@@ -33,13 +33,19 @@ defmodule ScreenplayWeb.Plugs.Metadata do
     |> assign(:screens_url, Application.get_env(:screenplay, :screens_url))
     |> assign(:signs_ui_url, Application.get_env(:screenplay, :signs_ui_url))
     |> assign(:is_admin, admin?(conn))
```
hannahpurcell, line 35: If you're open to it, mind changing the references to `is_admin` into `is_emergency_admin` for clarity?

hannahpurcell (approved):
Great! One request to extend your rename work (since I see you renamed the `authenticated_emergency_admin` group) but not blocking

lib/screenplay_web/plugs/metadata.ex
cmaddox5: Sure thing!

PR #265, opened 2024-03-19
feat: Automated Testing for OFM Takeover Tool

cmaddox5:
**Asana task**: [OFM Takeover periodic tests/monitoring](https://app.asana.com/0/1164574158255689/1206301091214212/f)

Due to the Takeover Tool issues we've had in the last 12 months, this automated test will ensure that directory structure is what we expect and that we can upload/delete images to the test folder OFM provided. This test folder is the only folder that images can be uploaded to in this job. OFM confirmed that this folder is not configured for any screen so we don't have to worry about impacting live screens.


lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -0,0 +1,137 @@
...
+      sftp_client_module().disconnect(conn)
+    end
+  end
+
+  # sobelow_skip ["Traversal.FileModule"]
+  defp test_creating_and_removing_images(conn) do
+    local_image_path = Path.join(:code.priv_dir(:screenplay), "takeover_test.png")
+    local_image_data = File.read!(local_image_path)
+
+    Enum.each(["Portrait", "Landscape"], fn orientation ->
```
digitalcora, line 36: We use these specific directory names both here and I think twice in `all_directories_exist` — could we extract them to constants? It's a small thing, but I'd want to make sure that if Outfront decides to rename these, we can change the name in one place and be confident we got everything.

```elixir
@landscape_dir "Landscape"
@portrait_dir "Portrait"
```

Following that train of thought... are these also hard-coded in the "real" tool?

lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -0,0 +1,137 @@
...
+        Logger.error(message)
+        Sentry.capture_message(message, level: "error")
+    after
+      sftp_client_module().disconnect(conn)
+    end
+  end
+
+  # sobelow_skip ["Traversal.FileModule"]
+  defp test_creating_and_removing_images(conn) do
+    local_image_path = Path.join(:code.priv_dir(:screenplay), "takeover_test.png")
```
digitalcora, line 33: I think we could compile this image data into the app as a module attribute, and then the app wouldn't compile if the image wasn't present as expected (instead of crashing later when the job runs). Freehanded so idk if this exact code works:

```elixir
@test_image :code.priv_dir(:screenplay) |> Path.join("takeover_test.png") |> File.read!
```

Technically we should also specify this as an [`@external_resource`](https://hexdocs.pm/elixir/Module.html#module-external_resource) of the module, though it's unlikely to change much.

assets/js/components/OutfrontTakeoverTool/OutfrontTakeoverTool.tsx
```diff
@@ -56,14 +67,27 @@ class OutfrontTakeoverTool extends React.Component<
         confirmJSX: <></>,
         onSubmit: this.toggleModal,
       },
+      stationScreenOrientationList: {
+        red: [],
+        orange: [],
+        blue: [],
+        silver: [],
+        green: [],
```
digitalcora, line 75: What's the reason we have to specify this both here and on L171?

lib/screenplay/outfront/sftp.ex
```diff
@@ -96,6 +96,10 @@ defmodule Screenplay.Outfront.SFTP do
     sftp_client_module().disconnect(conn)
   end
 
+  def get_outfront_directory_for_station(station) do
+    Map.get(@stations_to_outfront_directories, station)
```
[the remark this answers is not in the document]
cmaddox5: Yep, totally. Was able to move those to the app config and removed the old module attribute.

.gitignore
```diff
@@ -34,8 +34,12 @@ npm-debug.log
...
 
-# All files in /priv/ are going to be for local development
+# All config files in /priv/ are going to be for local development
 # so do not need to be tracked.
-/priv/*
+/priv/alerts.json
+/priv/place_descriptions.json
+/priv/places_and_screens.json
+/priv/screen_locations.json
+/priv/screen_locations.json
```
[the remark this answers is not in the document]
cmaddox5: Great idea. I'll move all that around and change the code and README setup instructions.

lib/screenplay_web/controllers/outfront_takeover_tool/page_controller.ex
```diff
@@ -21,4 +21,17 @@ defmodule ScreenplayWeb.OutfrontTakeoverTool.PageController do
   def takeover_redirect(conn, _params) do
     redirect(conn, to: ~p"/emergency-takeover")
   end
+
+  def station_screen_orientation_list(conn, _params) do
+    station_list_json =
+      Application.get_env(:screenplay, :station_screen_orientation_list, %{
```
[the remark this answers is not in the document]
cmaddox5: Good call. Default was definitely overly cautious.

lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -0,0 +1,137 @@
...
+    portrait_stations = Enum.filter(station_screen_orientation_list, & &1.portrait)
+    landscape_stations = Enum.filter(station_screen_orientation_list, & &1.landscape)
+
+    log_missing_dirs(portrait_dirs, portrait_stations, "Portrait")
+    log_missing_dirs(landscape_dirs, landscape_stations, "Landscape")
+  end
+
+  defp log_missing_dirs(sftp_dirs, stations, orientation) do
+    Enum.each(stations, fn station ->
+      %{name: station_name} = station
```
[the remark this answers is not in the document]
cmaddox5: Oops. I did that with this [anti-pattern](https://hexdocs.pm/elixir/1.16.1/code-anti-patterns.html#complex-extractions-in-clauses) in mind, but I actually misunderstood the main concern of the anti-pattern. Will get this changed.

lib/screenplay/jobs/takeover_tool_testing_job.ex
cmaddox5: They are hard-coded, yes. OFM's SFTP directory structure is `./Landscape/...` and `./Portrait/...`.

assets/js/components/OutfrontTakeoverTool/OutfrontTakeoverTool.tsx
cmaddox5: There isn't one. I think this is leftover from when I was passing this state down via `props` instead of using the context. It will still need a default of `{}`, but no need to provide the full object shape in it.

lib/screenplay/jobs/takeover_tool_testing_job.ex
cmaddox5: I think I misunderstood what you meant by "real" tool for some reason. They were indeed hardcoded over there are well. I made those module constants as well, but am not a fan of having two sets of constants that refer to the same thing. What would be the best way for these two modules to share these constants? App config maybe?

lib/screenplay/jobs/takeover_tool_testing_job.ex
cmaddox5: Hmm, that's a cool idea. Love it.

lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -0,0 +1,137 @@
...
+          "[takeover_tool_testing sftp_connection_error] missing #{orientation} directory for station #{station_name}"
+
+        Logger.error(message)
+        Sentry.capture_message(message, level: "error")
+      end
+    end)
+  end
+
+  defp sftp_client_module do
+    Application.get_env(:screenplay, :sftp_client_module)
```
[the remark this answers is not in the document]
cmaddox5: Yep, definitely could be compile-time.

lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -0,0 +1,137 @@
...
+    local_image_path = Path.join(:code.priv_dir(:screenplay), "takeover_test.png")
+    local_image_data = File.read!(local_image_path)
+
+    Enum.each(["Portrait", "Landscape"], fn orientation ->
+      write_image(conn, orientation, local_image_data)
+      delete_image(conn, orientation)
+    end)
+  end
+
+  defp start_connection do
```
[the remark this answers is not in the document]
cmaddox5: I added a new function `SFTP.run/1` to help with this. Is this what you had in mind?

mix.exs
```diff
@@ -57,7 +57,8 @@ defmodule Screenplay.MixProject do
       {:lcov_ex, "~> 0.2", only: [:dev, :test], runtime: false},
       {:sobelow, "~> 0.8", only: :dev},
       {:sentry, "~> 8.0"},
-      {:stream_data, "~> 0.5", only: :test}
+      {:stream_data, "~> 0.5", only: :test},
+      {:quantum, "~> 3.0"}
```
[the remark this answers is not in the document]
cmaddox5: It currently has a single instance. And honestly, I think it will only have a single instance for a long time. The userbase for Screenplay is very small so traffic should always be light enough so ECS does not scale us to use another task. In the next feature for Screenplay, we will likely be using a psql DB. When that happens, we could consider switching to [Oban](https://github.com/sorentwo/oban) for job processing. It has a feature that prevents duplicate jobs from running.

lib/screenplay_web/router.ex
```diff
@@ -94,7 +94,7 @@ defmodule ScreenplayWeb.Router do
     get("/active_alerts", AlertController, :active_alerts)
     get("/past_alerts", AlertController, :past_alerts)
 
-    get("/station_screen_orientation_list", PageController, :station_screen_orientation_list)
+    get("/outfront_takeover_tool_screens", PageController, :outfront_takeover_tool_screens)
```
digitalcora, line 97: It looks like this is already in a `ScreenplayWeb.OutfrontTakeoverTool` scope, so is it redundant to specify `outfront_takeover_tool` in the path / action name? (though confusingly it shares an `/api` path prefix with other, non-OFTT routes...)

lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -32,11 +33,8 @@ defmodule Screenplay.Jobs.TakeoverToolTestingJob do
 
   # sobelow_skip ["Traversal.FileModule"]
```
digitalcora, line 27: I think we no longer need this skip, since we're not doing the file reading at runtime.

lib/screenplay/jobs/takeover_tool_testing_job.ex
```diff
@@ -14,6 +14,7 @@ defmodule Screenplay.Jobs.TakeoverToolTestingJob do
   @landscape_dir "Landscape"
   @portrait_dir "Portrait"
   @test_image :screenplay |> :code.priv_dir() |> Path.join("takeover_test.png") |> File.read!()
+  @sftp_client_module Application.compile_env(:screenplay, :sftp_client_module)
```
digitalcora, line 18: If this is not allowed to be nil, we should use the shoutier `compile_env!`

lib/screenplay/outfront/sftp.ex
```diff
@@ -8,6 +8,7 @@ defmodule Screenplay.Outfront.SFTP do
   @landscape_dir "Landscape"
   @portrait_dir "Portrait"
   @retries 3
+  @sftp_client_module Application.compile_env(:screenplay, :sftp_client_module)
```
digitalcora, line 12: ...and here

lib/screenplay/outfront/sftp.ex
```diff
@@ -11,29 +11,38 @@ defmodule Screenplay.Outfront.SFTP do
...
+  def run(work_fn) do
     conn = start_connection()
 
-    Enum.each(stations, fn station ->
-      write_image(conn, station, @portrait_dir, portrait_png)
-      write_image(conn, station, @landscape_dir, landscape_png)
-    end)
+    try do
+      work_fn.(conn)
+    rescue
```
digitalcora, line 19: I'm wondering about the purpose of this `rescue`. Is there a reason we shouldn't either "let it crash" or return an error to the caller? Generally, it's a good idea to avoid silently swallowing errors (from the caller's perspective, even if the error is logged somewhere as a side effect).

lib/screenplay/outfront/sftp.ex
```diff
@@ -11,29 +11,38 @@ defmodule Screenplay.Outfront.SFTP do
...
-    @sftp_client_module.disconnect(conn)
+    run(fn conn ->
+      Enum.each(
+        stations,
+        fn station ->
+          delete_image(conn, station, @portrait_dir)
+          delete_image(conn, station, @landscape_dir)
+        end
+      )
+    end)
```
digitalcora, line 45: Nice, it looks like we're gaining some better cleanup of the connection by wrapping these ✨

lib/screenplay/jobs/takeover_tool_testing_job.ex
digitalcora: Maybe that new config file you extracted, since we are already storing some other directory names in there? They could then still be module constants, just loaded with `compile_env!`.

lib/screenplay/outfront/sftp.ex
```diff
@@ -2,102 +2,58 @@ defmodule Screenplay.Outfront.SFTP do
   @moduledoc """
   This module handles the CRUD functions to the SFTP server
   """
+  @dialyzer {:no_match, start_connection: 1, write_image: 5, delete_image: 4}
```
digitalcora, line 5: What were the errors here? I don't have this repo set up on my machine yet >.>

digitalcora (commented):
Looking good, just a few minor things I noticed. Can we test this out in a real deployed environment? (by changing the Quantum conditional to `dev-green` maybe)

lib/screenplay/outfront/sftp.ex
cmaddox5: `Pattern will never match` errors for the error cases in these functions. The `FakeSFTPClient` doesn't really any errors so when dialyzer complains while that is the module configured for local development.

digitalcora: Hmm... I wonder if there's a way to fix that while preserving the warnings for the case where we actually do forget to handle a match. If we explicitly spec the function in the SFTP client as returning `:ok | {:error, ...}`, does that then result in an overspec error?

lib/screenplay/outfront/sftp.ex
cmaddox5: The main purpose was to log to Sentry. The big gotcha that I missed is we don't have Sentry logging for server errors at the moment 🙃 . That's not great cause that was what I assumed we would use for getting notified of these issues... I'll remove the `rescue` regardless because if we set up Sentry correctly, error logs will show up there automatically.

digitalcora: Good thing we tested it 😅 Will you set that up as part of this task? Seems pretty essential to the "feature".

cmaddox5: Yeah, I think it's important enough to be in scope for this task. Will get started on that and make sure it's working before merging this branch.

lib/screenplay/outfront/sftp.ex
cmaddox5: Unfortunately, dialyzer seems to ignore the spec and looks at what the function actually returns. So I would either need to fake an error return for these or ignore the warnings.

digitalcora: > Unfortunately, dialyzer seems to ignore the spec and looks at what the function actually returns

Well, normally I'd say that's _fortunate_ behavior, but 😅

I wonder why this hasn't been a problem in the past when I've had fake versions of services that e.g. always returned `:ok`. I think the setup I've used was where the real and fake versions shared a `@behaviour`? But anyway, I don't think it's super important to get to the bottom of this right now unless we've got extra time.

lib/screenplay/outfront/sftp.ex
cmaddox5: Looks like I was fooled by how Screenplay was set up in Sentry. The platform selected in the Project Settings was React so I assumed it was only set up for the frontend. We do have the necessary back configuration for server errors as well. Only thing missing was a separate DSN for dev-green (was still using the dev DSN. PR for that change [here](https://github.com/mbta/devops/pull/1639)). I changed the platform and project name in Sentry so it is set up the way Screens is (which we seem happy with so a good standard for Screenplay).

digitalcora: Wait, it needs a separate DSN? I thought the environment name was just configured in the app and sent along with error reports. I don't remember having to set up more than one DSN per project.

cmaddox5: Very interesting. For whatever reason, I thought that each environment needed their own DSN when I initially set up Sentry for our apps. You're right though, we can use the `environment_name` to identify environment and set up alerts based on that. I closed that devops PR. Not going to mess with existing DSNs for now.

lib/screenplay/outfront/sftp.ex
cmaddox5: Funnily enough, I did try using a `behaviour` and it still yelled about it. I wonder if it's because the "real" module we are using is not ours. I'm not entirely sure, but I'm not super worried about ignoring the warnings specifically for these functions.

digitalcora (approved):
Presuming this has been tested in a deployed environment, lgtm!

PR #266, opened 2024-03-25
fix: Missing compile-time env

cmaddox5:
**Asana task**: ad-hoc

I thought I tested the new automated job in a deployed environment, but I guess not. ~~Now that `sftp_client_module` is needed in module attributes, it can't live in `runtime.exs`. I moved it to `config.exs` instead.~~ Keeping it in the runtime config, but changed from a module attribute back to a helper function.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -24,7 +24,6 @@ sftp_client_module =
 config :screenplay,
   config_fetcher: Screenplay.Config.S3Fetch,
   config_s3_bucket: "mbta-ctd-config",
-  sftp_client_module: sftp_client_module,
   record_sentry: false
 
 # Include 2 logger backends
--- a/config/config.exs
+++ b/config/config.exs
@@ -15,9 +15,16 @@ config :screenplay, ScreenplayWeb.Endpoint,
   pubsub_server: Screenplay.PubSub,
   live_view: [signing_salt: "vSiyKz7D"]
 
+sftp_client_module =
+  case System.get_env("SFTP_SERVER") do
+    "outfront" -> SFTPClient
+    _ -> Screenplay.Outfront.FakeSFTPClient
+  end
+
 config :screenplay,
   config_fetcher: Screenplay.Config.S3Fetch,
   config_s3_bucket: "mbta-ctd-config",
+  sftp_client_module: sftp_client_module,
   record_sentry: false
 
 # Include 2 logger backends
--- a/config/config.exs
+++ b/config/config.exs
@@ -15,12 +15,6 @@ config :screenplay, ScreenplayWeb.Endpoint,
   pubsub_server: Screenplay.PubSub,
   live_view: [signing_salt: "vSiyKz7D"]
 
-sftp_client_module =
-  case System.get_env("SFTP_SERVER") do
-    "outfront" -> SFTPClient
-    _ -> Screenplay.Outfront.FakeSFTPClient
-  end
-
 config :screenplay,
   config_fetcher: Screenplay.Config.S3Fetch,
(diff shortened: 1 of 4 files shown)


config/config.exs
```diff
@@ -15,9 +15,16 @@ config :screenplay, ScreenplayWeb.Endpoint,
   pubsub_server: Screenplay.PubSub,
   live_view: [signing_salt: "vSiyKz7D"]
 
+sftp_client_module =
+  case System.get_env("SFTP_SERVER") do
```
digitalcora, line 19: I didn't realize we were setting the real/fake client module in this way. If we move this to `config`, won't it no longer work as expected, since it is now reading the environment variable at build time instead of once it starts up on the server? Is there a reason to use an env var for this rather than e.g. setting it to use the fake client in `test.exs` and the real one otherwise?

cmaddox5: Not 100% on the OG reason, but I think it's because prod is the only environment that uses the real `SFTPClient` module. Dev and dev-green use the fake one (unless we decide to hook up the logic in those environments to use the new test folder we write to in the new job). I feel like needing a separate variable for that is a bit overkill considering we have access to `ENVIRONMENT_NAME`. Should I move this back to `runtime.exs` and go back to using a helper function in `sftp.ex` and `takeover_tool_testing_job.ex` to get the correct client module?

digitalcora: For the moment, to get this to a working state since it's already on the main branch, that might be the best approach.

For the future, I like your idea of writing to the test directory in the other environments. The more closely aligned the `dev` environments are with the code that runs on `prod`, the better, otherwise we can't have as much confidence that if something works in one place it will work in the other.

config/runtime.exs
```diff
@@ -6,14 +6,6 @@ config :screenplay, ScreenplayWeb.Endpoint,
...
   secret_key_base: System.get_env("SECRET_KEY_BASE")
 
-sftp_client_module =
-  case System.get_env("SFTP_SERVER") do
-    "outfront" -> SFTPClient
-    _ -> Screenplay.Outfront.FakeSFTPClient
-  end
-
-env = System.get_env("ENVIRONMENT_NAME")
-
```
digitalcora, line 16: The changes in this file now appear to be a no-op — just moving things around. Can we revert it back to the original content?

digitalcora (approved):
I think this should work, but let's make sure to test it on dev-green first by temporarily swapping out the `env` check so it uses "real" SFTP there.

One more thing I just thought of: Since we have the SFTP-module-per-environment thing, do we also need the `prod` conditional on configuring the Quantum job? Again, thinking about minimizing differences between environments — maybe we could unknowingly introduce a bug in the job and not learn about it until the changes were already merged and it crashed in `dev`.

cmaddox5:
> do we also need the prod conditional on configuring the Quantum job?
I think we should avoid the job running anywhere the fake module is used because the tests will never pass there. The `list_dir` calls in the job expect a specific output or it logs it as an error. I agree that there's no reason this job shouldn't run on those environments once we set them up to point only at the test directory, but until then it seems unnecessary. Open to pushback though.

I will make sure to test in dev-green like you suggested. I think I'll make a temporary `job_sftp_client_module` just so there is no way the tool can upload to live screens while we are testing. I suppose I should also tweak the schedule so it runs more often and we can see an actual test run.

config/runtime.exs
cmaddox5: I did make a small change to the `case` so it looks at `env` now. There was no reason I should have moved the code to a different part of the file. Moving it back to where it was.

PR #267, opened 2024-03-25
feat: Permanent configuration

cmaddox5:
Final PR needed to put the new feature in `main`.


panentheos (approved):
Approving without additional review, since all changes against this branch went through individual reviews.

PR #269, opened 2024-03-27
Pending Page empty state

cmaddox5:
**Asana task**: ad-hoc

Found [this](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc/Permanent-Configuration?type=design&node-id=2497-153679&mode=design&t=yo4LzwAKiS91RyuQ-4) in the Figma so added it in real quick. Also improved the `page-content__body` bg color a bit so it always extends to bottom of screen.

### DIFF
--- a/assets/css/screenplay-container.scss
+++ b/assets/css/screenplay-container.scss
@@ -7,7 +7,6 @@
     flex-grow: 1;
     display: flex;
     flex-direction: column;
-    height: 100%;
     overflow-y: auto;
     overflow-x: hidden;
     color: $text-primary;
@@ -23,7 +22,7 @@
       background-color: $cool-gray-15;
       padding: 24px 20px;
       border-radius: 16px 0px 0px 0px;
-      flex-grow: 1;
+      height: 100%;
     }
   }
 }
--- a/assets/js/components/Dashboard/PendingScreensPage.tsx
+++ b/assets/js/components/Dashboard/PendingScreensPage.tsx
@@ -121,10 +121,12 @@ const PendingScreensPage: ComponentType = () => {
 
   useEffect(fetchData, []);
 
-  return (
-    <div className="pending-screens-page">
-      <div className="page-content__header">Pending</div>
-      <div className="page-content__body" style={{ color: "white" }}>
+  const screens = Object.entries(existingScreens);
+  let layout;
+
+  if (screens.length) {
+    layout = (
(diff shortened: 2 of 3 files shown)


assets/js/components/Dashboard/PendingScreensPage.tsx
```diff
@@ -154,6 +156,19 @@ const PendingScreensPage: ComponentType = () => {
             }
           )}
         </Accordion>
+      </>
+    );
+  } else {
+    layout = (
+      <div className="no-screens-text h5">There are no pending screens.</div>
```
digitalcora, line 163: I'm not familiar with the design system here, curious why `<div class="h5">` and not `<h5>`?

assets/css/screenplay-container.scss
```diff
@@ -23,7 +22,7 @@
       background-color: $cool-gray-15;
       padding: 24px 20px;
       border-radius: 16px 0px 0px 0px;
-      flex-grow: 1;
+      height: 100%;
```
digitalcora, line 25: To be more explicit I think this could be `flex-basis: 100%`. (`flex-basis` defaults to `auto` which means to use any defined width/height as the basis, so it's one less layer of indirection)

digitalcora (approved):
Some non-blocking comments!

assets/js/components/Dashboard/PendingScreensPage.tsx
cmaddox5: I want to say the original reason was due to how Mary defined the `h5` style, and that style didn't match up with the properties of an `h5` element. In hindsight, we probably should have pushed back a bit and suggested better naming. Typography can be found [here](https://www.figma.com/file/FKWRVrSdoMjZOLorhuykvc/Permanent-Configuration?type=design&node-id=915-31883&mode=design&t=FnHKPCmAspNX09TL-4).

assets/css/screenplay-container.scss
cmaddox5: So I was able to get this to work, but it requires other styling changes in other files (specifically needs changes to `.pending-screens-page`). I think I'll leave it alone for now unless you prefer making those changes.

digitalcora: Interesting. Well, if this works, I don't see anything wrong with it.

PR #270, opened 2024-03-27
feat: Sentry replay integration

cmaddox5:
**Asana task**: ad-hoc

I noticed that Screenplay has some errors that give very little info. Setting up source mapping will help us see readable stacktraces in Sentry, but it requires a secret that I'll need to track down. Replay setup was super easy so didn't think we needed to wait. 

Sample replay can be found [here](https://mbtace.sentry.io/replays/c0c3526f8db84b9e96d0b9002ff0d58a/?environment=dev-green&project=6061951&query=&referrer=%2Freplays%2F&statsPeriod=7d&yAxis=count%28%29). Note that all text is masked by default. We can use our own masking by adding the CSS class `sentry-mask` to any element, but I'm not sure that's 100% necessary at this point.

### DIFF
--- a/assets/js/app.tsx
+++ b/assets/js/app.tsx
@@ -35,6 +35,9 @@ if (sentryDsn) {
   Sentry.init({
     dsn: sentryDsn,
     environment: environment,
+    integrations: [Sentry.replayIntegration()],
+    replaysSessionSampleRate: 0.1,
+    replaysOnErrorSampleRate: 1.0,
   });
 
   if (username) {
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -2013,62 +2013,94 @@
         "@babel/runtime": "^7.6.2"
       }
     },
-    "@sentry/browser": {
-      "version": "7.14.0",
-      "resolved": "https://registry.npmjs.org/@sentry/browser/-/browser-7.14.0.tgz",
-      "integrity": "sha512-AdLmqeOXvCVYgJAgMUUby+TRh+yIeZO16NPpZWQPGggXIjnhSzoN4liyXJvQ7Mhm326GboFUKjQwqpCEviQcyg==",
+    "@sentry-internal/feedback": {
+      "version": "7.108.0",
+      "resolved": "https://registry.npmjs.org/@sentry-internal/feedback/-/feedback-7.108.0.tgz",
+      "integrity": "sha512-8JcgZEnk1uWrXJhsd3iRvFtEiVeaWOEhN0NZwhwQXHfvODqep6JtrkY1yCIyxbpA37aZmrPc2JhyotRERGfUjg==",
       "requires": {
-        "@sentry/core": "7.14.0",
-        "@sentry/types": "7.14.0",
-        "@sentry/utils": "7.14.0",
-        "tslib": "^1.9.3"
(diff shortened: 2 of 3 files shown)


digitalcora:
Is this _only_ a stopgap for not having sourcemaps, i.e. are we going to turn it off once we have those? If this is going to be a permanent thing I think it deserves some wider team awareness/discussion.

assets/js/app.tsx
```diff
@@ -35,6 +35,9 @@ if (sentryDsn) {
   Sentry.init({
     dsn: sentryDsn,
     environment: environment,
+    integrations: [Sentry.replayIntegration()],
+    replaysSessionSampleRate: 0.1,
```
digitalcora, line 39: Is there value in recording sessions without errors? If we're currently just adding this in support of error tracking, maybe we don't need to.

cmaddox5: Yeah I suppose not. I just remembered that we are using Fullstory for full sessions. Devs would only need sessions for errors.

digitalcora: Hm, I wonder if this would be capable of replacing current usage of Fullstory, or if they have different limitations.

cmaddox5:
> Is this _only_ a stopgap for not having sourcemaps, i.e. are we going to turn it off once we have those? If this is going to be a permanent thing I think it deserves some wider team awareness/discussion.

I envisioned this being an additional tool that we would use with sourcemaps, so not something we would remove later. Happy to bring this up to the team though.

assets/js/app.tsx
cmaddox5: The issue with Fullstory is the very limited number of seats. AFAIK, only PM and Design get them. I wasn't even given a temporary one while I was implementing FS. I know there is a FS/Sentry integration, but that gives Sentry issues a link to the FS replay where the error occurred. Because we don't have seats, that doesn't really benefit us.

digitalcora: > I wonder if this would be capable of replacing current usage of Fullstory

In other words, using Sentry's own session recording for whatever tasks we currently use Fullstory for (not the other way around). But this would require that A) Sentry's session recording has similar capabilities as Fullstory, and B) everyone currently using Fullstory on the project could be given Sentry access.

jzimbel-mbta (approved):
I'm also wondering whether we should drop Fullstory assuming Sentry Replay provides essentially the same features, but this change seems fine to merge while we come up with an answer to that question.

If there isn't one already, could you write up a short task to consider dropping Fullstory? Or, start a slack thread if that seems more expedient.

cmaddox5:
@jzimbel-mbta @digitalcora Something I forgot to consider is pricing. I know the department has hit the Sentry plan maximum for errors. Replays included are even less. Our plan only allows for 500 replays a month. It doesn't look like any teams are using the feature, but I can see that maxing out if other teams start using them. I'm personally leaning more towards asking for a team Fullstory account and using that integration. Other teams have that, so there is precedent for asking for a shared account. I think I'll go down that path for a bit before merging this.

digitalcora:
500 a month _might_ be somewhat useful for errors, but definitely wouldn't let us use Sentry as a complete Fullstory replacement. It probably is worth asking about the Fullstory account.

PR #273, opened 2024-04-01
Auth needs for PA Message Creation

cmaddox5:
**Asana task**: [Permission needs for Screenplay](https://app.asana.com/0/1185117109217413/1206790423002398/f)

In the new PA Messaging integration, only users with the `pa-message-admin` can access the tool. This PR adds all of the necessary auth needed to check permissions when implementation for the feature starts.

### DIFF
--- a/lib/screenplay_web/auth_manager/auth_manager.ex
+++ b/lib/screenplay_web/auth_manager/auth_manager.ex
@@ -3,10 +3,12 @@ defmodule ScreenplayWeb.AuthManager do
 
   use Guardian, otp_app: :screenplay
 
-  @type access_level :: :none | :read_only | :emergency_admin | :screens_config_admin
+  @type access_level ::
+          :none | :read_only | :emergency_admin | :screens_config_admin | :pa_message_admin
 
   @screenplay_emergency_admin_role "screenplay-emergency-admin"
   @screens_admin "screens-admin"
+  @pa_message_admin "pa-message-admin"
 
   @spec subject_for_token(
           resource :: Guardian.Token.resource(),
@@ -30,6 +32,7 @@ defmodule ScreenplayWeb.AuthManager do
       []
       |> append_if(@screenplay_emergency_admin_role in roles, :emergency_admin)
       |> append_if(@screens_admin in roles, :screens_admin)
+      |> append_if(@pa_message_admin in roles, :pa_message_admin)
 
     if access_levels == [] do
       [:read_only]
--- a/lib/screenplay_web/auth_manager/auth_manager.ex
+++ b/lib/screenplay_web/auth_manager/auth_manager.ex
@@ -3,8 +3,7 @@ defmodule ScreenplayWeb.AuthManager do
 
   use Guardian, otp_app: :screenplay
 
-  @type access_level ::
-          :none | :read_only | :emergency_admin | :screens_config_admin | :pa_message_admin
+  @type access_level :: :emergency_admin | :screens_config_admin | :pa_message_admin
 
   @roles %{
     "screenplay-emergency-admin" => :emergency_admin,
@@ -30,16 +29,10 @@ defmodule ScreenplayWeb.AuthManager do
 
   @spec claims_access_level(Guardian.Token.claims()) :: list(access_level())
   def claims_access_level(%{"roles" => roles}) when not is_nil(roles) do
-    access_levels = Enum.map(roles, &Map.get(@roles, &1)) |> Enum.reject(&is_nil/1)
-
-    if access_levels == [] do
-      [:read_only]
-    else
-      access_levels
-    end
+    Enum.map(roles, &Map.get(@roles, &1)) |> Enum.reject(&is_nil/1)
   end
 
   def claims_access_level(_claims) do
-    [:read_only]
+    []
   end
 end
--- a/lib/screenplay_web/auth_manager/auth_manager.ex
+++ b/lib/screenplay_web/auth_manager/auth_manager.ex
@@ -3,7 +3,7 @@ defmodule ScreenplayWeb.AuthManager do
 
   use Guardian, otp_app: :screenplay
 
-  @type access_level :: :none | :read_only | :emergency_admin | :screens_config_admmin
+  @type access_level :: :none | :read_only | :emergency_admin | :screens_config_admin
 
   @screenplay_admin_role "screenplay-emergency-admin"
   @screens_admin "screens-admin"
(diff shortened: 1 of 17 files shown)


lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -24,21 +26,26 @@ defmodule ScreenplayWeb.AuthManager do
...
-      @screens_admin in roles ->
-        :screens_config_admmin
-
-      true ->
-        :read_only
+    access_levels =
+      []
+      |> append_if(@screenplay_emergency_admin_role in roles, :emergency_admin)
+      |> append_if(@screens_admin in roles, :screens_admin)
+      |> append_if(@pa_message_admin in roles, :pa_message_admin)
```
digitalcora, line 35: It might be easier to see the correspondence between keycloak roles and, uh, auth manager roles? if we declared them all together like:

```elixir
@roles %{"screenplay-emergency-admin" => :emergency_admin, ...}
```

Then determining the access levels would be a breezier:

```elixir
Enum.map(roles, &Map.get(@roles, &1)) |> Enum.reject(&is_nil/1)
```

lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -24,21 +26,26 @@ defmodule ScreenplayWeb.AuthManager do
 
   def resource_from_claims(_), do: {:error, :invalid_claims}
 
-  @spec claims_access_level(Guardian.Token.claims()) :: access_level()
+  @spec claims_access_level(Guardian.Token.claims()) :: list(access_level())
```
digitalcora, line 29: Are there no tests for this module? I guess not, otherwise they would have had to change here. Seems like something we should have given the importance of getting access control right. (I think Guardian has some helper functions for testing with claims)

lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -24,21 +26,26 @@ defmodule ScreenplayWeb.AuthManager do
...
-      true ->
-        :read_only
+    access_levels =
+      []
+      |> append_if(@screenplay_emergency_admin_role in roles, :emergency_admin)
+      |> append_if(@screens_admin in roles, :screens_admin)
+      |> append_if(@pa_message_admin in roles, :pa_message_admin)
+
+    if access_levels == [] do
+      [:read_only]
```
digitalcora, line 38: Is there a reason you don't also have `read_only` if you have some higher level of access?

lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -3,10 +3,12 @@ defmodule ScreenplayWeb.AuthManager do
 
   use Guardian, otp_app: :screenplay
 
-  @type access_level :: :none | :read_only | :emergency_admin | :screens_config_admmin
+  @type access_level ::
+          :none | :read_only | :emergency_admin | :screens_config_admin | :pa_message_admin
 
-  @screenplay_admin_role "screenplay-emergency-admin"
+  @screenplay_emergency_admin_role "screenplay-emergency-admin"
   @screens_admin "screens-admin"
```
digitalcora, line 10: What's the distinction between an admin and an emergency-admin? I notice in our fake dev permissions, we only grant the "emergency" one.

lib/screenplay_web/auth_manager/ensure_pa_message_admin.ex
```diff
@@ -0,0 +1,22 @@
...
+  """
+
+  import Plug.Conn, only: [halt: 1]
+  import Phoenix.Controller, only: [redirect: 2]
+
+  def init(options), do: options
+
+  def call(conn, _opts) do
+    with claims <- Guardian.Plug.current_claims(conn),
+         true <- :pa_message_admin in ScreenplayWeb.AuthManager.claims_access_level(claims) do
```
digitalcora, line 13: We seem to be doing this same logic in quite a lot of places: here, the other `ensure` plug, twice in `metadata`... can we do it once in a common plug (maybe `metadata`) and put the "finished" list of roles in the conn? The `ensure` plugs would still need to exist, but could just check if e.g. `:pa_message_admin in roles`.

lib/screenplay_web/router.ex
```diff
@@ -63,6 +67,17 @@ defmodule ScreenplayWeb.Router do
     get("/unauthorized", UnauthorizedController, :index)
   end
 
+  scope "/", ScreenplayWeb do
+    pipe_through([
+      :redirect_prod_http,
+      :browser,
+      :auth,
+      :ensure_auth,
+      :ensure_pa_message_admin,
```
digitalcora, line 76: `pipe_through` accepts plug modules as well as pipelines; we technically don't need to define a pipeline if it only consists of one module. (Though maybe this could be better for readability?)

lib/screenplay_web/auth_manager/auth_manager.ex
cmaddox5: It might actually be something we can remove and have the function return an empty list instead. The `:read_only` permission is not used directly because we enforce auth for all endpoints. If you make it through that, read only is implied.

lib/screenplay_web/auth_manager/auth_manager.ex
cmaddox5: `emergency-admin` is the permission needed to access the takeover tool. `screens-admin` will be used to access the new Permanent Configuration tool (still in a feature branch). I'm not 100% sure why it is in this file already considering there is a feature branch for that. It must have been done before realizing we should be merging to that branch instead. It will be added to the fake permissions when that branch is merged into main.

lib/screenplay_web/auth_manager/auth_manager.ex
cmaddox5: Woah, that is slick. Definitely using that.

lib/screenplay_web/auth_manager/auth_manager.ex
cmaddox5: There are tests for the ensure plug (realizing now that the new one does not have tests) but not this module specifically. Those tests use tags to add user and roles and then asserts on the return of the plug itself. Do you think that is enough?

lib/screenplay_web/auth_manager/ensure_pa_message_admin.ex
cmaddox5: Good call. Will move some code around to make this work.

lib/screenplay_web/router.ex
cmaddox5: Hmm, I could go either way on this. A part of me does feel like a list of only atoms is more readable than a list of atoms and module names. Unless you feel strongly one way or the other, I think I'll keep it as-is.

test/support/conn_case.ex
```diff
@@ -55,6 +57,21 @@ defmodule ScreenplayWeb.ConnCase do
...
+          user = "test_user"
+
+          conn =
+            Phoenix.ConnTest.build_conn()
+            |> Plug.Test.init_test_session(%{})
+            |> Guardian.Plug.sign_in(ScreenplayWeb.AuthManager, user, %{
+              "roles" => ["pa-message-admin"]
+            })
+            |> Plug.Conn.put_session(:username, user)
+            |> assign(:roles, [:pa_message_admin])
```
digitalcora, line 74: Can we instead run this through the plug that assigns the roles? Under the new scheme, the `"roles" => ["pa-message-admin"]` above isn't actually doing anything. (Maybe a case for putting that role-assignment in its own plug rather than lumped in with `metadata`.)

Alternatively, we could just delete the `"roles" => ...`. But then that code path is completely untested.

digitalcora (approved):
Looking good to me, just one lingering question, but I'll pre-approve and be fine with whatever the resolution is, as long as it's been considered.

PR #277, opened 2024-04-04
fix: User Session TTL

cmaddox5:
**Asana task**: ad-hoc

Was finally able to discover the cause of some [Sentry errors](https://mbtace.sentry.io/issues/5102241771/events/a8c2509681314021885c47f6181fef43/) with terribly unhelpful stack traces. Basically, our session TTL was only 5 minutes. Because we have an auth enforced endpoint that is requested every 4 seconds, this throws a ton of errors if the tab is left open. Users should be able to leave the app open without worrying about this every 5 minutes. Bumped it up to 30 days (same as Glides) to help with this. 

In the future, we will consider storing refresh tokens in the DB and refreshing sessions before they expire.

### DIFF
--- a/lib/screenplay_web/controllers/auth_controller.ex
+++ b/lib/screenplay_web/controllers/auth_controller.ex
@@ -6,8 +6,7 @@ defmodule ScreenplayWeb.AuthController do
   def callback(conn = %{assigns: %{ueberauth_auth: auth}}, _params) do
     username = auth.uid
     name = auth.info.name
-    expiration = auth.credentials.expires_at
-    current_time = System.system_time(:second)
+    session_ttl_hours = 24 * 30
 
     keycloak_client_id =
       get_in(Application.get_env(:ueberauth_oidcc, :providers), [:keycloak, :client_id])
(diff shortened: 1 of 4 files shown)


lib/screenplay_web/controllers/auth_controller.ex
```diff
@@ -23,7 +22,7 @@ defmodule ScreenplayWeb.AuthController do
       ScreenplayWeb.AuthManager,
       username,
       %{roles: roles},
-      ttl: {expiration - current_time, :seconds}
+      ttl: {session_ttl_hours, :hours}
```
jzimbel-mbta, line 25: Does this change mean that every time this `callback` action gets hit (via `/:provider/callback`, looks like) the session TTL gets reset to 30 days? As opposed to the previous code, which appeared to have the session expire at the date/time specified by `auth.credentials.expires_at` no matter how many times this action ran.

(When does this action run? Is it only once per session, or every time a request that requires authorization comes in?)

jzimbel-mbta (approved):
I think this is good, but I am very unfamiliar with how exactly this auth & user session stuff works so please excuse the million questions

lib/screenplay_web/controllers/auth_controller.ex
cmaddox5: Gonna be honest that I'm still a little fuzzy on this. I believe this code only runs when no session exists at all. So basically the first time you log in to the app. It takes `ttl` amount of time for the session to expire. After it expires, you would need to refresh the session (can only be done via a page refresh atm, but could use a refresh token to handle this in the future) and run this action again. Bumping it out to 30 days means that we can run this action a lot less frequently because it takes much longer for the session to expire.

PR #290, opened 2024-04-05
feat: In Memory Alerts

cmaddox5:
**Asana task**: [Background process for Alert polling](https://app.asana.com/0/1185117109217413/1206860070503125/f)

This PR adds two new background processes that fetch and store alerts. `fetcher.ex` is in charge of getting the most recent set of alerts from the API every 4 seconds and updating the configured store with these alerts. `store.ex` manages and ets table with the alerts fetched from the API. Each update call removes all old alerts and inserts the new alerts. This will help keep expired alerts out of the table. The table comes with two getter functions: one for all alerts, another for a specific alert given an ID.

**Update
Consolidated down to one module for both fetching and storing. All functionality remained the same, but only one `GenServer` is needed now. Also added a config value to prevent process from starting in tests.


lib/screenplay/alerts/cache/store.ex
```diff
@@ -0,0 +1,64 @@
+defmodule Screenplay.Alerts.Cache.Store do
```
panentheos, line 1: Do we need a second module/process for this? It's not storing any state or providing significant parallelism, so it seems like we could inline this code into the other module to simplify things.

cmaddox5: That's a fair point. Managed to consolidate while keeping test coverage at 100%.

lib/screenplay/alerts/cache.ex
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  @impl true
+  def handle_info(:fetch, state = {get_json_fn, update_interval_ms}) do
+    case Alert.fetch(get_json_fn) do
+      {:ok, alerts} ->
+        alerts_to_insert =
+          Enum.reduce(alerts, [], fn alert, acc ->
+            [{alert.id, alert} | acc]
+          end)
```
PaulJKim, line 63: Couldn't this be done a little simpler with `Enum.map`?

PaulJKim (commented):
Couple of minor questions but otherwise LGTM. Also, it might be helpful for posterity's sake to update the PR description since one of the mentioned modules was removed

PaulJKim (approved):

lib/screenplay/alerts/cache.ex
```diff
@@ -0,0 +1,80 @@
...
+  """
+  @spec alerts :: list(Alert.t())
+  def alerts do
+    :ets.select(:alerts, [{{:_, :"$1"}, [], [:"$1"]}])
+  end
+
+  # Server
+  @impl true
+  def init(opts) do
+    _ = :ets.new(:alerts, [:protected, :named_table, read_concurrency: true])
```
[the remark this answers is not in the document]
cmaddox5: Hmm, I thought leaving that out would cause a dialyzer issue, but I'm wrong about that. I've seen errors from dialyzer before when you don't use the return of a function (was `Logger` related stuff so I shouldn't have assumed this was the same). Will remove it.

lib/screenplay/alerts/cache.ex
cmaddox5: Oops, yeah definitely. Thanks.

PR #294, opened 2024-04-08
Update webpack

cmaddox5:
**Asana task**: ad-hoc

The main goal of this PR is to update Node and webpack. Node is being bumped from `14.17.3` to `v18.20.1` and webpack `v4.41.5` to `v5.91.0`. Lots of other dependencies needed to be updated to make this happen. I only updated packages that threw errors or warnings after upgrading Node and webpack. 

Some notable changes (besides Node and webpack):
- `node-sass` has been deprecated. Will now use `sass` as per package's recommendation.
- `optimize-css-assets-webpack-plugin` is incompatible with webpack 5. Replaced it with `css-minimizer-webpack-plugin`
- webpack cli no longer uses arg `--watch-stdin`. Now use a combo of `--watch` and `--watch-options-stdin`. 

Local installation should just need a fresh `asdf install` and `npm install`. I recommend removing your `node_modules` and `package-lock.json` before installing.

---
  - https://app.asana.com/0/0/1207059891028337


cmaddox5:
This diff is huge, but 99% of the diff is the lock file.

assets/package.json
```diff
@@ -53,19 +53,17 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^28.1.0",
     "jest-environment-jsdom": "^28.1.0",
-    "mini-css-extract-plugin": "^0.9.0",
-    "node-sass": "^4.13.1",
-    "optimize-css-assets-webpack-plugin": "^5.0.1",
+    "mini-css-extract-plugin": "^2.8.1",
+    "sass": "^1.69.5",
+    "sass-loader": "^13.3.2",
```
digitalcora, line 58: Some of these appear to have escaped from their formerly-sorted order (S here is followed by P). Can we do a quick re-sort?

assets/js/components/OutfrontTakeoverTool/AlertWizard/AlertWizard.tsx
```diff
@@ -14,8 +14,7 @@ import WizardSidebar from "./WizardSidebar";
 import { svgLongSide, svgScale, svgShortSide } from "../../../constants/misc";
 import { matchStation } from "../../../util";
 
-import parseISO from "date-fns/parseISO";
-import differenceInHours from "date-fns/differenceInHours";
+import { differenceInHours, parseISO } from "date-fns";
```
digitalcora, line 17: Earlier in this PR `date-fns` is removed as an explicit dependency, but here we are indeed depending on it in our own code, so probably we should add it back?

.tool-versions
```diff
@@ -1,3 +1,3 @@
 erlang 26.2
 elixir 1.15.7-otp-26
-nodejs 14.17.3
\ No newline at end of file
+nodejs 18.20.1
```
digitalcora, line 3: Why Node 18 in particular? Is this the minimum version for Webpack 5?

digitalcora (commented):
I'm glad this doesn't appear to be too big of a change, for when we eventually do the same thing on Screens!

.tool-versions
cmaddox5: Webpack is ok but the latest versions of our plugins support a minimum of 18. My first reaction was to bring down the plugin versions, but I saw that node 16 is actually not [maintained anymore](https://endoflife.date/nodejs). Even 18 is in maintenance mode, but we have a year until we are EOL for that version.

assets/js/components/OutfrontTakeoverTool/AlertWizard/AlertWizard.tsx
cmaddox5: Hmm, very surprised I didn't see any errors after removing it. It must be indirectly imported by something else. Will definitely add it back.

digitalcora: It definitely is, since in the commit where it was removed, `package-lock` has a diff where it updates the version of the package rather than removing it (indicating our version constraint was more restrictive than whatever package has it as a dependency, I suppose).

.tool-versions
digitalcora: I see. I had some issues on `screens` trying to update Node higher than 16, but I think they were related to Webpack, which I didn't want to tackle at that time just to get the repo working on Apple Silicon.

digitalcora (approved):
Onward and upward!

PR #295, opened 2024-04-09
Update Typescript

cmaddox5:
**Asana task**: ad-hoc

Updates to TS and packages that depend on TS. Required an additional `jest` update and a `transform` change to webpack config.

---
  - https://app.asana.com/0/0/1207059891028343


jzimbel-mbta (approved):
LGTM, should be good to merge assuming CI passes!

PR #303, opened 2024-04-17
fix: Ignore local dependencies in dependabot.

cmaddox5:
**Asana task**: ad-hoc

Dependabot isn't successfully retrieving updates because it doesn't know where to look for Phoenix upgrades. These are local dependencies anyway so can just ignore them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,5 +16,7 @@ updates:
       time: "10:00"
     open-pull-requests-limit: 5
     ignore:
+      - dependency-name: "phoenix"
+      - dependency-name: "phoenix_html"
       - dependency-name: "*"
         update-types: ["version-update:semver-minor"]


jzimbel-mbta (approved):
Approved, but I'm a little confused: How are phoenix and phoenix_html local dependencies? They're definitely used in deployment environments.

And is there any reason why it has trouble finding new versions of those two packages? Seems like it doesn't have trouble with other elixir packages.

cmaddox5:
> How are phoenix and phoenix_html local dependencies?

@jzimbel-mbta Maybe local dependency isn't the right phrase to use. Phoenix is very much an external dependency we manage in `mix.exs`. npm is just using that same dependency. So we only have to keep the elixir dependency up to date instead of maintaining an npm package AND an elixir package. You can find that [here](https://github.com/mbta/screenplay/blob/666c5fd29159735e8fa934d55a77ed117acbdedd/assets/package.json#L22).

The problem is that Dependabot looks at npm packages and mix deps separately. That means that the Dependabot npm check does not know how to find something in `deps` because it does not run `mix deps.get` before doing it's check. Since we are only maintaining a single `phoenix` dependency, we can skip checking it in npm and just let the mix check keep it up to date.

PR #305, opened 2024-04-19
fix: Allow screen IDs to have any format

jzimbel-mbta:
**Asana task**: Permanent Config GL E-Ink QA (Too-strict screen ID validations)

The validation code enforced a particular format on the GL e-ink screen IDs, but we don't need that.

---

Notes for posterity:
- The other validation issue found during eng QA, where duplicate screen IDs were not always caught, was actually just one more symptom of the auth issue that Christian fixed in #277.
- I originally went down a rabbit hole trying to fix a large number of issues with state value mutation on the client, but ended up dropping that work in favor of a more surgical fix. In case we do end up finding problems caused by the mutations in the future, I've pushed my WIP, semi-broken code to the [jz/fix-pending-screen-validation](https://github.com/mbta/screenplay/compare/permanent-configuration...jz/fix-pending-screen-validation) as a possible reference or starting point for a more complete fix.


cmaddox5 (commented):
May be a design/product question, but if we aren't doing these checks anymore should we bother pre-populating the field? When a new row is added, it will have `EIG-` in the box. Is that necessary?

jzimbel-mbta:
@cmaddox5 I think it's good as-is. That's a convenience thing since in most cases the IDs _will_ follow that format. We just don't want to stop people from using a different format if they feel the need.

cmaddox5 (approved):

PR #307, opened 2024-04-22
fix: Remove Phoenix path dependencies

cmaddox5:
**Asana task**: ad-hoc

These path dependencies are giving Dependabot trouble. We don't even use them anyway so I think removing them is the solution.


digitalcora:
As I mentioned in the Slack thread, I think we _are_ at least using Phoenix sockets for live-reload in development; I wouldn't want to break that in the name of getting Dependabot working. Maybe we can kick this out into a new task if figuring out a real solution is non-trivial?

cmaddox5:
If I'm reading the code correctly, we are not using the live reload feature in any of our React code. [Here](https://github.com/mbta/screenplay/blob/c6b700981fa67d6863b0fb2407aba5d384d60e89/config/dev.exs#L50) is the configuration for live reload and it has the default patterns for phoenix views and templates, but our client code is housed in a folder not listed in the list. Phoenix looks to have switched from an opt-out strategy of implementing Phoenix sockets in client code to opt-in by not including these JS files in the new app template. I also tested out these changes locally and TS updates still trigger a reload and Elixir updates are still recompiled when necessary.

digitalcora:
> I also tested out these changes locally and TS updates still trigger a reload and Elixir updates are still recompiled when necessary.

I expect updates to server-side templates aren't live-reloaded, though? Maybe that's okay if the frontend is mostly React and we don't expect to be changing server templates often... still feels like a bit of a sledgehammer for swatting this fly.

digitalcora (approved):
We should also remove the entries added in #303.

PR #309, opened 2024-04-23
fix: duplicate ID edge case

cmaddox5:
**Asana task**: ad-hoc

Found a weird edge case with validation logic. If you are deleting an existing pending screen from one place and add it as a new pending screen at another, the ID was flagged as duplicate. This is not true because the screen is being replaced after all updates are done.


lib/screenplay/config/permanent_config.ex
```diff
@@ -175,15 +175,27 @@ defmodule Screenplay.Config.PermanentConfig do
...
+
+    deleted_ids = Enum.map(deleted_screens, & &1["screen_id"])
+    updated_ids = Enum.map(updated_screens, & &1["new_id"])
+
+    # Do not call an ID a duplicate if we know we are going to delete it.
+    existing_after_deletes =
+      existing_screens
+      |> Enum.reject(fn {screen_id, _} -> screen_id in deleted_ids end)
+      |> Enum.into(%{})
+      |> Map.keys()
```
panentheos, line 196: nit: This could be:
```suggestion
      |> Enum.map(&elem(&1, 0))
      |> Enum.reject(&(&1 in deleted_ids))
```

lib/screenplay/config/permanent_config.ex
```diff
@@ -311,8 +314,14 @@ defmodule Screenplay.Config.PermanentConfig do
...
   defp update_existing_pending_screens(place_id, platform_ids, updated_pending_screens, acc) do
     Enum.reduce(updated_pending_screens, acc, fn
-      %{"is_deleted" => true, "screen_id" => screen_id}, acc ->
-        Map.delete(acc, screen_id)
+      %{"is_deleted" => true}, acc ->
+        # Only delete screen if "is_deleted" is true AND platform is at current place.
+        # If a screen ID is deleted at one place and moved to another,
+        # we need to make sure we are removing the correct screen.
+        Map.reject(acc, fn
+          %{alerts: %Alerts{stop_id: platform_id}} -> platform_id in platform_ids
```
panentheos, line 322: This doesn't look right, the function given to `Map.reject` should take a 2-tuple of key and value. Also, shouldn't we need to use `screen_id` somehow?

lib/screenplay/config/permanent_config.ex
```diff
@@ -196,15 +208,6 @@ defmodule Screenplay.Config.PermanentConfig do
       else: {:error, {:duplicate_screen_ids, duplicate_screen_ids}}
   end
 
-  defp get_updated_pending_screen_ids(updated_pending_screens) do
-    Enum.filter(updated_pending_screens, fn screen ->
-      not is_nil(screen) and Map.has_key?(screen, "new_id")
```
panentheos, line 201: I think we still need to keep this filter somewhere in the pipeline, because otherwise when we map out `"new_id"` above, we could get nils, which would mess up the later logic.

panentheos (commented):
How about framing the algorithm something like this: `existing_ids - deleted_ids - old_changed_ids + new_changed_ids + new_ids`

lib/screenplay/config/permanent_config.ex
cmaddox5: Good catch. I was actually embarrassingly off. Fixed it so it now looks for platform ID and screen ID.

lib/screenplay/config/permanent_config.ex
cmaddox5: I made an attempt to improve the algorithm in the function that was calling this. It should more clearly group IDs based on the type of change. That should allow a guarantee that we won't see `nil` values when performing our checks.

lib/screenplay/config/permanent_config.ex
```diff
@@ -175,15 +175,33 @@ defmodule Screenplay.Config.PermanentConfig do
...
+                                      {new_screen_ids, changed_screen_ids, deleted_screen_ids} ->
+        cond do
+          screen["is_deleted"] ->
+            {new_screen_ids, changed_screen_ids, [screen["screen_id"] | deleted_screen_ids]}
+
+          screen["new_id"] != nil and screen["screen_id"] == nil ->
+            {[screen["new_id"] | new_screen_ids], changed_screen_ids, deleted_screen_ids}
+
+          # screen is existing but had its ID changed.
+          true ->
```
panentheos, line 196: What about screens that are being updated, but without an id change? Do they come with a "new_id" value, such that this clause would handle them?

lib/screenplay/config/permanent_config.ex
```diff
@@ -175,15 +175,33 @@ defmodule Screenplay.Config.PermanentConfig do
...
+            {[screen["new_id"] | new_screen_ids], [screen | changed_screen_ids],
```
panentheos, line 197: This should be
```suggestion
            {[screen["new_id"] | new_screen_ids], [screen["screen_id"] | changed_screen_ids],
```

lib/screenplay/config/permanent_config.ex
cmaddox5: Hmm that's a good point. They do not have a `new_id` value so would always hit the `true` condition here. But they honestly could be filtered out of any of the three lists here because they will exist in the `Map.keys(existing_screens)` call a few lines down. I should probably have the `true` condition return the `acc` as-is and create one more condition for if `new_id` and `screen_id` both have non-nil values. Then the `cond` will cover all of our bases:
1. deleted screens
2. new screens
3. existing screens that have a new ID
4. any other screen that we can ignore and will be in `existing_screens`.

PR #310, opened 2024-04-23
fix: State issue on refresh

cmaddox5:
**Asana task**: ad-hoc

When a refresh occurs on the `Edit Pending` page, the places render correctly but the hook does not fetch existing screens correctly. I think this is due to hook execution timing. The hook fetching existing screens has already run by the time the state has been updated with the places from `location.state`. Making `selectedPlaces.length` part of the dep array fixes this. #308 removes the `Remove Location` button so this won't happen unless the page is being loaded. The Edit page state will continue to look at location state for the `selectedPlaces` and the New page state will look at the list passed in through props.


digitalcora:
I'm very much a stranger to the context here — is the expectation basically that `selectedPlaces` may be null or empty on initial render, then some time later it will be populated with values once, and not change again after that? If so I think using `.length` as the dependency makes sense, though maybe we should add a comment that this is the expectation. If not (that is, if the content of `selectedPlaces` may change other than going from "none" to "some"), I'd think this could still have a bug in some cases, namely when the contents change but the length stays the same.

cmaddox5:
> I'm very much a stranger to the context here — is the expectation basically that `selectedPlaces` may be null or empty on initial render, then some time later it will be populated with values once, and not change again after that? If so I think using `.length` as the dependency makes sense, though maybe we should add a comment that this is the expectation. If not (that is, if the content of `selectedPlaces` may change other than going from "none" to "some"), I'd think this could still have a bug in some cases, namely when the contents change but the length stays the same.

Yep, your comment is correct. There are two ways to get to this page
1. from the previous step in the workflow. In this way, `selectedPlaces` is populated before the first render because it comes from `props`.
2. From the `Edit Pending` button on the Pending tab. In this way, `selectedPlaces` is populated as soon as [this](https://github.com/mbta/screenplay/blob/8646d4c55e7e396bd782493c8eda7ff28673d700/assets/js/components/Dashboard/PermanentConfiguration/Workflows/GlEink/GlEinkWorkflow.tsx#L54) setter is executed which will be after first render.

In both scenarios, `selectedPlaces` is guaranteed to not change again after #308 is merged. The user will have to leave the page to change the list. The comment is a good idea, will do that now.

digitalcora (approved):
🚀

PR #321, opened 2024-04-23
PA message table migration and postgres configuration

PaulJKim:
**Asana task**: [DB schema definition](https://app.asana.com/0/1185117109217413/1206860070503119/f)

This PR adds the migration for setting up the PA message table and a couple indexes.

It also adds the configuration needed for Screenplay to use Postgres.


.envrc.template
```diff
@@ -15,3 +15,9 @@
 # export SCREENS_URL=
 # export SIGNS_UI_URL=
 # export ALERTS_UI_URL=
+
+## Postgres configuration: username, password, and hostname
+## * Your local Postgres server should go here
+# export DATABASE_USERNAME=
```
cmaddox5, line 21: The variable name in TF is `DATABASE_USER` so we should use that here.

config/config.exs
```diff
@@ -7,6 +7,17 @@
...
+config :screenplay, Screenplay.Repo,
+  database: "screenplay_dev",
+  username: System.get_env("DATABASE_USER", ""),
+  password: System.get_env("DATABASE_PASSWORD", ""),
+  hostname: System.get_env("DATABASE_HOST", "localhost"),
+  port: System.get_env("DATABASE_PORT", "5432") |> String.to_integer(),
+  show_sensitive_data_on_connection_error: true,
+  backoff_min: 5_000
+
+config :screenplay, ecto_repos: [Screenplay.Repo]
```
cmaddox5, line 19: Just curious, what does this do? I'm not seeing it used anywhere else, but maybe Ecto needs it?

cmaddox5 (changes requested):
Looks and works great! Just have a couple of small changes I found while trying to run locally. Also, can this be deployed to dev-green before merging just to make sure everything works correctly in a deployed environment? The migration we have looks good so shouldn't have to worry about needing to roll it back before merging.

cmaddox5:
I just noticed the failed CI and didn't even think about there needing to be DB configuration there. Seems like we'll need to get a "local" DB configured in order to build the app in CI.

lib/screenplay/migrate.ex
```diff
@@ -0,0 +1,53 @@
...
+  def init(opts) do
+    Logger.info("#{__MODULE__} synchronous migrations starting")
+    Keyword.get(opts, :sync_migrate_fn, &default_migrate_fn/1).("migrations")
+
+    Logger.info("#{__MODULE__} synchronous migrations finished")
+    {:ok, opts, {:continue, :async_migrations}}
+  end
+
+  @impl GenServer
+  def handle_continue(:async_migrations, opts) do
```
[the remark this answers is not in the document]
PaulJKim: Sounds good to me!

config/config.exs
PaulJKim: Yeah, it's just a piece of configuration that Ecto needs, according to the documentation

> There's one final bit of configuration that we'll need to add ourselves, since the generator does not add it. Underneath the configuration in config/config.exs, add this line:
>
> `config :friends, ecto_repos: [Friends.Repo]`
>
> This tells our application about the repo, which will allow us to run commands such as [mix ecto.create](https://hexdocs.pm/ecto/Mix.Tasks.Ecto.Create.html) very soon.

config/config.exs
```diff
@@ -7,6 +7,17 @@
 # General application configuration
 import Config
 
+config :screenplay, Screenplay.Repo,
```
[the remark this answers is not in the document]
PaulJKim: These actually will get overwritten when deployed to an environment via the code we borrowed from Skate in `repo.ex`

PaulJKim:
> Looks and works great! Just have a couple of small changes I found while trying to run locally. Also, can this be deployed to dev-green before merging just to make sure everything works correctly in a deployed environment? The migration we have looks good so shouldn't have to worry about needing to roll it back before merging.

@cmaddox5 Alright I think I've fixed the CI issue!

cmaddox5 (approved):
Awesome! Good to go!

cmaddox5:
@PaulJKim Also, can you change the base to the feature branch? Just want to keep everything there until ready to ship to prod.

PR #324, opened 2024-04-29
Doc/template updates for initial setup

digitalcora:
Getting on board with another app!

👉 [Rendered version of updated README](https://github.com/mbta/screenplay/blob/eb3266a04ef81fb5677d15193a330c5eb225b60a/README.md)

* Adds a version of the same `pull_configs.sh` script added to Screens earlier. This replaces a couple "ask another engineer for this file" setup steps, and removes the requirement to have AWS credentials set up in environment variables for `fetch_places_and_screens.exs`.

* Provides (what I hope are) reasonable default values for URLs and other non-sensitive items in `.envrc.template`. This cuts down the values that must be filled in as part of setup to just a V3 API key.

* Remove the root `package.json` that "proxied" some `npm` commands to the one in `assets`. It doesn't appear this was widely used, since using it generates a `package-lock.json` that was never committed. (Because of this, I'm making the change on the assumption that it's not disrupting anyone's workflow — please let me know if I'm mistaken!)

* Went ahead and assumed the reader is going to use `direnv` to manage the several environment variables this app needs, as we already had the template/gitignore set up, and this greatly simplifies the written steps. We still have the comment in the template explaining that you can also directly `source` this file if not using `direnv`.

---
  - https://app.asana.com/0/0/1207194888793668


cmaddox5 (approved):
Awesome, thanks for doing this!

PR #327, opened 2024-04-30
tweak: Workflow entry change

cmaddox5:
**Asana task**: [Perm. Config. Workflow Entry](https://app.asana.com/0/1185117109217413/1207059005820020/f)

This PR removes the Configure tab from the sidebar and adds a new button to the Pending page for navigation to the beginning of the Perm Config workflow. No functionality has been changed. The beginning of the workflow can still be accessed via URL.

### DIFF
--- a/assets/css/dashboard/pending-screens-page.scss
+++ b/assets/css/dashboard/pending-screens-page.scss
@@ -4,7 +4,7 @@
   .add-new-button {
     height: 48px;
     background-color: $button-primary;
-    color: #212529;
+    color: $button-secondary;
   }
 }
 
@@ -78,7 +78,7 @@
 
       .publish-button {
         background-color: $button-primary;
-        color: #212529;
+        color: $button-secondary;
       }
     }
   }


assets/css/dashboard/pending-screens-page.scss
```diff
@@ -1,5 +1,11 @@
 .pending-screens-page {
   height: 100%;
+
+  .add-new-button {
+    height: 48px;
+    background-color: #8ecdff;
+    color: #212529;
```
digitalcora, line 7: Do we have design tokens for things like this? (i.e. variables)

assets/js/components/Dashboard/PendingScreensPage.tsx
```diff
@@ -165,7 +168,21 @@ const PendingScreensPage: ComponentType = () => {
 
   return (
     <div className="pending-screens-page">
-      <div className="page-content__header">Pending</div>
+      <div className="page-content__header">
+        <Container fluid>
+          <Row className="align-items-center text-white">
+            <Col>Pending</Col>
+            <Col className="d-flex justify-content-end">
```
digitalcora, line 175: Assuming the class name `d-flex` applies `display: flex`, does this do anything in this context? The element only has one child (the button), so there's not really anything for flexbox to lay out.

digitalcora (approved):
Just a couple non-blocking questions!

assets/js/components/Dashboard/PendingScreensPage.tsx
panentheos: It looks like it's using `justify-content` to push the button to the right, so it needs to be a flex container for that to work.

digitalcora: Ah, right, I was thinking the column layout was already handling that. Do we have a `justify-self-end` utility? Since this column is already a child of a flex layout (or grid layout? unclear to me what `fluid` actually does).

cmaddox5: `fluid` is something that React bootstrap has that allows you to force a `Container` to use all available horizontal space. Docs are [here](https://react-bootstrap.github.io/docs/layout/grid/#container-1). Without it, it pushes its children towards the center.

Unfortunately, Bootstrap does not have a `justify-self-*` utility, but I think I found a better way to accomplish the styling here. You're right that the `Row` is already flex by default. I can add `justify-content-between` to it and change the responsive breakpoints of the `Col`s to `auto` so they only take up as much space as they need. Then the space will be added between the elements and no `Col` elements need to be flex unnecessarily.

assets/css/dashboard/pending-screens-page.scss
cmaddox5: We did not for this color and we should have. I double checked designs and this color is showing as primary in Figma. This is not what `$button-primary` was set to though so I renamed the current `$button-primary` to `$button-secondary` (which is how we were using it anyway from what I can tell) and added this color as `$button-primary`.

digitalcora: I'm a little confused; from the commit, it looks like you meant _this_ color (`212529`) should be "secondary", and the one above (`8ecdff`) should be "primary". It also looks like this instance of `212529` is still here, rather than being updated to use the "secondary" variable.

cmaddox5: Woops, sorry about that. Forgot to do one more Find All before committing the change.

PR #328, opened 2024-04-30
fix pull_configs script

panentheos:
This adds a clause to the setup script so that it downloads the files if they don't already exist.


digitalcora (approved):
oops! This mistake was copied straight over from `screens`, so I will go fix it there too.

PR #330, opened 2024-04-30
redirect root to /dashboard

panentheos:
**Asana task**: [Revise Screenplay entry URL](https://app.asana.com/0/1185117109217413/1206753590642890/f)

See title.


digitalcora:
Looking at the Asana task:

> we'd like to re-route https://screenplay.mbta.com/ to the main live view screen. "/dashboard" should continue to route there as well for legacy link/bookmark purposes.

This sounds like the desired outcome is that `/` _is_ the dashboard page and `/dashboard` redirects to it (only to support old links). This PR does it the other way around, where `/dashboard` continues to be the canonical path for the dashboard and `/` redirects to it. In this case I don't imagine we're too picky about the specific approach, but I am curious about the reasoning.

panentheos:
This seemed like the more conservative change, since we're already doing a redirect at the root, and this just changes where it's pointing, so there's little risk of something breaking.

digitalcora (approved):
↪️ ✅

PR #331, opened 2024-04-30
rework takeover alert state handling

panentheos:
This simplifies handling of persisted state within the takeover tool, with the initial goal of eliminating spontaneous diffs in the fixture file when running tests locally. Notes:
* Adds a new module to handle test mode specifically, which reads the fixture and does not write anything.
* Moves JSON encoding/decoding steps into the caller, so they're not repeated in each module.
* Removes some hedging around failure cases, since the only caller already does a hard match against a successful return value, meaning we already crash/log if something goes wrong.


panentheos:
Let's go with `!` names, since that communicates the behavior without the added complexity of parsing out error cases.

PR #332, opened 2024-04-30
chore: support Busway v2 screens

digitalcora:
~~There is some existing duplicative logic here around screen types and how certain ones may be displayed differently on the frontend, which I haven't yet attempted to clean up, but I did rename some things to hopefully make the intent clearer.~~

When a config migrated using https://github.com/mbta/screens/pull/2030 is loaded into the updated places-and-screens script, and Screenplay is pointing to an instance of Screens that has the changes in https://github.com/mbta/screens/pull/2039, this branch creates the following result:

Screenshot 2024-04-30 at 5 46 01 PM

Reasons this still looks a bit weird:

* Both the old `solari` and new `busway_v2` screens are displayed at once. In production I don't think this will ever be the case, as one of the two will always be `hidden_from_screenplay`, and there will be a "cutover" at some point. But I'd like to check this assumption before merging.

* The section resizing code on the Screens frontend doesn't actually know how to handle multiple sections yet (nor headers, "Later Departures", etc.), so we only get the first section on the v2 screen.

* The "location name" from `screen_locations.json` for this screen (and others) was set to "Busway". When the name for this category of screen was "Solari", this made sense and resulted in the header being **Solari / Busway**. ~~But now that the category name itself is "Busway", there is some redundancy. We might want to update this file in deployed environments to account for the rename.~~ We'll continue to use "Solari" for now and rename it once the larger naming discussion has happened.

---
  - https://app.asana.com/0/0/1206730847640133

### DIFF
--- a/assets/js/components/Dashboard/ScreenDetail.tsx
+++ b/assets/js/components/Dashboard/ScreenDetail.tsx
@@ -14,11 +14,13 @@ interface ScreenDetailProps {
 }
 
 const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
-  const isSolari = props.screens.every((screen) => screen.type === "solari");
+  const isInlineGroup = props.screens.every((screen) =>
+    ["busway_v2", "solari"].includes(screen.type),
+  );
   const isMultipleScreens = props.screens.length > 1;
 
-  return isSolari ? (
-    <div className="screen-detail__solari-layout">
+  return isInlineGroup ? (
+    <div className="screen-detail__inline-layout">
       {props.screens.map((screens, index) => (
         <ScreenCard
           {...props}
@@ -36,7 +38,9 @@ const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
 const ScreenCard = (props: ScreenDetailProps) => {
   const { screens, eventKey, isMultipleScreens } = props;
   const isPaess = screens.every((screen) => screen.type === "pa_ess");
-  const isSolari = screens.every((screen) => screen.type === "solari");
+  const isInline = screens.every((screen) =>
+    ["busway_v2", "solari"].includes(screen.type),
+  );
   const isTriptych = screens.every((screen) => screen.type === "triptych_v2");
   const paessRouteLetter = screens[0].station_code
     ? screens[0].station_code.charAt(0).toLowerCase()
@@ -94,7 +98,7 @@ const ScreenCard = (props: ScreenDetailProps) => {
       className={classNames("screen-detail__container", {
         [`screen-detail__container--paess screen-detail__container--paess-${paessRouteLetter}`]:
           isPaess,
-        [`screen-detail__container--solari`]: isSolari,
+        [`screen-detail__container--inline`]: isInline,
       })}
       onClick={(e: SyntheticEvent) => e.stopPropagation()}
     >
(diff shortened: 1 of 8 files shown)


assets/js/components/Dashboard/ScreenDetail.tsx
```diff
@@ -12,11 +12,13 @@ interface ScreenDetailProps {
 }
 
 const ScreenDetail = (props: ScreenDetailProps): JSX.Element => {
-  const isSolari = props.screens.every((screen) => screen.type === "solari");
+  const isInlineGroup = props.screens.every((screen) =>
+    ["busway_v2", "solari"].includes(screen.type),
+  );
```
PaulJKim, line 17: it it worth sharing this logic and lines 39-41? As simple as it is.

scripts/build_places.exs
```diff
@@ -1,14 +1,32 @@
...
+
+  Options:
+
+    --env <ENV>
+    Load screen configuration from S3 using the specified Screens environment.
+    Requires ExAWS to have valid credentials, e.g. from `AWS_` environment variables.
+
+    --path <PATH>
+    Load screen configuration from a local file.
+  """)
```
PaulJKim, line 24: This is helpful! Thank you for adding this

PaulJKim (approved):
One minor comment, non-blocking. Otherwise looks great!

digitalcora:
@PaulJKim This has been brought up-to-date with the newly merged Perm Config changes, and I've also improved the duplicative handling of screen types. Would appreciate a second look!

assets/js/util.ts
```diff
@@ -178,22 +178,22 @@ export const placesWithSelectedAlert = (
...
+  "dup_v2",
+  "bus_shelter_v2",
+  "bus_eink",
+  "bus_eink_v2",
+  "gl_eink_single",
+  "gl_eink_double",
+  "gl_eink_v2",
+  "pre_fare_v2",
+  "triptych_v2",
+  "solari",
```
cmaddox5, line 192: Looks like "busway_v2" still needs to be added here.

scripts/build_places.exs
```diff
@@ -80,13 +98,20 @@ add_routes_to_stops = fn
...
 
 # Get live config from S3
-{:ok, file} = get_config.("mbta-ctd-config", "screens/screens-#{environment}.json")
-parsed = Jason.decode!(file)
+config =
+  if environment do
+    {:ok, body} = get_config.("mbta-ctd-config", "screens/screens-#{environment}.json")
+    body
+  else
+    File.read!(local_path)
```
cmaddox5, line 106: This is a great idea.

cmaddox5:
> as one of the two will always be `hidden_from_screenplay`

Yep, this is correct. After Ana officially sets the hardware to look at the v2 URLs, we'll update that field on v1 configs and rerun the script.

> But now that the category name itself is "Busway"

I know the whole app name discussion has been very active recently, but Screenplay is where I see the name working against us. For example, Harvard's screen will now be labelled `Busway / Lobby` which is very confusing.

cmaddox5 (approved):
Just have one minor comment, but this is looking good! Really appreciate the script updates.

assets/js/util.ts
digitalcora: oops, good catch!

digitalcora:
> I know the whole app name discussion has been very active recently, but Screenplay is where I see the name working against us. For example, Harvard's screen will now be labelled `Busway / Lobby` which is very confusing.

Ergh, yeah. On the other hand, I don't want to leave this PR open (and have to keep it up-to-date) until this question is finally resolved. We've already discussed "Multimodal" as an okay-but-not-great name for these screens, so maybe I'll just make it that for now, if I can get Dave's blessing. It will be very easy to change the label later if we want.

PR #333, opened 2024-05-02
fix slack reminder errors when running locally

panentheos:
This fixes the slack reminder logic to correctly suppress reminders when running locally with no `SLACK_WEBHOOK_URL` environment variable set. Note that the default value in `Application.get_env` was not doing anything, since there's always a value present from `config/runtime.exs`, even if it's `nil`.

### DIFF
--- a/lib/screenplay/outfront_takeover_tool/alerts/reminders.ex
+++ b/lib/screenplay/outfront_takeover_tool/alerts/reminders.ex
@@ -17,14 +17,11 @@ defmodule Screenplay.OutfrontTakeoverTool.Alerts.Reminders do
   end
 
   def handle_info(:work, state) do
-    url = Application.get_env(:screenplay, :slack_webhook_url, "")
-
     schedule_work()
 
-    if url == "" do
-      Logger.info("No Slack Webhook URL found")
-    else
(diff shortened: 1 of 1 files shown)


cmaddox5:
> I notice the module doc refers to `#ctd-pio-livepa`

I think this webhook was overlooked when that channel was archived. The webhook still attempts to send messages there. I also don't have access to the channel that replaced it, and honestly these messages would probably be more appropriate in a channel like #screens-team-pios. Luckily this is something we set up just in case OIOs forget to close a takeover alert, but that situation has never happened. I can make a task to change both the Slack app and the code so it sends messages to a different TBD channel.

digitalcora:
> but that situation has never happened.

Hopefully it would have failed noisily if it had ever tried to send a message to the archived channel.

cmaddox5 (approved):
Looks good. One suggestion: would it be useful to conditionally add this job to application startup? The only environment that will ever need the job is prod so it might be worth preventing it from starting in other cases.

digitalcora:
I'd be hesitant about changing the process layout per-environment, because it increases the likelihood that we could accidentally break something and not notice until it surprises us in production.

PR #334, opened 2024-05-02
remove automatic asana status changes

panentheos:
This removes the workflows that update ticket state on PR/merge, in order to allow multiple PRs for a ticket, if desired. Note that this keeps the workflow that updates the PR link on the ticket, so that will continue working. Also fixes the branch ignore  pattern for dependabot branches.


digitalcora (approved):
+1 for deleting code. I'm not even convinced of the value of the other Asana job — I've never used it (I use the Asana-side integration where adding a PR link to a task automatically creates the reverse link in the PR description).

PR #336, opened 2024-05-03
feat: PA Message endpoint

cmaddox5:
This PR adds the endpoint RTS will use to retrieve PA messages from Screenplay. Endpoint is protected by a plug that checks for the correct API key stored for the environment (PR for that is [here](https://github.com/mbta/devops/pull/1756)). I had a slight pivot from the API spec on the full URL, but I updated the doc to reflect the change. Also made a small edit to the README regarding psql connection stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ Screenplay requires Postgres. If you don't already have Postgres installed, and 
 1. Install [`direnv`](https://direnv.net/)
 1. `cp .envrc.template .envrc`
 1. Fill in `API_V3_KEY` with a [V3 API key](https://api-v3.mbta.com/)
+1. Fill in `DATABASE_USER` and `DATABASE_PASSWORD` with the username and password of a DB user configured in your local psql server
 1. `direnv allow`
+1. `mix ecto.create` to stand up DB used by PA Messaging features
 
 Note the various `_URL` values in `.envrc`, which default to the production
 environments of the relevant apps — change these to e.g. point Screenplay to
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -50,7 +50,8 @@ config :screenplay,
   screens_url: System.get_env("SCREENS_URL"),
   signs_ui_url: System.get_env("SIGNS_UI_URL"),
   alerts_ui_url: System.get_env("ALERTS_UI_URL"),
-  fullstory_org_id: System.get_env("FULLSTORY_ORG_ID")
+  fullstory_org_id: System.get_env("FULLSTORY_ORG_ID"),
+  api_key: System.get_env("SCREENPLAY_API_KEY")
 
 if sentry_dsn not in [nil, ""] do
   config :sentry,
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -35,6 +35,9 @@ end
 
 sentry_dsn = System.get_env("SENTRY_DSN")
 
+api_key = System.get_env("SCREENPLAY_API_KEY")
+if api_key in [nil, ""], do: raise("SCREENPLAY_API_KEY key not set")
+
 config :screenplay,
   alerts_s3_path: "screenplay/" <> System.get_env("ALERTS_S3_FILENAME", ""),
   sftp_client_module: sftp_client_module,
@@ -51,7 +54,7 @@ config :screenplay,
   signs_ui_url: System.get_env("SIGNS_UI_URL"),
   alerts_ui_url: System.get_env("ALERTS_UI_URL"),
   fullstory_org_id: System.get_env("FULLSTORY_ORG_ID"),
-  api_key: System.get_env("SCREENPLAY_API_KEY")
+  api_key: api_key
 
 if sentry_dsn not in [nil, ""] do
   config :sentry,
--- a/docs/tech_specs/pa_message_api.md
+++ b/docs/tech_specs/pa_message_api.md
@@ -2,7 +2,7 @@
 
 Lists all PA messages that are currently eligible to play.
 
-**URL** : `/api/active_pa_messages`
+**URL** : `/api/pa_messages/active_pa_messages`
 
 **Method** : `GET`
 
--- a/lib/screenplay_web/plugs/enforce_api_key.ex
+++ b/lib/screenplay_web/plugs/enforce_api_key.ex
@@ -5,13 +5,15 @@ defmodule ScreenplayWeb.Plugs.EnforceApiKey do
 
   import Plug.Conn
 
+  @api_key Application.compile_env!(:screenplay, :api_key)
+
   def init(default), do: default
 
   def call(conn, _default) do
     api_key =
       Enum.find_value(conn.req_headers, fn {key, value} -> if(key == "x-api-key", do: value) end)
 
-    if api_key != Application.get_env(:screenplay, :api_key) do
+    if api_key != @api_key do
       conn |> send_resp(403, "Invalid API key") |> halt()
     else
       conn
--- a/lib/screenplay_web/plugs/enforce_api_key.ex
+++ b/lib/screenplay_web/plugs/enforce_api_key.ex
@@ -1,0 +1,20 @@
+defmodule ScreenplayWeb.Plugs.EnforceApiKey do
+  @moduledoc """
+  Plug used to verify a request contains a valid API key in the x-api-key request header.
+  """
+
+  import Plug.Conn
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    api_key =
+      Enum.find_value(conn.req_headers, fn {key, value} -> if(key == "x-api-key", do: value) end)
+
+    if api_key != Application.get_env(:screenplay, :api_key) do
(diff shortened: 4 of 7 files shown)


docs/tech_specs/pa_message_api.md
```diff
@@ -2,7 +2,7 @@
 
 Lists all PA messages that are currently eligible to play.
 
-**URL** : `/api/active_pa_messages`
+**URL** : `/api/pa_messages/active_pa_messages`
```
digitalcora, line 5: This seems a bit redundant now it has the prefix. Can it be `/api/pa_messages/active`?

Alternatively, just `/api/pa_messages`, unless there's some more general "index" action we should reserve that for. To my recollection, though, "active" PA messages are the only kind we really care about?

lib/screenplay_web/plugs/enforce_api_key.ex
```diff
@@ -0,0 +1,20 @@
...
+  Plug used to verify a request contains a valid API key in the x-api-key request header.
+  """
+
+  import Plug.Conn
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    api_key =
+      Enum.find_value(conn.req_headers, fn {key, value} -> if(key == "x-api-key", do: value) end)
```
digitalcora, line 12: We can use `Plug.Conn.get_req_header/2` here.

lib/screenplay_web/router.ex
```diff
@@ -39,6 +39,10 @@ defmodule ScreenplayWeb.Router do
     plug(ScreenplayWeb.EnsurePaMessageAdmin)
   end
 
+  pipeline :enforce_api_key do
+    plug(ScreenplayWeb.Plugs.EnforceApiKey)
```
digitalcora, line 43: Minor naming thing: The other pipelines and plugs involving auth(entication|orization) all start with `Ensure`. If this is the same kind of thing as those, maybe it could be named `EnsureApiKey`? Or `EnsureApiAuth`, as we are not just ensuring there _is_ an API key in the request, but that it's valid.

config/runtime.exs
```diff
@@ -35,6 +35,9 @@ end
 
 sentry_dsn = System.get_env("SENTRY_DSN")
 
+api_key = System.get_env("SCREENPLAY_API_KEY")
+if config_env() != :test and api_key in [nil, ""], do: raise("SCREENPLAY_API_KEY key not set")
```
digitalcora, line 39: To avoid subtle differences in behavior between environments, I think it'd be better to add a test API key to the CI env and have this always be required.

We could also use `System.fetch_env!` here, though that would allow a set-but-empty API key, if we have reason to believe that could happen unintentionally.

README.md
```diff
@@ -22,7 +22,9 @@ Screenplay requires Postgres. If you don't already have Postgres installed, and
 1. Install [`direnv`](https://direnv.net/)
 1. `cp .envrc.template .envrc`
 1. Fill in `API_V3_KEY` with a [V3 API key](https://api-v3.mbta.com/)
+1. Fill in `DATABASE_USER` and `DATABASE_PASSWORD` with the username and password of a DB user configured in your local psql server
 1. `direnv allow`
+1. `mix ecto.create` to stand up DB used by PA Messaging features
```
digitalcora, line 27: This is before dependencies have been installed, so it wouldn't actually work at this part of the setup. I think it should be down below in "Start the server".

docs/tech_specs/pa_message_api.md
cmaddox5: Hmm, yeah I didn't notice that before but it is definitely redundant. I think your simplified `api/pa_messages` is a good one. I imagine that any feature needing inactive messages will live in Screenplay so wouldn't need this particular API scope to retrieve them.

config/runtime.exs
cmaddox5: That's a good idea. The scenario I imagine for set-but-empty is in local development, and I'm not super concerned with that. TF enforces values to exist in deploy environments.

lib/screenplay_web/plugs/ensure_api_auth.ex
```diff
@@ -0,0 +1,19 @@
...
+  @moduledoc """
+  Plug used to verify a request contains a valid API key in the x-api-key request header.
+  """
+
+  import Plug.Conn
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    api_key = Plug.Conn.get_req_header(conn, "x-api-key")
```
digitalcora, line 11: `get_req_header` returns a list:

```elixir
@spec get_req_header(t(), binary()) :: [binary()]
```

...since some HTTP headers can have multiple values. This won't crash, but also will never allow anything through the plug. We could handle this by matching like `[api_key] = ...`, which should be fine as multiple values being provided for this header would definitely be a "developer error".

Seeing this made me realize there aren't actually any tests for this controller/plug in the PR — if we had one, it would have surfaced this issue right away. Let's add a couple of basic controller tests? (one for getting the `[]` successfully, and one that fails on an invalid API key; we can expand these as we build the real functionality)

cmaddox5: Yikes, sorry about that.

test/screenplay_web/plugs/ensure_api_auth_test.exs
```diff
@@ -0,0 +1,29 @@
+defmodule ScreenplayWeb.EnsureApiAuthTest do
+  use ScreenplayWeb.ConnCase
+
+  describe "init/1" do
+    test "passes options through unchanged" do
+      assert ScreenplayWeb.EnsureScreenplayEmergencyAdminGroup.init([]) == []
```
digitalcora, line 6: Maybe not the module you meant to test 😛

test/support/conn_case.ex
```diff
@@ -74,6 +74,19 @@ defmodule ScreenplayWeb.ConnCase do
...
+        tags[:api_unauthenticated] ->
+          conn =
+            Phoenix.ConnTest.build_conn() |> Plug.Conn.put_req_header("x-api-key", "1234")
+
+          {conn, nil}
+
+        tags[:api_authenticated] ->
+          conn =
+            Phoenix.ConnTest.build_conn()
+            |> Plug.Conn.put_req_header("x-api-key", "local_api_key")
```
digitalcora, line 86: This won't work if someone changes their API key from the default one in `.envrc.template` — we should fetch the real one that's configured. `put_req_header("x-api-key", Application.fetch_env!(...))`

test/support/conn_case.ex
```diff
@@ -74,6 +74,19 @@ defmodule ScreenplayWeb.ConnCase do
 
           {conn, user}
 
+        tags[:api_unauthenticated] ->
```
digitalcora, line 77: Being unauthenticated is sort of the default state — do we need a tag for this? I think in the couple of tests where we specifically test providing an _incorrect_ API key, it would actually make things clearer to put that logic in the tests themselves. (It seems like we'll want the `authenticated` tag though, because we'll have many tests that require an authenticated connection but aren't really intended to test the authentication itself.)

test/screenplay_web/controllers/pa_messages_api_controller_test.exs
```diff
@@ -0,0 +1,25 @@
...
+
+    @tag :api_unauthenticated
+    test "responds 403 if x-api-key does not match app API key", %{conn: conn} do
+      conn = get(conn, "/api/pa_messages")
+
+      assert %{status: 403, halted: true, resp_body: "Invalid API key"} = conn
+    end
+
+    @tag :api_authenticated
+    test "responds list of active messages if x-api-key mathces app API key", %{conn: conn} do
```
digitalcora, line 19: typo: `mathces`

test/screenplay_web/plugs/ensure_api_auth_test.exs
```diff
@@ -0,0 +1,29 @@
...
+
+    @tag :api_unauthenticated
+    test "returns 403 when x-api-key does not match app API key", %{conn: conn} do
+      conn = ScreenplayWeb.Plugs.EnsureApiAuth.call(conn, [])
+
+      assert %{status: 403, halted: true, resp_body: "Invalid API key"} = conn
+    end
+
+    @tag :api_authenticated
+    test "returns conn unchange if x-api-key header matches app API key", %{conn: conn} do
```
digitalcora, line 25: typo: `unchange`

digitalcora (approved):
Looking good, will go ahead and ✅ this with some minor comments!

test/screenplay_web/plugs/ensure_api_auth_test.exs
cmaddox5: 🤦

test/support/conn_case.ex
cmaddox5: Ahh, I did think about that, but was a little too optimistic about the local value not changing from the template.

test/support/conn_case.ex
cmaddox5: Good point. Went a little tag happy.

PR #342, opened 2024-05-13
feat: Active Message Endpoint

cmaddox5:
This PR adds the business logic for endpoint that retrieves active messages. 

A message is considered active if either of the following are true:
1. Row does not have a `start_time_utc` or `end_time_utc` and `alert_id` matches any alert in the cache
2. `now` is between `start_time_utc` and `end_time_utc` and current day is in `days_of_week`.

In case 2 above, `alert_id` is used. A message linked to an alert using custom dates will use the alert ID for UI purposes only.

---
  - https://app.asana.com/0/0/1206860070503129


priv/repo/migrations/20240509181234_change_days_of_week_column_type.exs
```diff
@@ -0,0 +1,15 @@
...
+    execute(&execute_up/0, &execute_down/0)
+  end
+
+  defp execute_up do
+    repo().query!(
+      "ALTER TABLE pa_message ALTER COLUMN days_of_week TYPE integer[] USING (days_of_week::integer[])"
+    )
+  end
+
+  defp execute_down, do: nil
```
sloanelybutsurely, line 14: **nibble (non-blocking):** any reason to not pass this SQL to `execute` directly?

```suggestion
  def change do
    execute "ALTER TABLE pa_message ALTER COLUMN days_of_week TYPE integer[] USING (days_of_week::integer[])", ""
  end
```

priv/repo/migrations/20240509181234_change_days_of_week_column_type.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Screenplay.Repo.Migrations.ChangeDaysOfWeekColumnType do
+  use Ecto.Migration
+
+  def change do
+    execute(&execute_up/0, &execute_down/0)
+  end
+
+  defp execute_up do
+    repo().query!(
+      "ALTER TABLE pa_message ALTER COLUMN days_of_week TYPE integer[] USING (days_of_week::integer[])"
```
sloanelybutsurely, line 10: **question (non-blocking):** do we want a GIN (or GiST) index on this column?

priv/repo/migrations/20240508174055_change_time_columns_name_and_type.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Screenplay.Repo.Migrations.ChangeTimeColumnsNameAndType do
+  use Ecto.Migration
+
+  def change do
+    drop(index("pa_message", [:start_time, :end_time]))
+
+    alter table("pa_message") do
+      remove(:start_time, :utc_datetime)
+      remove(:end_time, :utc_datetime)
```
panentheos, line 9: This doesn't look right to me. Where will we store the start/end dates?

priv/repo/migrations/20240509181234_change_days_of_week_column_type.exs
cmaddox5: I think this was a case of me tunneling on the first example I found for this function 😅 Got this cleaned up.

priv/repo/migrations/20240509181234_change_days_of_week_column_type.exs
cmaddox5: Good idea. Got one added.

lib/screenplay/pa_messages/pa_message.ex
```diff
@@ -0,0 +1,63 @@
...
+            (is_nil(m.start_time_utc) and is_nil(m.end_time_utc) and not is_nil(m.alert_id)) or
+              (^day_of_week in m.days_of_week and
+                 (m.start_time_utc <= ^current_time and m.end_time_utc >= ^current_time))
+      )
+      |> Enum.split_with(&is_nil(&1.alert_id))
+
+    alert_ids = AlertsCache.alert_ids()
+
+    alert_messages =
+      Enum.filter(alert_messages, fn %{alert_id: alert_id} -> alert_id in alert_ids end)
```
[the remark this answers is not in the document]
cmaddox5: Good idea. Still getting warmed up to how Ecto queries work. And I agree that the query string should grow _too_ large because we only list open alert IDs. That is currently at 97 and I've never seen it grow much larger than that.

lib/screenplay/pa_messages/pa_message.ex
```diff
@@ -0,0 +1,63 @@
...
+    day_of_week = Date.day_of_week(now_adjusted)
+
+    current_time = DateTime.to_time(now_adjusted)
+
+    {custom_messages, alert_messages} =
+      Repo.all(
+        from m in __MODULE__,
+          where:
+            (is_nil(m.start_time_utc) and is_nil(m.end_time_utc) and not is_nil(m.alert_id)) or
+              (^day_of_week in m.days_of_week and
```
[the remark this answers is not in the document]
cmaddox5: Yep, definitely. I was working under the assumption that we only linked a message when an alert is active. Got that corrected.

cmaddox5:
@panentheos Got this corrected to account for the most recent designs. The original designs did not have date inputs, so a lot of the original changes probably didn't make a whole lot of sense. Should be good for re-review now.

test/screenplay/pa_messages/pa_message_test.exs
```diff
@@ -0,0 +1,125 @@
...
+
+    {:ok, fetcher} = start_supervised({Screenplay.Alerts.Cache, get_json_fn: get_json_fn})
+    send(fetcher, :fetch)
+
+    insert(:pa_message, %{
+      id: 1,
+      alert_id: "1",
+      start_time: ~U[2024-05-01T04:00:00Z],
+      end_time: ~U[2024-05-02T04:00:00Z],
+      days_of_week: [now_service_day]
```
panentheos, line 112: nit: Since we're hardcoding `now`, we could also hardcode the expected day-of-week in these tests, which would give us verification of `get_current_service_day` for free.

panentheos (approved):
See a couple small comments, then looks good.

priv/repo/migrations/20240509181234_change_days_of_week_column_type.exs
```diff
@@ -0,0 +1,8 @@
+defmodule Screenplay.Repo.Migrations.ChangeDaysOfWeekColumnType do
+  use Ecto.Migration
+
+  def change do
+    execute "ALTER TABLE pa_message ALTER COLUMN days_of_week TYPE integer[] USING (days_of_week::integer[])"
```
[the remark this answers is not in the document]
cmaddox5: I think that's ok. This keeps us from needing to switch back and forth from string to int in queries.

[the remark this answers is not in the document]
PaulJKim: question: what makes this migration irreversible? Is it just because once there are values in this column using whatever type that it's defined to use, we can't change it anymore?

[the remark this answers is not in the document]
panentheos: Certain constructs make a migration irreversible, like `execute/1`, because Ecto has no way to know how to undo the described change. You can avoid it with e.g. `execute/2` by supplying the reverse transformation, though in some cases as you mentioned that might require losing some data. In practice this mostly matters in local dev, where you might end up switching branches a lot.

PR #345, opened 2024-05-16
implement new sidebar design

panentheos:
**Asana task**: [Nav Bar redesign](https://app.asana.com/0/1185117109217413/1206992047975424/f)

This implements the new nav bar designs. This also includes changing the page body background colors to match the new designs. Includes a link to the upcoming PA messages page, but logically commented out for now.


assets/css/screenplay-container.scss
```diff
@@ -10,6 +10,7 @@
     overflow-y: auto;
     overflow-x: hidden;
     color: $text-primary;
+    background-color: #0f1417;
```
cmaddox5, line 13: nit: Can this be put in `variables.scss`? I see it being used as a bg-color elsewhere so would be helpful to have it only defined once in case it changes again.

cmaddox5 (approved):
Wow, the component code is so much cleaner now. Looks great too! Just one suggestion, but otherwise good to go.

PR #346, opened 2024-05-17
fix takeover tool modal styles

panentheos:
Fixes the styling of the confirmation modals in the outfront takeover tool. This mostly replicates the original style, but makes a few simplifications.


sloanelybutsurely (approved):
code lgtm!

PR #348, opened 2024-05-20
tweak: Auth rework

cmaddox5:
**Asana task**: ad-hoc

Paul S. recently created a [Notion doc](https://www.notion.so/mbta-downtown-crossing/Session-Management-ad53a774fd39462c82146f60700db5fc?pvs=4) that outlines best practices for application session management. In the past, we have seen issues with auth expiring way too early and the band-aid fix was to arbitrarily increase TTL to get around this. This doc gives recommendations to managing TTLs so that use recommended TTL values and lean on `ueberatuh_oidcc` to keep sessions alive for only as long as they should be.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -49,12 +49,25 @@ config :phoenix, :json_library, Jason
 # Use Jason for JSON parsing in ExAws
 config :ex_aws, json_codec: Jason
 
-config :screenplay, ScreenplayWeb.AuthManager, issuer: "screenplay"
+# 12 hours in seconds
+max_session_time = 12 * 60 * 60
 
-# Placeholder for Keycloak authentication, defined for real in environment configs
+config :screenplay, ScreenplayWeb.AuthManager,
+  issuer: "screenplay",
+  max_session_time: max_session_time,
+  # 30 minutes
+  idle_time: 30 * 60
+
 config :ueberauth, Ueberauth,
   providers: [
-    keycloak: nil
+    keycloak:
+      {Ueberauth.Strategy.Oidcc,
+       issuer: :keycloak_issuer,
+       userinfo: true,
+       uid_field: "email",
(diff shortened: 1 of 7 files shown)


config/config.exs
```diff
@@ -49,12 +49,25 @@ config :phoenix, :json_library, Jason
...
 
-config :screenplay, ScreenplayWeb.AuthManager, issuer: "screenplay"
+# 12 hours in seconds
+max_session_time = 12 * 60 * 60
+
+config :screenplay, ScreenplayWeb.AuthManager,
+  issuer: "screenplay",
+  max_session_time: max_session_time,
+  # 30 minutes
+  idle_time: 30 * 60
```
sloanelybutsurely, line 59: **question:** Is this also a "max" value? Should we call it `max_idle_time`?

sloanelybutsurely (approved):
LGTM! Just a couple non-blocking suggestions/questions

config/config.exs
cmaddox5: Sure, that seems more appropriate here.

lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -12,6 +12,15 @@ defmodule ScreenplayWeb.AuthManager do
     "pa-message-admin" => :pa_message_admin
   }
 
+  def max_session_time do
+    Application.get_env(:screenplay, __MODULE__)[:max_session_time]
+  end
+
+  def idle_time do
+    Application.get_env(:screenplay, __MODULE__)[:idle_time]
+  end
```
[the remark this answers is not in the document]
cmaddox5: Noting this just in case. A straight copy/paste of these `Application` calls is not quite what is needed. Since the `AuthManager` config is split between compile time and runtime, we needed to use a path in the `compile_env` calls.

PR #349, opened 2024-05-23
PA Messges Page

PaulJKim:
**Asana task**: [PA Messages page](https://app.asana.com/0/1185117109217413/1206992047975431/f)

Description
This PR adds a page for viewing a table of all PA Messages that is only accessible to users that are paess admins. Messages are fetched from the DB and ordered by time added, with the most recently added messages at the top.

Some of the routes were adjusts as well in a way that I felt made more sense. The route to get all active messages is now `/api/pa-messages/active` so that the route to get all messages can be `/api/pa-messages`.

All of the actions (Search, Pause, Save, etc.) are added as placeholders. The filter selection on the left side of the screen is also just a placeholder. Tested locally with local DB data as shown in the screenshot:
Screenshot 2024-05-23 at 1 58 40 PM


test/screenplay/pa_messages/pa_message_test.exs
```diff
@@ -27,6 +27,38 @@ defmodule Screenplay.PaMessages.PaMessageTest do
...
+      insert(:pa_message, %{
+        id: 3,
+        start_time: ~U[2024-05-02T12:00:00Z],
+        end_time: ~U[2024-05-02T12:00:00Z],
+        days_of_week: [3]
+      })
+
+      assert [%PaMessage{id: 3}, %PaMessage{id: 2}, %PaMessage{id: 1}] =
+               PaMessage.get_all_messages()
+    end
```
sloanelybutsurely, line 59: **question:** can we explicitly set `inserted_at` on these test pa messages and remove these calls to `Process.sleep/1` so this test doesn't have to take a fully 2s to run every time?

assets/js/models/pa_message.ts
```diff
@@ -0,0 +1,11 @@
+export interface PaMessage {
+  id: number;
+  sign_ids: string[];
+  priority: number;
+  interval_in_minutes: number;
+  visual_text: string;
+  audio_text: string;
+  start_time: Date;
+  end_time: Date;
+  inserted_at: Date;
```
sloanelybutsurely, line 10: **question:** if this supposed to represent the data returned by the JSON API won't these be `string` values that can be parsed into `Date`s?

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { fetchPaMessages } from "../../utils/api";
+import { PaMessage } from "../../models/pa_message";
+
+const PaMessagesPage: ComponentType = () => {
+  const [paMessages, setPaMessages] = useState<PaMessage[]>([]);
+
+  useEffect(() => {
+    fetchPaMessages().then((allPaMessages) => {
+      setPaMessages(allPaMessages);
+    });
```
sloanelybutsurely, line 13: **note (non-blocking):** I know that React isn't thrilled when you return a Promise (and thus also when you use an `async` function) from a `useEffect` callback. If you'd like to stick with `async`/`await` (I know I like to...) you could create a new `async` function within the callback and call it.

<details>
<summary>Suggested change 📝 </summary>

```suggestion
    const fetchAndSetPaMessages = async () => {
      const allPaMessages = await fetchPaMessages()
      setPaMessages(allPaMessages)
    }
    fetchAndSetPaMessages();
```
</details>

**issue (non-blocking):** Additionally, it looks like this page does not handle loading or errors states for fetching PA messages. If loading PA messages takes a while to load either due to server response time or just a slower connection _or_ the request fails altogether the page will erroneously report **"There are no active PA/ESS Messages."**

Is this something we typically handle? Something like [SWR](https://swr.vercel.app/) could do a lot of the additional state tracking required to support handling something like that.

lib/screenplay/pa_messages/pa_message.ex
```diff
@@ -42,4 +42,8 @@ defmodule Screenplay.PaMessages.PaMessage do
             ((is_nil(m.end_time) and m.alert_id in ^alert_ids) or m.end_time >= ^now)
     )
   end
+
+  def get_all_messages do
+    Repo.all(from(m in __MODULE__, order_by: [desc: m.inserted_at]))
+  end
```
sloanelybutsurely, line 48: **question:** How many rows do we expect this table to contain? Are we concerned about it growing over time? I ask because we aren't limiting this query in any way. Should we add pagination to this resource?

sloanelybutsurely:
@PaulJKim ⚠️ It looks like this PR has a couple commits from https://github.com/mbta/screenplay/pull/345 that have since been squashed and merged into `main`. We should probably rebase over main and drop those commits from this branch.

assets/js/models/pa_message.ts
PaulJKim: ah yes good point, thank you!

assets/js/components/Dashboard/PaMessagesPage.tsx
PaulJKim: 1. Works for me!

2. I don't think we typically handle error/loading states in Screenplay with much special treatment, generally speaking, and normally just default to some empty state for sake of simplicity. That said, it could be something worth discussing especially as the application grows along with its user base. Since it wasn't explicitly called out in this task/designs and I could see it requiring some level of design work if we incorporated that here though, I think it could be a little out of scope for this one.

lib/screenplay/pa_messages/pa_message.ex
PaulJKim: Ah good question. I had this thought in the back of my mind but figured we could circle back to adding pagination later if we felt the need, which I'm sure we eventually would.

I'm not familiar with the specific of how often OIOs create new PAs but my guess would be that it's a little sporadic. Regardless, it seems like this table is meant to show _all_ PA Messages past, present, and future, so I guess that means this table would grow indefinitely. Unless there's a deletion workflow that I missed somewhere, or we decided to expire them from the table after some period of time.

PaulJKim: If it's helpful at all, some rough back of the napkin math has me guesstimating an average of <5 PA Messages created per day.

sloanelybutsurely (approved):
Thanks for making those changes. This LGTM if we're alright handling things like loading states, error states, and pagination when we realize we need them.

PR #361, opened 2024-06-06
feat: New PA Message Page

cmaddox5:
This PR sets up the initial page in the New PA Message creation workflow. Buttons and links on the page that direct the user to a different part of the creation process are just placeholders for now. All state is set up for the components on this page.

Branch is currently on dev-green in case you feel more comfortable testing the changes there.

---
  - https://app.asana.com/0/0/1207371755597655


assets/js/components/Dashboard/PaMessaging/NePaMessagePage/WhereCard.tsx
```diff
@@ -0,0 +1,25 @@
...
+  NewPaMessagePageReducerAction,
+} from "./NewPaMessagePage";
+import { PlusLg } from "react-bootstrap-icons";
+
+interface Props {
+  pageState: NewPaMessagePageState;
+  dispatch: React.Dispatch<NewPaMessagePageReducerAction>;
+}
+
+const WhereCard: ComponentType<Props> = (_props) => {
```
sloanelybutsurely, line 14: **quibble:** I've noticed using `ComponentType` is pretty common but I find it a bit broad and heavy handed. 

`ComponentType` is defined as either a [`ClassComponent` or `FunctionComponent`](
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b59dc3ac1e2770fbd6cdbb90ba52abe04c168196/types/react/index.d.ts#L104) and while `WhereCard` is most close to a `FunctionComponent`, we aren't setting any of the [static properties](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b59dc3ac1e2770fbd6cdbb90ba52abe04c168196/types/react/index.d.ts#L1123-L1184) on these functions.

It is just as correct to use a regular TypeScript function argument annotation:

```suggestion
const WhereCard = (_props: Props) => {
```

assets/js/components/Dashboard/DatePicker.tsx
```diff
@@ -0,0 +1,78 @@
...
+          onChange(inputString);
+        }}
+        onClick={() => setShowOverlay(!showOverlay)}
+      >
+        <Form.Control
+          value={inputString}
+          onChange={(input) => {
+            const inputValue = input.target.value;
+            setInputString(inputValue);
+            onChange(inputValue);
```
sloanelybutsurely, line 41: **question:** is there a reason to maintain this state locally instead of having this be a fully [controlled component](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components)

assets/js/components/Dashboard/DatePicker.tsx
```diff
@@ -0,0 +1,78 @@
...
+      >
+        {({
+          placement: _placement,
+          arrowProps: _arrowProps,
+          show: _show,
+          popper: _popper,
+          hasDoneInitialMeasure: _hasDoneInitialMeasure,
+          ...props
+        }) => (
+          <Popover className="calendar-popover" {...props}>
```
sloanelybutsurely, line 60: **note:** lodash has both [`_.pick`](https://lodash.com/docs/4.17.15#pick) and [`_.omit`](https://lodash.com/docs/4.17.15#omit) that might make this operation more ergonomic.

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/MessageCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { ArrowRightShort, VolumeUpFill } from "react-bootstrap-icons";
+import {
+  NewPaMessagePageState,
+  NewPaMessagePageReducerAction,
+} from "./NewPaMessagePage";
+
+interface Props {
+  pageState: NewPaMessagePageState;
+  dispatch: React.Dispatch<NewPaMessagePageReducerAction>;
+}
```
sloanelybutsurely, line 12: **quibble:** I would personally prefer if these components were explicit about what data from `NewPaMesagePageState` they needed and if they were also passed explicit callbacks that were implemented using the `dispatch` function in the component that renders them

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/NewPaMessagePage.tsx
```diff
@@ -0,0 +1,145 @@
...
+    selectedEndTime: now.add(1, "hour").format("HH:mm"),
+    selectedDayLabel: "All days",
+    selectedDays: [1, 2, 3, 4, 5, 6, 7],
+    selectedPriority: 2,
+    selectedInterval: 4,
+    visualText: "",
+    phoneticText: "",
+  };
+
+  const [state, dispath] = useReducer(reducer, initialState);
```
sloanelybutsurely, line 109: **suggestion:** as there is no logic beyond setting individual keys in the reducer it might actually be simpler to remove the reducer, use a single `useState<NewPaMessagePageState>(initialState)` and update single properties on state in callbacks passed to the children of this component.

Example:

```typescript
onChangeVisualText={(visualText) => 
  setNewPaMessageState(state => ({...state, visualText })}
```

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/NewPaMessagePage.tsx
```diff
@@ -0,0 +1,145 @@
...
+          </div>
+        </Row>
+        <WhenCard pageState={state} dispatch={dispath} />
+        <WhereCard pageState={state} dispatch={dispath} />
+        <MessageCard pageState={state} dispatch={dispath} />
+        <Row
+          md="auto"
+          className="justify-content-end new-pa-message-page__form-buttons"
+        >
+          <Button className="cancel-button" onClick={() => navigate(-1)}>
```
sloanelybutsurely, line 135: **question:** should this link to a specific page? what does this button do if someone opens a link to this page from an external program like Slack or Teams?

**suggestion:** if changing to link to a specific page i'm a big fan of using links/anchor tags to perform navigation

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/NewPaMessagePage.tsx
```diff
@@ -0,0 +1,145 @@
...
+            Cancel
+          </Button>
+          <Button className="submit-button">Submit</Button>
```
sloanelybutsurely, line 138: **question:** Do you want to set a `type="submit"` on this? Or is that part of the submission work?

assets/js/components/Dashboard/TimePicker.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const formatState = () =>
+    moment(
+      `${selectedHour}:${selectedMinute} ${selectedAmPm}`,
+      "hh:mm A",
+    ).format("HH:mm");
+
+  useEffect(() => {
+    onChange(formatState());
+  }, [selectedHour, selectedMinute, selectedAmPm]);
```
sloanelybutsurely, line 32: **suggestion:** this looks like `useMemo` might be handy here

```suggestion
  const formattedState = useMemo(() => 
    moment(
      `${selectedHour}:${selectedMinute} ${selectedAmPm}`,
      "hh:mm A",
    ).format("HH:mm"),
    [selectedHour, selectedMinute, selectedAmPm]
  );

  useEffect(() => {
    onChange(formattedState);
  }, [formattedState]);
```

you could then update the other `onChange(formatState())` call in this file as well.

assets/js/components/Dashboard/DatePicker.tsx
cmaddox5: Hmm, definitely not. Will adjust.

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/NewPaMessagePage.tsx
cmaddox5: I hadn't thought about what this would do if this page was opened in a new tab. I'll adjust this to navigate to `/pa-messages` explicitly in case there is no navigation history to go back to.

cmaddox5: Actually, I'll confirm with Mary what she wants here. I also see value in closing the tab if there is no history so my original idea may not be what she wants.

assets/js/components/Dashboard/PaMessaging/PaMessagesPage.tsx
```diff
@@ -27,7 +29,7 @@ const PaMessagesPage: ComponentType = () => {
           <Col className="pa-message-table-container">
             <Row className="pa-message-table-action-bar">
               <Col>
-                <Button>
+                <Button onClick={() => navigate("/pa-messages/new")}>
```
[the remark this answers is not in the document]
cmaddox5: I think I can wrap this in an `a` tag and make the `navigate` the `href` to make that happen. Will give it a shot.

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/MessageCard.tsx
cmaddox5: Honestly, my reason for going with a reducer is a little silly so I agree that explicit data and callbacks in props is the cleaner way to handle this.

assets/js/components/Dashboard/PaMessaging/NePaMessagePage/NewPaMessagePage.tsx
cmaddox5: Honestly, my reason for going with a reducer is a little silly so I agree that explicit data and callbacks in props is the cleaner way to handle this.

assets/js/components/Dashboard/TimePicker.tsx
panentheos: `useMemo` is mainly used to avoid re-running expensive computations multiple times. Since this formatting is trivial, it should be fine to just compute it inline without `useMemo`

assets/js/components/Dashboard/PaMessaging/PaMessagesPage.tsx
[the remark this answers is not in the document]
sloanelybutsurely: @cmaddox5 why not use [`<Link />`](https://reactrouter.com/en/main/components/link#link)?

cmaddox5:
@panentheos @sloanelybutsurely Tried my hand at another way of organizing the components. My solution at making the page more readable resulted in the state management being too hard to follow. The latest iteration seems like the best of both worlds when it comes to individual state variables and minimal number of props in each component.

assets/js/components/Dashboard/DatePicker.tsx
```diff
@@ -0,0 +1,68 @@
...
+const DatePicker = ({
+  selectedDate,
+  onChange,
+  minDateString,
+  maxDateString,
+}: DatePickerProps) => {
+  const [showOverlay, setShowOverlay] = useState(false);
+  const ref = useRef(null);
+  const minDate = minDateString ? new Date(minDateString) : undefined;
+  const maxDate = maxDateString ? new Date(maxDateString) : undefined;
```
sloanelybutsurely, line 22: **quibble (non-blocking):** i like to avoid `undefined` as a programmer-set value and would suggest using `null` unless something _absolutely needs_ an `undefined` value

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/DaysPicker.tsx
```diff
@@ -0,0 +1,107 @@
...
+                  key={value}
+                  inline
+                  checked={days.includes(value)}
+                  label={label}
+                  type="checkbox"
+                  onChange={(checkbox) => {
+                    if (checkbox.target.checked) {
+                      onChangeDays(fp.concat(days, [value]));
+                    } else {
+                      onChangeDays(fp.pull(value, days));
```
sloanelybutsurely, line 94: **quibble (non-blocking):** i have a slight personal preference for [`_.without`](https://lodash.com/docs/4.17.15#without) because i find it reads more clearly

assets/js/components/Dashboard/TimePicker.tsx
```diff
@@ -0,0 +1,155 @@
...
+      moment(
+        `${selectedHour}:${selectedMinute} ${selectedAmPm}`,
+        "hh:mm A",
+      ).format("HH:mm"),
+    [selectedHour, selectedMinute, selectedAmPm],
+  );
+
+  useEffect(() => {
+    onChange(formattedState);
+  }, [selectedHour, selectedMinute, selectedAmPm]);
```
sloanelybutsurely, line 32: **suggestion:** looks like you missed updating the dependencies of this effect

```suggestion
  }, [formattedState]);
```

sloanelybutsurely (approved):

assets/js/components/Dashboard/DatePicker.tsx
cmaddox5: I normally agree, but this was due to the accepted props in the `Calendar` component. Only `Date` objects are accepted and unfortunately `new Date(undefined)` returns a valid object.

assets/js/components/Dashboard/PaMessaging/PaMessagesPage.tsx
```diff
@@ -27,10 +28,12 @@ const PaMessagesPage: ComponentType = () => {
           <Col className="pa-message-table-container">
             <Row className="pa-message-table-action-bar">
               <Col>
-                <Button>
-                  <PlusCircleFill className="pa-message-table-action-bar__plus" />{" "}
-                  Add New
-                </Button>
+                <Link to="/pa-messages/new">
+                  <Button>
```
panentheos, line 32: Nesting links and buttons isn't proper markup. If we want a link that looks like a button, it should be possible to apply the class names to the link to give it that appearance.

panentheos (commented):
I see that this got merged while I was writing up comments, which is okay since this page is still being worked on. I'll put them here anyway, and we can use them as a reference for follow-up work.

assets/js/components/Dashboard/TimePicker.tsx
```diff
@@ -0,0 +1,155 @@
...
+  selectedTime: string;
+  onChange: (time: string) => void;
+}
+
+const TimePicker = ({ selectedTime, onChange }: TimePickerProps) => {
+  const timeMoment = moment(selectedTime, "h:mm A");
+  const [showOverlay, setShowOverlay] = useState(false);
+  const [selectedHour, setSelectedHour] = useState(timeMoment.format("hh"));
+  const [selectedMinute, setSelectedMinute] = useState(timeMoment.format("mm"));
+  const [selectedAmPm, setSelectedAmPm] = useState(timeMoment.format("A"));
```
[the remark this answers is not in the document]
cmaddox5: Good point. This is a little more convenient but not at all necessary.

PR #366, opened 2024-06-11
feat: include client_ip in logger metadata

sloanelybutsurely:
**Asana task** (related): [Research impacts of new load balancer restrictions](https://app.asana.com/0/1185117109217413/1207372071035389/f)

### Description

Needed to evaluate the number of requests per IP over time

- Adds `LoggerMetadata` Plug to extract `client_ip` and set it on `Logger` metadata
- Includes `:client_ip` in the Logger metadata config to output


digitalcora:
Ah, it's a slightly late comment now, but we can use [`remote_ip`](https://hexdocs.pm/remote_ip/RemoteIp.html) to avoid implementing this ourselves (theirs is probably more robust). It "upgrades" the existing `remote_ip` field of the conn to account for things like x-forwarded-for and also adds Logger metadata.

sloanelybutsurely:
@digitalcora happy to make that change! i was just porting over [what we have in `screens` already](https://github.com/mbta/screens/blob/fdc46a33f193ea233f457676312bafe1b2c25840/lib/screens_web/plugs/metadata.ex)

digitalcora:
More refactoring opportunities ahoy! 😅

PR #372, opened 2024-06-14
feat: Preview Audio

cmaddox5:
This PR adds the ability to preview audio for a PA message. Text that is entered is sent to Watts, which returns an audio file that is played in the browser. An error toast is displayed if the Watts call fails. While this sets up the foundation for what is need in form validation, this doesn't tackle the actual validation on form submit. I felt it was appropriate to save that for the Form Submission task. 

---
  - https://app.asana.com/0/0/1207371755597657

### DIFF
--- a/.envrc.template
+++ b/.envrc.template
@@ -26,3 +26,8 @@ export DATABASE_HOSTNAME="localhost"
 
 # Access key for allowing PA Message API calls. Required for endpoints in the /api/pa-messages scope. Generate a unique random string for local use.
 export SCREENPLAY_API_KEY="local_api_key"
+
+# Host and access key needed to communicate with Watts. Required when previewing PA message audio.
+# For local use, set to the same value as your local instance of Watts.
+#export WATTS_URL=""
+#export WATTS_API_KEY=""
--- a/.envrc.template
+++ b/.envrc.template
@@ -29,5 +29,6 @@ export SCREENPLAY_API_KEY="local_api_key"
 
 # Host and access key needed to communicate with Watts. Required when previewing PA message audio.
 # For local use, set to the same value as your local instance of Watts.
-#export WATTS_URL=""
+#export WATTS_URL="https://watts-staging.mbtace.com"
+# API Key is in the Screens 1Password vault under `Watts Staging API Key`
 #export WATTS_API_KEY=""
--- a/.envrc.template
+++ b/.envrc.template
@@ -29,6 +29,7 @@ export SCREENPLAY_API_KEY="local_api_key"
 
 # Host and access key needed to communicate with Watts. Required when previewing PA message audio.
 # For local use, set to the same value as your local instance of Watts.
+# VPN is required for requests to watts-staging.
 #export WATTS_URL="https://watts-staging.mbtace.com"
 # API Key is in the Screens 1Password vault under `Watts Staging API Key`
 #export WATTS_API_KEY=""
--- a/assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
+++ b/assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import {
   Alert,
   Button,
@@ -44,6 +44,10 @@ const NewPaMessagePage = () => {
   const [errorMessage, setErrorMessage] = useState("");
 
   const navigate = useNavigate();
+
+  useEffect(() => {
+    if (!audioPlaying) setAudioReviewed(false);
+  }, [visualText, phoneticText]);
 
   const previewAudio = () => {
     if (audioPlaying) return;
--- a/assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
+++ b/assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
@@ -57,6 +57,7 @@ const NewPaMessagePage = () => {
     setAudioPlaying(true);
   };
 
+  // Called after the audio preview plays in full
   const onAudioEnded = () => {
     setAudioPlaying(false);
     setAudioReviewed(true);
@@ -164,7 +165,10 @@ const NewPaMessagePage = () => {
                 <MessageTextBox
                   id="visual-text-box"
                   text={visualText}
-                  onChangeText={setVisualText}
+                  onChangeText={(text) => {
+                    setVisualText(text);
+                    setAudioPlaying(false);
+                  }}
                   label="Text"
                 />
               </Col>
@@ -172,7 +176,10 @@ const NewPaMessagePage = () => {
                 <Button
                   disabled={visualText.length === 0}
                   className="copy-text-button"
-                  onClick={() => setPhoneticText(visualText)}
+                  onClick={() => {
+                    setPhoneticText(visualText);
+                    setAudioPlaying(false);
+                  }}
                   aria-label="copy-visual-to-phonetic"
                 >
                   <ArrowRightShort />
@@ -184,7 +191,10 @@ const NewPaMessagePage = () => {
                     <MessageTextBox
                       id="phonetic-audio-text-box"
                       text={phoneticText}
-                      onChangeText={setPhoneticText}
+                      onChangeText={(text) => {
+                        setPhoneticText(text);
+                        setAudioPlaying(false);
+                      }}
                       disabled={phoneticText.length === 0}
                       label="Phonetic Audio"
                     />
@@ -193,9 +203,6 @@ const NewPaMessagePage = () => {
                       audioReviewed={audioReviewed}
                       showErrorIcon={errorMessage.length > 0}
                       onClick={previewAudio}
-                      audioText={phoneticText}
-                      onAudioEnded={onAudioEnded}
-                      onAudioError={onAudioError}
                     />
                   </>
                 ) : (
@@ -207,12 +214,17 @@ const NewPaMessagePage = () => {
                         audioReviewed={audioReviewed}
                         disabled={visualText.length === 0}
                         onClick={previewAudio}
-                        audioText={phoneticText}
-                        onAudioEnded={onAudioEnded}
-                        onAudioError={onAudioError}
                       />
                     </Card>
                   </>
+                )}
+                {audioPlaying && (
+                  <audio
+                    src={`/api/pa-messages/preview_audio?text=${phoneticText}`}
+                    autoPlay
+                    onEnded={onAudioEnded}
+                    onError={onAudioError}
+                  />
                 )}
               </Col>
             </Row>
@@ -257,9 +269,6 @@ interface ReviewAudioButtonProps {
   disabled?: boolean;
   showErrorIcon?: boolean;
   onClick: () => void;
-  audioText: string;
-  onAudioEnded: () => void;
-  onAudioError: () => void;
 }
 
 const ReviewAudioButton = ({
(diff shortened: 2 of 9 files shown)


assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -22,9 +23,27 @@ const NewPaMessagePage = () => {
...
 
   const navigate = useNavigate();
 
+  const previewAudio = () => {
+    setAudioPlaying(true);
+    fetchAudioPreview(phoneticText.length ? phoneticText : visualText).then(
+      async (audioData) => {
+        const blob = await audioData.blob();
+        const url = URL.createObjectURL(blob);
+        const audio = new Audio(url);
```
digitalcora, line 72: It feels like we should be able to streamline this by letting the browser itself fetch the audio (constructing the `Audio` by passing in a URL like `/preview_audio?text=Here%20is%20my%20message`)... though I don't think there's a way to have this use a non-GET request, which means the length of the text we could preview would be limited by the length of a URL. Still, a URL can be around 2000 characters (8000 if you ignore Edge), which feels like it should be more than enough for a PA message. What do you think of this possibility?

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -31,6 +40,8 @@ const NewPaMessagePage = () => {
   const [phoneticText, setPhoneticText] = useState("");
   const [audioPlaying, setAudioPlaying] = useState(false);
   const [audioReviewed, setAudioReviewed] = useState(false);
+  const [showErrorAlert, setShowErrorAlert] = useState(false);
+  const [errorMessage, setErrorMessage] = useState("");
```
digitalcora, line 44: Do these need to be separate pieces of state? It seems like either we are showing an error _and_ there is an error message, or we are not showing an error (and there is no error message).

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -43,6 +54,15 @@ const NewPaMessagePage = () => {
     setAudioPlaying(true);
     fetchAudioPreview(phoneticText.length ? phoneticText : visualText).then(
       async (audioData) => {
+        if (audioData.status !== 200) {
```
digitalcora, line 61: `response` might be a better name for this variable, as it's not just the raw data.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -224,13 +257,15 @@ interface ReviewAudioButtonProps {
...
   disabled?: boolean;
+  showErrorIcon?: boolean;
   onClick: () => void;
 }
 
 const ReviewAudioButton = ({
   audioPlaying,
   audioReviewed,
   disabled,
+  showErrorIcon,
```
digitalcora, line 272: Since I think we're only ever in one of the "playing", "reviewed", or "errored" states at a time, maybe those three props could be replaced with a single one, defined as an enum? Or does it make sense to possibly have multiple of these states at once?

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -45,6 +45,10 @@ const NewPaMessagePage = () => {
 
   const navigate = useNavigate();
 
+  useEffect(() => {
+    if (!audioPlaying) setAudioReviewed(false);
+  }, [visualText, phoneticText]);
```
digitalcora, line 50: Suppose I start playing the audio preview, then entirely within the duration of the preview, edit the text (maybe fixing a typo). In this case the audio would still count as "reviewed", despite the fact that I haven't actually previewed the audio for the new version of the text. This seems like a gap in the logic — maybe we are not really modeling the states correctly (in other words there should be an "audio has been reviewed" flag which is separate from the display state of the review button?).

.envrc.template
```diff
@@ -26,3 +26,8 @@ export DATABASE_HOSTNAME="localhost"
 
 # Access key for allowing PA Message API calls. Required for endpoints in the /api/pa-messages scope. Generate a unique random string for local use.
 export SCREENPLAY_API_KEY="local_api_key"
+
+# Host and access key needed to communicate with Watts. Required when previewing PA message audio.
+# For local use, set to the same value as your local instance of Watts.
+#export WATTS_URL=""
```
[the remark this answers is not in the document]
cmaddox5: The issue with that is not having a way to retrieve the API key. It might be worth requesting the staging key be added to 1Password and then a comment in this file can include details on where to find it. Until then, a default that isn't the local instance would not help with using the app.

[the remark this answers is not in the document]
panentheos: The staging key might be in the shared 1P already. Keep in mind that watts staging returns an empty response to all calls, to avoid polly costs. (It might be good to have it return a hard-coded but real mp3 file, so we could actually hear something (without linking with prod)

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
cmaddox5: That's a good point. The boolean can just be `.length > 0`.

.envrc.template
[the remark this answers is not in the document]
cmaddox5: It doesn't seem to be in any vault I have access to. Do you know if there is a separate vault for PA/ESS related stuff? If not, I can file an infra ticket to have the staging key added and will make the template change here. 

Good idea on the staging return change. I can tackle that as part of this task.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
cmaddox5: I like this idea. The length requirement doesn't bother me because messages that exceed that length are probably too long to play anyway. I'd hate to add that without the user having knowledge of an actual limit, though. Maybe I should check with Mary/Jim and see if we should just make the decision to add a limit to the textboxes?

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -9,6 +9,7 @@ import moment from "moment";
 import DatePicker from "../../DatePicker";
 import TimePicker from "../../TimePicker";
 import { ArrowRightShort, PlusLg, VolumeUpFill } from "react-bootstrap-icons";
+import { fetchAudioPreview } from "../../../../utils/api";
```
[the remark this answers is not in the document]
cmaddox5: Totally on board with that. Can that be a followup PR to keep scope of this review limited, or would you be ok with that being included here?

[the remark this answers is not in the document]
panentheos: A flatter directory structure can help a lot with this, too. Right now the "full name" of this component is `Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage`, whereas e.g. `Dashboard/NewPaMessagePage` or even `NewPaMessagePage` would be just as unambiguous.

Less nesting also means less arbitrary taxonomy that can drift over time. As an example, if something like `IntervalPicker` ended up being a useful component and we use it somewhere else, we'd be faced with either moving the whole file out of that directory, or else having it be "wrong".

This is all just food for thought, not requesting changes here.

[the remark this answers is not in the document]
digitalcora: Good thought, I hadn't even noticed the duplicativeness in the path of this file, which I agree could be cut down a bit. For path aliases, since you could argue we should go back and use them to replace existing relative paths, I'm fine with that being a separately-reviewed thing.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
digitalcora: If we had a real-time character count a la Twitter it would jump around a lot, since non-URL-safe characters (most notably spaces) would take up more than one character. We could at least prevent entering text that would take the input over the limit, and given how unlikely this is, maybe that's enough. I'd ask if having the limit is okay, of course, but I don't see why it wouldn't be.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
cmaddox5: Mary made the decision to stop audio from playing when either textbox changes. This should eliminate the edge case you mentioned where audio is marked as reviewed if the text changes while playing. Now, the `onEnded` callback will only fire if the whole audio file is played without interruption.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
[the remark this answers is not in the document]
cmaddox5: Task to track this: https://app.asana.com/0/1185117109217413/1207589542683791/f

.envrc.template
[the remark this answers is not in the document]
cmaddox5: [Here](https://app.asana.com/0/1113179098808463/1207601226251965/f) is the infra task for adding/giving our team access to the API key in 1Password. [Here](https://app.asana.com/0/1185117109217413/1207589542683798/f) is the task for changing the response of non-prod environments.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
cmaddox5: Took a stab at this. Hopefully I didn't make it too complicated.

assets/js/components/Dashboard/PaMessaging/NewPaMessagePage/NewPaMessagePage.tsx
```diff
@@ -8,7 +17,23 @@ import { useNavigate } from "react-router-dom";
...
+} from "react-bootstrap-icons";
+import cx from "classnames";
+
+const MAX_TEXT_LENGTH = 2000;
+
+enum AudioState {
+  HasNotBeenPlayed,
+  IsPlaying,
+  HasBeenReviewed,
+  NeedsReview,
```
digitalcora, line 35: What is the difference between `HasNotBeenPlayed` and `NeedsReview`? Playing the audio _is_ reviewing it, right? I'm a bit confused by the naming here since half the states refer to "playing" and half to "reviewing".

cmaddox5: `HasNotBeenPlayed` is used as the default state. Without some kind of default, the icon next to `Review audio` will change prematurely to a warning icon when the text in the boxes is updated before audio is played. Another idea I had was just using `null` as the default considering it is only used on page load. Didn't see it before, but I agree that the two names overlap a bit too much.

digitalcora: I see. What do you think of this naming, then? (same order, just took a stab at different names)

```ts
enum AudioState {
  Unreviewed,
  Playing,
  Reviewed,
  Outdated,
}
```

digitalcora (approved):
Looking good, let's deploy to dev-green to make sure we also have the infrastructure pieces in place and working (I didn't see this happen yet in the PR feed).

cmaddox5:
Was finally able to work through all of the dev-green problems. TL;DR, Screenplay now requests from Watts via the URL to the internal LB. Was also able to get the `watts-staging` API key into 1Password so `.envrc.template` has been updated.

PR #389, opened 2024-07-02
fix: Publish error fixes

cmaddox5:
**Asana task**: [[extra] Screenplay prod crash in PermConfig when trying to publish changes](https://app.asana.com/0/1185117109217413/1207697513391315/f)

Some context from Slack: https://mbta.slack.com/archives/G012H4G4U06/p1719868261122079?thread_ts=1719859672.363579&cid=G012H4G4U06.

Tl;dr, Ana had trouble publishing screens due to an error when trying to upload a new `places_and_screens` file to S3. Two big problems here:
1. The data we send to S3 wasn't being encoded. That's the reason for the Sentry error.
2. We shouldn't bother with talking to S3 if we know the config isn't changing.

We know the config isn't changing here because Ana checked `Hide on Places page` for each of the pending screens. This results in no changes to `places_and_screens`.

### DIFF
--- a/lib/screenplay/config/permanent_config.ex
+++ b/lib/screenplay/config/permanent_config.ex
@@ -4,7 +4,7 @@ defmodule Screenplay.Config.PermanentConfig do
   # Suppress dialyzer warning until more app_ids are implemented.
   @dialyzer [{:nowarn_function, get_route_id: 3}, {:nowarn_function, json_to_struct: 4}]
 
-  alias Screenplay.Config.PlaceAndScreens
+  alias Screenplay.Config.{Fetch, PlaceAndScreens}
   alias Screenplay.PendingScreensConfig.Fetch, as: PendingScreensFetch
   alias Screenplay.RoutePatterns.RoutePattern
   alias Screenplay.ScreensConfig.Cache, as: ScreensConfigCache
@@ -86,7 +86,7 @@ defmodule Screenplay.Config.PermanentConfig do
     }
   end
 
-  @spec put_pending_screens(map(), screen_type(), binary()) ::
+  @spec put_pending_screens(map(), screen_type(), Fetch.version_id()) ::
           {:error,
            :version_mismatch
            | :config_not_fetched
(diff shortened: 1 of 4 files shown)


lib/screenplay/config/s3_fetch.ex
```diff
@@ -37,9 +37,15 @@ defmodule Screenplay.Config.S3Fetch do
 
   @impl true
```
digitalcora, line 38: We should update the spec for this function in the `Screenplay.Config.Fetch` behaviour — it was specced to accept a `String.t()`, which was sort-of correct for the old version of the S3 function but not for the local function. I'm not sure why Dialyzer accepted this (a phrase I find myself using a lot), but it might be down to a lack of explicit types elsewhere in this code path.

It looks like this accepts an argument of the type `get_places_and_screens` returns in its `:ok` case, which is... currently specced as `term` 🙃 I know ultimately we could get back anything from decoding JSON, but is there an easy way we could clarify this type? Even if it's just "list of maps" or something.

lib/screenplay/config/permanent_config.ex
```diff
@@ -148,12 +148,17 @@ defmodule Screenplay.Config.PermanentConfig do
 
     new_published_screens_config = get_new_published_screens(published_config, screens_to_publish)
 
+    screenplay_screens_to_add =
+      screens_to_publish
+      |> Enum.into(%{})
```
digitalcora, line 153: Is this needed? It seems like we're going from a list, into a map, and then immediately back into a list.

lib/screenplay/config/s3_fetch.ex
cmaddox5: Huh, thanks for calling that out. I leaned too heavily on Dialyzer 😬 

As for the `term` spec, that's due to the face that `Jason.decode/1` is specced  to return `{:ok, term()}` so Dialyzer sees that and rejects `map()` in the spec. Have you solved something like that before?

cmaddox5: Nevermind... I think this was a problem with a spec higher up.

lib/screenplay/config/permanent_config.ex
cmaddox5: It is not, thank you.

digitalcora (approved):
Pre-emptively approving, pending a check on dev-green!

cmaddox5:
Was able to publish a screen with no error on dev-green. Will merge after CI finishes.

PR #397, opened 2024-07-11
Add associate alerts workflow

PaulJKim:
**Asana task**: [Associate alert workflow](https://app.asana.com/0/1185117109217413/1207371755597661/f)

Description
Add the workflow for associating a new PA Message with an alert and potentially importing some of it's attributes


assets/js/components/Dashboard/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,327 @@
...
+  Button,
+  Container,
+  Row,
+  Col,
+  ButtonGroup,
+  Modal,
+  Form,
+} from "react-bootstrap";
+import { useNavigate } from "react-router-dom";
+import { useScreenplayContext } from "../../hooks/useScreenplayContext";
```
cmaddox5, line 12: We have some aliases now that makes these paths a bit easier to read. For example, you can say `Hooks/useScreenplayContext` here. You can find all aliases [here](https://github.com/mbta/screenplay/blob/59a41717febf6b907531ec18947b07791e66b3ea/assets/webpack.config.js#L60).

assets/js/components/Dashboard/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+const AssociateAlertPage = () => {
+  const navigate = useNavigate();
+  const {
+    associatedAlert,
+    endWithEffectPeriod: endWithEffectPeriodState,
+    importLocations: importLocationsState,
+    importMessage: importMessageState,
+  } = useAssociatedAlertContext();
+  const { alerts } = useScreenplayContext();
```
cmaddox5, line 29: I'm realizing now that we might need to fetch these alerts separately on page load. The alerts in context filter to only alerts that can show on showtime screens. This rules out a large number of effects and also all of Mattapan and Silver Line. It also includes Access alerts which we don't want here.

assets/css/dashboard/new-pa-message-page/associate-alert-page.scss
```diff
@@ -0,0 +1,201 @@
...
+    height: 38px;
+    min-width: 200px;
+    color: $text-secondary;
+    background-color: $cool-gray-20;
+    border: none;
+  }
+}
+
+.associate-alert-table {
+  margin: 40px 48px;
```
cmaddox5, line 55: For some reason, it seems like the page is ignoring the right margin here. Table is overflowing out of the `col` and out of the viewport. If you remove the `margin-right` here and make it `padding-right` in `.associate-alert-page-body` it might be a bit more what you are looking for.

Minor note: As I build out some of the PA Message components, I'm coming to the conclusion that the Bootstrap magic in the `Row` and `Col` components are just frustrating to work with. On the Select Stations page, I started using `div`s in a flexbox and find that it's much easier to get margins and paddings to work how I expect them to. Thinking a bit into the future, I wonder if we should do our own layouts and just use Bootstrap for the visual components. Nothing to do here, just thinking out loud.

assets/js/components/Dashboard/NewPaMessagePage.tsx
```diff
@@ -129,15 +195,40 @@ const NewPaMessagePage = () => {
...
                   End
                 </Form.Label>
-                <div className="datetime-picker-group">
-                  <DatePicker
-                    selectedDate={endDate}
-                    onChange={setEndDate}
-                    minDateString={startDate}
-                    id="end-date-picker"
+                {associatedAlert.id && (
+                  <Form.Check
```
cmaddox5, line 199: You can actually use `Form.Switch` here and remove the `type`. Results in the same component being rendered.

assets/js/components/Dashboard/NewPaMessagePage.tsx
```diff
@@ -38,15 +44,26 @@ enum AudioPreview {
 const NewPaMessagePage = () => {
   const now = moment();
 
+  const associatedAlertState = useAssociatedAlertContext();
```
cmaddox5, line 47: My hunch is this shouldn't be in the global context. This is the only page that will need this info and there aren't many layers in this workflow that require it. Should this instead be three additional pieces of local state? I know it requires some additional props on the `AssociateAlertPage` component, but that seems more appropriate than putting it in global state. What do you think?

assets/js/components/Dashboard/NewPaMessagePage.tsx
```diff
@@ -86,17 +113,56 @@ const NewPaMessagePage = () => {
...
+            )}
           </Row>
+          {!associatedAlert.id && (
+            <Row>
+              <div className="new-pa-message-page__associate-alert-subtext">
+                Linking will allow you to share end time with alert, and import
+                location and message.
+              </div>
+            </Row>
+          )}
```
cmaddox5, line 165: Can this be pulled out into its own component? Some of this is getting a bit hard to follow with all the different conditionals.

assets/js/components/App.tsx
```diff
@@ -51,6 +53,10 @@ class AppRoutes extends React.Component {
             </Route>
             <Route path="pa-messages" element={<PaMessagesPage />} />
             <Route path="pa-messages/new" element={<NewPaMessagePage />} />
+            <Route
```
cmaddox5, line 56: I don't think the user should be able to get to this page from anywhere but the link on `pa-messages/new`. Sloane and I thought about the same thing in the Select Stations task and came to [this solution](https://github.com/mbta/screenplay/blob/select-stations-page/assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx). Basically, each of these separate pages are an Enum value and this container component conditionally renders based on what value is saved in state. This allows the pages to feel separate without needing to handle the user navigating directly to the page with this specific URL.

assets/js/components/Dashboard/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,327 @@
...
+  const [selectedMessageState, setSelectedMessageState] =
+    useState<string>("active");
+  const [showAlertModal, setShowAlertModal] = useState<boolean>(
+    Object.keys(associatedAlert).length !== 0,
```
cmaddox5, line 33: Could this be `associateAlert.id`?

assets/js/components/Dashboard/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,327 @@
...
+              </div>
+              <Button
+                className="alert-description__header-change-alert"
+                onClick={() => {
+                  setSelectedAlert({} as Alert);
+                  setShowAlertModal(false);
+                }}
+                variant="link"
+              >
+                Change alert
```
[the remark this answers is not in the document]
PaulJKim: Yeah that's a good point considering they do the exact same thing. I'll ask Mary about it

assets/css/dashboard/new-pa-message-page/associate-alert-page.scss
PaulJKim: Ah yep that does look better to me, thank you!

And yeah I was kind of running into those struggles at first as well. I think eventually I just sort of got used to it and worked it out but I'm onboard for avoiding using `Row` and `Col` moving forward

assets/js/components/Dashboard/NewPaMessagePage.tsx
PaulJKim: I agree it does seem more appropriate, but the issue I ran into was that you can't pass the `setState` callbacks via `navigate` because they're not serializable. Using context is just pattern that I was familiar with which let me achieve what I needed to do, but I'd certainly rather go with a pattern that make more sense rather than bloating the global state. Any suggestions on alternative approaches?

PaulJKim: Actually I guess I could just not use React Router here and just conditionally render either the associate alerts page or the new PA message page based off a new flag

PaulJKim: Ah, but that means the sidebar still shows. Not sure if that's technically an issue but it isn't explicitly part of the design

assets/js/components/Dashboard/AssociateAlertPage.tsx
PaulJKim: I see. There's a not about deferring any alert filtering to a later task, but that's just for Access alerts. We certainly shouldn't rule out Mattapan and Silver Line in this task

assets/js/components/Dashboard/NewPaMessagePage.tsx
cmaddox5: The visible Sidebar was called out by Mary in my branch so it might also be a problem here. I was able to handle it like [this](https://github.com/mbta/screenplay/blob/7cd5ecc41c4cecb6e41c4e6542d9af65b4c4de46/assets/css/dashboard/new-pa-message/select-stations-page.scss#L3). Not super elegant but might be the best way considering it doesn't have a route to filter by.

assets/js/hooks/useScreenplayContext.tsx
```diff
@@ -77,6 +77,14 @@ type ConfigValidationReducerAction = {
   pendingScreenValidationErrors: ConfigValidationErrors;
 };
 
+type AssociatedAlertReducerAction = {
```
cmaddox5, line 80: Can the changes in this file be removed?

assets/js/components/Dashboard/NewPaMessage/NewPaMessagePage.tsx
```diff
@@ -314,4 +389,92 @@ const ReviewAudioButton = ({
...
+      <div className="affect-period">
+        Alert Effect period: {start} - {end}
+      </div>
+    );
+  };
+
+  return (
+    <>
+      <Row md="auto" className="align-items-center">
+        {associatedAlert.id ? (
```
cmaddox5, line 432: suggestion: It might be a bit easier to follow if you wrap the whole component in this conditional. So basically:

```ts
return {associatedAlert.id ? AssociatedAlertStuff : NotAssociatedAlertStuff}
```

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,339 @@
...
+            className="cancel-button"
+            variant="link"
+          >
+            Cancel
+          </Button>
+          <Button
+            onClick={() => {
+              if (importMessage) {
+                setVisualText(associatedAlert.header);
+                setPhoneticText(associatedAlert.header);
```
cmaddox5, line 212: Because users will still need to review audio and reviewing audio sets the phonetic text, you can leave this out and just set visual.

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+  const last_modified = new Date(alert.updated_at)
+    .toLocaleTimeString("en-US", {
+      day: "numeric",
+      month: "numeric",
+      year: "numeric",
+      hour: "numeric",
+      minute: "numeric",
+    })
+    .replace(",", "");
```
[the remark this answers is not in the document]
PaulJKim: Ah thanks!

assets/css/dashboard/new-pa-message-page/new-pa-message-page.scss
```diff
@@ -66,6 +66,10 @@
     .end-datetime {
       width: 100%;
 
+      .effect-period-switch {
```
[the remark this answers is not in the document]
PaulJKim: Yeah I noticed this as well and need to dig into this one

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,347 @@
...
+            <th className="associate-alert-table__last-modified">
+              Last modified
+            </th>
+            <th className="associate-alert-table__select"></th>
+          </tr>
+        </thead>
+        <tbody>
+          {filteredAlerts.length == 0 ? (
+            <tr>
+              <div className="associate-alert-table__empty">
```
cmaddox5, line 279: Noticed a console error here. I think direct children of `tr` tags need to be either `th` or `td` elements. Should be able to just wrap the `div` in a `td`.

assets/css/dashboard/new-pa-message-page/associate-alert-page.scss
```diff
@@ -0,0 +1,220 @@
...
+    font-weight: 400;
+  }
+}
+
+.alert-items-modal {
+  color: white;
+
+  .modal-content {
+    background-color: $cool-gray-20;
+    min-width: 800px;
```
cmaddox5, line 157: I noticed that the modal looks a bit off centered horizontally and I think it's because the parent container is not changing its width based on the children. It looks like React Bootstrap adds a `.modal-dialog` element between `<Modal>` and its first child by default. If you move the `min-width` in `.modal-content` to `.modal-dialog` it should handle centering a bit better.

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const NewPaMessage = () => {
+  const [page, setPage] = useState<Page>(Page.NEW);
+
+  const now = moment();
+
+  const [associatedAlert, setAssociatedAlert] = useState<Alert>({} as Alert);
+  const [importLocations, setImportLocations] = useState<boolean>(true);
+  const [importMessage, setImportMessage] = useState<boolean>(true);
+  const [endWithEffectPeriod, setEndWithEffectPeriod] = useState<boolean>(true);
```
cmaddox5, line 17: Default should be `false`, right? I imagine this was hardcoded for testing so just want to make sure.

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,347 @@
...
+  alert,
+  setAssociatedAlert,
+  setShowAlertModal,
+}: AssociateAlertsTableRowProps) => {
+  const [start, end] = getAlertEarliestStartLatestEnd(alert.active_period);
+
+  const last_modified = moment(alert.updated_at).format("l LT");
+
+  return (
+    <tr
```
cmaddox5, line 317: Since you can click the row itself to select the alert, should it have a pointer cursor? I think it would help the user recognize that the row is clickable.

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,347 @@
...
+      <td>{alert.header}</td>
+      <td>{alert.id}</td>
+      <td>
+        {start}
+        <br />
+        {end}
+      </td>
+      <td>{last_modified}</td>
+      <td className="associate-alert-table__select">
+        <Button
```
cmaddox5, line 333: The focus border on this button doesn't match the height of the container like I think it should. I think adding `height: fit-content` to this button element should fix that.

cmaddox5 (commented):
Sorry for taking so long on this pass.

assets/js/components/Dashboard/NewPaMessage/NewPaMessagePage.tsx
```diff
@@ -314,4 +389,88 @@ const ReviewAudioButton = ({
...
+  return associatedAlert.id ? (
+    <Row md="auto" className="align-items-center">
+      <div className="associated-alert-header">
+        Associated Alert: Alert ID {associatedAlert.id}
+        <Button variant="link" onClick={() => navigateTo(Page.ALERTS)}>
+          Edit
+        </Button>
+        <Button
+          variant="link"
+          onClick={() => {
```
panentheos, line 438: With sub-components like this, I find it's helpful to think in terms of passing callbacks (as opposed to setters), which sometimes leads to some natural abstraction. In this case, if we passed a single `onClear` callback, we could perform all these state updates in the parent's lexical scope, which would eliminate a lot of boilerplate (passing, destructuring, type signatures)

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
PaulJKim: Mary actually called out during QA that the checkboxes should default to selected

cmaddox5: That makes sense, but causes a problem when the page loads. Because it's true by default and there is no associated alert, the end date and time inputs are not visible.

assets/css/dashboard/new-pa-message-page/associate-alert-page.scss
PaulJKim: ah, nice thank you!

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
PaulJKim: Yeah I think it doesn't hurt. I'll add that in

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
PaulJKim: I added some more conditional rendering to let us keep the default as true but still render the date input. Let me know what you think of it

cmaddox5:
@PaulJKim I see the recent changes to remove `Edit` but wanted to bump my most recent comments before those changes. There should be some state cleanup that can be done now that we can always assume the user is navigating to the page without an alert already selected.

PaulJKim:
@cmaddox5 Which comments are those? I'm not sure if I see/remember the ones you're referring to

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
```diff
@@ -20,6 +26,16 @@ const NewPaMessage = () => {
...
+  const onClearAssociatedAlert = () => {
+    setAssociatedAlert({} as Alert);
+    setVisualText("");
+    setPhoneticText("");
+    setEndWithEffectPeriod(false);
+    setImportLocations(false);
+    setImportMessage(false);
+    setEndDate(now.format("L"));
+    setEndTime(now.add(1, "hour").format("HH:mm"));
+  };
```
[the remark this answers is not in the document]
PaulJKim: Yeah conditionally clearing them makes sense to me I think. So just so I'm following clearly, you mean we would conditionally clearing the visual/phonetic text fields and the end date/time fields (and eventually locations) only if the field was imported?

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
```diff
@@ -0,0 +1,349 @@
...
+  setImportLocations: Dispatch<SetStateAction<boolean>>;
+  setImportMessage: Dispatch<SetStateAction<boolean>>;
+  setVisualText: Dispatch<SetStateAction<string>>;
+}
+
+const AssociateAlertPage = ({
+  associatedAlert,
+  endWithEffectPeriod,
+  importLocations,
+  importMessage,
```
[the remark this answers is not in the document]
PaulJKim: Hmm but don't these values need to be accessible in the parent since that's where the message and locations are defined too? If the values are local state variables in the `AssociateAlertPage` component, is there a way they can be accessed from or passed to the parent without using global state?

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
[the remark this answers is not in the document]
cmaddox5: Correct. Mostly just so no work is lost if the fields were not imported and modified since the alert was associated.

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
PaulJKim: @cmaddox5 going back to this one - I added another function `setAssociateAlertDefaults` for setting the default values of the import flags to `true` which I pass down to the button that takes the user to the associate alert page. If we just set the defaults in `NewPaMessage.tsx` as `false` and only set the flags to `true` once the user clicks navigate, then I think we can remove those additional conditions I added after your comment here.

assets/js/components/Dashboard/NewPaMessage/AssociateAlertPage.tsx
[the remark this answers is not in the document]
cmaddox5: Maybe these can be callbacks instead? So the parent will still be responsible for handling what happens when something is imported but won't have to keep track of whether an import happened. It makes sense that `endWithAffectPeriod` stays in the parent because it has a component that depends on that value, but these aren't needed once you navigate away. You would still have two props to pass but they could be something like:

```ts
onImportMessage={(alertMessage) => setVisualText(alertMessage)}
onImportLocations={() => ...}
```

But since importing locations is out of scope here, you could just leave it out for now and it can be added when zone selection is ready to add it.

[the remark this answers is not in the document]
PaulJKim: Oh interesting. It feels somewhat odd to me to have the flag and callback defined in two different places, but I do see how this simplifies things a bit!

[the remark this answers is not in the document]
PaulJKim: Ah but if we do things this way, I think that means we wouldn't be able to conditionally clear things out on "clear" since only the child would know whether or not a message/locations were imported.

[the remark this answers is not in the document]
cmaddox5: Yeah that's a good point. Maybe the simple solution is to only clear the associated alert and date import? Leave message and locations how they are and the user can do what they need to to get those updated. If they were to associate with a new alert and import those fields, they would be overwritten anyway.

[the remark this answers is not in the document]
PaulJKim: Sounds reasonable to me. Just pushed up the latest - let me know what you think

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
```diff
@@ -27,28 +25,16 @@ const NewPaMessage = () => {
...
-
-    if (importMessage) {
-      setVisualText("");
-      setPhoneticText("");
-    }
-
     setEndWithEffectPeriod(false);
-    setImportLocations(false);
-    setImportMessage(false);
+    setEndDate(now.format("L"));
```
cmaddox5, line 31: nit: I'll leave it up to you, but you might also be able to leave these out.

cmaddox5 (approved):
Looks great! Sorry it took so long to get through all this.

PR #399, opened 2024-07-16
Select stations page

cmaddox5:
This PR adds the Select Stations page and tweaks the `build_places.exs` script so we have the data we need to group signs by their stations and routes. 

---
  - https://app.asana.com/0/0/1207624689363633


sloanelybutsurely:
i was just taking a look at this on dev-green and it looks really great!

i did come across one issue with how the elements scroll. the horizontal scroll bar is hidden off screen and requires scrolling down fully to find.

https://github.com/user-attachments/assets/63fe70ea-755c-4b94-bf10-610b0e51ac23

if you take a look a asana's boards (mary's reference for the scrolling behavior of these columns) the horizontal scrollbar is at the bottom of the viewport and each column has its own vertical scrollbar.

and one quibble: the horizontal scroll bar is kind of floating away from the edge of the viewport. probably just a matter of swapping some margin for padding or vice-versa.

sloanelybutsurely:
okay one more thing actually:

so because these labels are only so large as to fit their contents, the clickable area for each of these options varies depending on the label. this is kind of a pain because if i want to click several items in a row i can't just click on one and move my cursor directly down; i have to adjust my cursor left and right also.

https://github.com/user-attachments/assets/702006b8-1aff-46fc-a596-327d0efe45c5

if the label instead grew to the remaining size of the column width (probably a flex container and `flex-grow: 1` on the labels) the clickable area of each of these would be consistent.

| Current | Proposed |
| - | - |
| image | image |

an aside: i noticed that the checkbox and label are siblings. if it helps, you can make the checkbox a child of the label and they will automatically be associated properly for screen readers / keyboard navigation without the need for `id` and `for`/`htmlFor` attributes.

assets/js/components/Dashboard/NewPaMessage/SelectStationsPage.tsx
```diff
@@ -0,0 +1,278 @@
...
+  "Green-E": ["Green-E"],
+  Red: ["Red"],
+  Orange: ["Orange"],
+  Blue: ["Blue"],
+  Mattapan: ["Mattapan"],
+  Silver: SILVER_LINE_ROUTES,
+};
+
+const SelectStationsPage = () => {
+  const [zones, setZones] = useState<string[]>([]);
```
sloanelybutsurely, line 49: **note:** this will need extracted to the parent component so state is maintained across sub-pages of the new pa/ess messsage flow. we can do that later though if you'd prefer

assets/js/components/Dashboard/NewPaMessage/SelectStationsPage.tsx
```diff
@@ -0,0 +1,278 @@
...
+                />
+              );
+            })}
+          </div>
+        </div>
+
+        {["Red", "Orange", "Blue", "Mattapan", "Silver"].map((route) => (
+          <div
+            key={route}
+            className={`route-col route-col--${route.toLowerCase()} col`}
```
sloanelybutsurely, line 249: **quibble:** i'm not a fan of dynamically building CSS classes because i think it makes it very difficult to determine where a particular CSS class is used in our markup.

**suggestion:** having a mapping of `route` to values that look like `route-col--red`, `route-col--blue`, etc. would make it so you could search for `route-col--red` and find where we use that value

config/runtime.exs
```diff
@@ -69,3 +69,11 @@ scheduler_jobs =
 config :screenplay, Screenplay.Scheduler, jobs: scheduler_jobs
 
 config :screenplay, Screenplay.Repo, pool_size: 10
+
+if config_env() in [:prod, :test] do
```
sloanelybutsurely, line 73: **suggestion:** we should be able to support local auth for `:test` also. probably just needs a `config, :screenplay, Screenplay.Repo, database: "screenplay_test"` in `config/test.exs` or something similar.

generating a new, empty phoenix project might be a helpful reference on how ecto is configured across environments by default

sloanelybutsurely (commented):
okay here are some of my initial comments. i've reviewed everything but ``SelectedStationsSummary.tsx`. i'll spend some time just on that file as requested!

assets/js/components/Dashboard/NewPaMessage/SelectedStationsSummary.tsx
```diff
@@ -0,0 +1,144 @@
...
+  numPlaces: number;
+  routeId: string;
+  onClick: () => void;
+}) => {
+  if (numPlaces === 0) return null;
+  return (
+    <div
+      className={classWithModifier("selected-group-tag", routeId.toLowerCase())}
+    >
+      {routeId}: {numPlaces} {pluralize("Station", numPlaces)}
```
sloanelybutsurely, line 30: **quibble/suggestion (non-blocking):** passing `true` as the third argument to `pluralize` will prefix the string with the passed numeric value.

```suggestion
      {routeId}: {pluralize("Station", numPlaces, true)}
```

assets/js/components/Dashboard/NewPaMessage/SelectedStationsSummary.tsx
```diff
@@ -0,0 +1,144 @@
...
+          screens: place.screens.filter((screen) =>
+            fp.includes(screen.id, value),
+          ),
+        };
+      }),
+      fp.filter((place) => place.screens.length > 0),
+    )(places);
+
+    setPlacesWithSelectedScreens(placesWithSelectedScreens);
+  }, [value]);
```
sloanelybutsurely, line 61: **suggestion:** a `useEffect` call which just computes a value and calls some `setThing` function can usually be replaced with a single `useMemo`

```typescript
  const placesWithSelectedScreens = useMemo<Place[]>(() => {
    return fp.flow(
      fp.map((place: Place) => {
        return {
          ...place,
          screens: place.screens.filter((screen) =>
            fp.includes(screen.id, value),
          ),
        };
      }),
      fp.filter((place) => place.screens.length > 0),
    )(places);
  }, [value]);
```
(formatting probably needs some help)

sloanelybutsurely (commented):
lgtm! going to leave approval to someone that hasn't had a large part in writing this code

assets/js/components/Dashboard/NewPaMessage/SelectedStationsSummary.tsx
cmaddox5: O nice. Guess it would be good for me to read the spec of the function fully as I bring it in 😅

assets/js/components/Dashboard/NewPaMessage/SelectedStationsSummary.tsx
cmaddox5: O sweet thank you.

config/runtime.exs
cmaddox5: Oof, thank you. Shouldn't have brute forced the solution into this file.

assets/js/components/Dashboard/NewPaMessage/SelectStationsPage.tsx
cmaddox5: Hmm yeah good point. Since the Cancel button reverts all changes back to whatever the initial state was (always empty here but when editing it will hold a list of zones) do we need to continue to keep state here and persist it to the parent after confirming the zones? Or is there a more elegant way to hold onto initial state?

sloanelybutsurely: I think we can take a list of `defaultZones` from the parent which we use to set a local value and then a callback to save to the parent zones when we save.

assets/js/components/Dashboard/NewPaMessage/RouteColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+  value,
+  onChange,
+  reverse = false,
+}: {
+  label: string;
+  routes: string[];
+  orderingRoute?: string;
+  places: Place[];
+  value: string[];
+  onChange: (zones: string[]) => void;
```
digitalcora, line 21: nit: It seems a little confusing that we refer to screen IDs as `zones` throughout these components (e.g. `routeZones`). The thing that we usually call a "zone" is just the zone identifier within the context of a single SCU, like `m` for Mezzanine. Something like `screenIds` or `signIds` might be clearer / more aligned with the terminology used elsewhere?

assets/js/components/Dashboard/NewPaMessage/SelectStationsPage.tsx
```diff
@@ -0,0 +1,298 @@
...
+      place.screens.flatMap((screen) => screen.route_ids ?? []),
+    ),
+  );
+
+  const busRoutes = fp.without(
+    Object.values(BASE_PLACE_ROUTE_TO_ROUTE_IDS).flat(),
+    allRoutes,
+  );
+
+  const PLACE_ROUTE_TO_ROUTE_IDS: { [key: string]: string[] } = {
```
digitalcora, line 77: nit: capitalization a bit misleading here since this is not really a "constant" in the same sense as the others are, it's built anew each time this component is rendered, with potentially different results.

config/test.exs
```diff
@@ -37,7 +37,10 @@ config :ueberauth_oidcc,
 config :screenplay, Screenplay.Repo,
   adapter: Ecto.Adapters.Postgres,
   database: "screenplay_test",
-  pool: Ecto.Adapters.SQL.Sandbox
+  pool: Ecto.Adapters.SQL.Sandbox,
+  username: "postgres",
+  password: "postgres",
```
digitalcora, line 42: If you install Postgres from Homebrew, the username is your own system username and the password is blank, so I don't think this is generally applicable (indeed when I run `mix test` on this branch I get a connection failure). What prompted these changes?

assets/css/dashboard/new-pa-message/selected-stations-summary.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  .label {
+    margin-right: 8px;
+  }
+
+  span {
+    color: $text-secondary;
+  }
+
+  .bi {
```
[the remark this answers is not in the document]
cmaddox5: It's actually a bootstrap style automatically added to icons in the `react-bootstrap-icons` package. This is a good call out because there's no reason I can't also add a custom class so we can track them down more easily.

config/test.exs
cmaddox5: Per Slack discussion, changes in this commit have been reverted. I will open a new PR this week proposing changes that work for all users regardless of how psql was installed.

assets/css/dashboard/new-pa-message/select-stations-page.scss
```diff
@@ -0,0 +1,199 @@
...
+        }
+      }
+    }
+
+    .route-col--red,
+    .route-col--mattapan,
+    .route-col--orange,
+    .route-col--blue,
+    .route-col--silver,
+    .route-col--bus {
```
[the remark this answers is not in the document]
cmaddox5: 💡

assets/js/components/Dashboard/NewPaMessage/RouteColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+  routes,
+  orderingRoute = routes[0],
+  places,
+  value,
+  onChange,
+  reverse = false,
+}: {
+  label: string;
+  routes: string[];
+  orderingRoute?: string;
```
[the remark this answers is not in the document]
cmaddox5: I agree for `routeIds` but `orderingRoute` is actually sometimes a consolidated label when we are ordering by bus. Maybe there's opportunity to tweak the sorting function so any route ID it doesn't recognize can just return the list unchanged. Then we could always use an ID in that prop and keep functionality the same.

assets/js/components/Dashboard/NewPaMessage/RouteColumn.tsx
cmaddox5: I agree, and I actually did make that change over in my Select Zones Page branch. I will go ahead and also make it here so that way we merge in what we want when this is approved.

assets/js/components/Dashboard/NewPaMessage/SelectStationsPage.tsx
cmaddox5: Good call. I also feel like the name doesn't accurately represent the value so renamed it to `routeNameToRouteIds`.

assets/css/dashboard/new-pa-message/select-stations-page.scss
```diff
@@ -0,0 +1,199 @@
+// Because this page is not a unique route, we cannot conditionally render
+// Sidebar the same way we do elsewhere.
+.sidebar-container:has(
+    + .page-content > .new-pa-message > .select-stations-page
+  ) {
+  display: none !important;
```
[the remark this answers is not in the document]
cmaddox5: This was a super interesting idea so I ran with it. Page looks the same as before but has all the functionality of a modal. The only thing I'm curious about a user accidentally pressing `Esc` and losing their progress. Luckily that piece is easy to take out (just remove the `onHide` prop).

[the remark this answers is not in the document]
digitalcora: Nice. Maybe if there's a concern about losing changes there could be a confirmation ("Really discard changes to selected zones?"). Though if we didn't already have a confirmation in the design for the Cancel button itself, I don't think pressing `Esc` poses more of a risk (if anything, it's harder to do that accidentally than it is for your mouse to slip and hit Cancel instead of Review Zones).

digitalcora (approved):
🚉 ☑️ ☑️ ☑️

PR #400, opened 2024-07-16
Switch to cache to avoid unnecessary fetch errors.

cmaddox5:
**Asana task**: ad-hoc

Came across [this](https://mbtace.sentry.io/issues/5611710385/events/bca6f7e2151b42e8958342e34b363ae2/) Sentry error today. After some digging, I saw that a timeout cause the issue. There really is no need to make a fetch here now that we have the cache. Switched the fetch to grab alerts from cache.

### DIFF
--- a/lib/screenplay_web/controllers/alerts_api_controller.ex
+++ b/lib/screenplay_web/controllers/alerts_api_controller.ex
@@ -2,17 +2,14 @@ defmodule ScreenplayWeb.AlertsApiController do
   use ScreenplayWeb, :controller
 
   alias Screenplay.Alerts.Alert
+  alias Screenplay.Alerts.Cache, as: AlertsCache
   alias Screenplay.Alerts.ScreensByAlert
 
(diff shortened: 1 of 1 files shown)


sloanelybutsurely (approved):

digitalcora:
While looking at this I noticed the cache module uses ETS with a mass-delete-and-insert method to update the table, a.k.a.

```elixir
:ets.delete_all_objects(:alerts)
:ets.insert(:alerts, alerts_to_insert)
```

Since _reads_ to the ETS table are not serialized through the GenServer, this means there's a (tiny, but frequently-occurring) possibility that a client could read the table after the delete but before the insert, and incorrectly get no results. We should keep an eye out for this pattern in case we see any strange issues. If read performance isn't a concern it might be fine to just serialize the reads through a genserver call, since only one message can be handled at a time, including the self-message used to update the table. Or we could use a "diff" approach based on alert IDs to update the table without clearing it.

PR #404, opened 2024-07-18
fix: When card QA

cmaddox5:
This PR addresses concerns brought up on the QA Miro board [here](https://miro.com/app/board/uXjVK0Nq3mE=/?moveToWidget=3458764594581098052&cot=14).

Major changes:
- Added an icon to date and time picker. This can be clicked to open the pickers. (Can still be opened by pressing Enter)
- Interval changes to the default of each priority when priority changes. Textbox can still be changed to a custom interval after changing priority.
- Removed min/max dates from date picker. User can put in any date they want and client will do the validation on form submission.
- Date picker colors have been improved and no longer crashes if an invalid date is typed into the input.

---
  - https://app.asana.com/0/0/1207867308989195


digitalcora:
Pulled this down to test it out, some notes:

* The picker buttons can be tab-focused and activated (e.g. with Enter/Space), but this doesn't place focus into the opened overlay, and the overlays don't appear to be very keyboard-friendly anyway (it's impossible to change the month in the calendar, for one). Since these are supplements for inputs that can be typed into directly, and the buttons present a bit of unexpected friction for someone trying to tab through the inputs, maybe it's fine to have the picker buttons excluded from the tab order. (`tabindex="-1"`) I'm also not sure why pressing Enter in the inputs opens the overlay; this has the same problems of the overlay not really being usable with the keyboard once you've opened it.

* The time fields allow pressing up/down arrow keys to respectively increase and decrease the hour/minute. When the picker overlay is open, this has the slightly weird result that pressing the _up_ arrow key results in the selection moving _down_ in the overlay (and vice-versa), because the value increases as you go "down" the list of hours/minutes. If we forego full keyboard support for the picker UI, I'm not sure there's a good way to address this, or maybe it's too much of an edge case anyway.

* "End of service" sets the time to 2:59am. This seems incorrect: both the start and end of service, in our model, are 3:00am. Which raises the question, should these even be separate links? Would someone expect that selecting "end of service" would (somehow) select 3:00am on the _following_ day from the one selected? I remember having a conversation about this many ~~moons~~ design reviews ago, but don't remember how it ended.

* The keyboard-focused state of dropdown menu items has the wrong colors, making the text almost invisible. <br> Screenshot 2024-07-22 at 2 39 54 PM

* I _think_ the descriptions of each priority level should have a slightly smaller font size than the number — I can't find a way to "open the dropdowns" in Figma, but see below for a screenshot-of-a-screenshot from the feedback in Miro (and I also remember it being this way in mockups we were shown). Compare this with the current appearance on this branch and you can see the font size is different. Maybe ask Mary for the specs on this. <br> Screenshot 2024-07-22 at 2 46 39 PM

assets/css/dashboard/date-picker.scss
```diff
@@ -1,9 +1,43 @@
 @import "react-calendar/dist/Calendar.css";
 
+.date-picker {
```
digitalcora, line 3: This and `.time-picker` seem to be duplicates of each other, which is quite a lot of lines to duplicate. Can we extract the common styles into a general "picker" class, maybe?

assets/js/components/Dashboard/DatePicker.tsx
```diff
@@ -1,43 +1,37 @@
...
-            setShowOverlay(true);
-          }
-        }}
-      />
+      <InputGroup className="date-picker" ref={ref}>
+        <Form.Control
+          id={id}
+          name={`${id}-input`}
+          value={selectedDate}
+          onChange={(input) => onChange(input.target.value)}
```
digitalcora, line 24: `input` seems like an inaccurate name for this argument — usually it's called `event`, as it's the [event object](https://developer.mozilla.org/en-US/docs/Web/API/Event) for the change event. (The `event.target` is what would actually be the input element.)

assets/js/components/Dashboard/TimePicker.tsx
```diff
@@ -13,28 +21,31 @@ interface TimePickerProps {
...
-        onKeyUp={(e) => {
-          if (e.key === "Enter") {
-            setShowOverlay(true);
+      <InputGroup className="time-picker" ref={ref}>
+        <Form.Control
+          id={id}
+          type="time"
+          value={selectedTime}
+          onChange={(input) =>
+            onChange(moment(input.target.value, "HH:mm").format("HH:mm"))
```
digitalcora, line 34: What does this do? It appears we're parsing a time from the input value, then formatting it, but not doing anything with it after that.

assets/js/components/Dashboard/TimePicker.tsx
```diff
@@ -13,28 +21,31 @@ interface TimePickerProps {
...
+          }}
+        />
+        <Button onClick={() => setShowOverlay(true)}>
+          <Clock />
+        </Button>
+      </InputGroup>
       <Overlay
         rootClose
-        rootCloseEvent="mousedown"
+        rootCloseEvent="click"
```
digitalcora, line 48: `DatePicker` uses `rootCloseEvent="mousedown"` — I'm guessing these are not supposed to be different?

assets/js/components/Dashboard/TimePicker.tsx
cmaddox5: This is unnecessary. Can just pass the value directly into `onChange`.

digitalcora: Oh, whoops, I was missing that this was wrapped in `onChange` which is a reference to the component's prop.

assets/js/components/Dashboard/TimePicker.tsx
cmaddox5: This was an unnecessary change. Reverting.

cmaddox5:
- How keyboard navigation should work for the pickers is something I've struggled with. Excluding the button from tab order is an interesting idea though. As for why Enter opens the overlay, that shouldn't be needed anymore. When initially working on this, I forgot the icons so there was no obvious way to open the overlay. 
- Wow, I missed that and it does feel weird. Similar to above, I'll take this to design cause maybe keyboard support is causing more harm than good.
- Good point, I should have also questioned this when implementing. If those buttons were for full datetimes, they may make more sense. Since they only pick time, differentiating between start and end shouldn't be necessary. Will take this to design.
- Was able to track down the design breakdown of these buttons in Figma, but I don't see a focused state. Need to ask for those.
- I implemented this incorrectly. Value should be 16px and label should be 14px.

digitalcora:
> Was able to track down the design breakdown of these buttons in Figma, but I don't see a focused state. Need to ask for those.

Since we have a functioning focus rectangle already I don't think it needs to be any different from the "normal" state. I'm guessing the problem is we overrode the default styles for the dropdown items but didn't also override the styles for the `:focus` pseudo-class.

cmaddox5:
@digitalcora After some discussions with Mary and Sloane, Mary decided to throw out the custom date and time picker components and just let the native inputs do that work. Should help alleviate concerns folks have had during QA and code review.

assets/js/components/Dashboard/NewPaMessagePage.tsx
```diff
@@ -130,13 +149,27 @@ const NewPaMessagePage = () => {
...
+                    type="time"
+                    className="time-picker picker"
+                    value={endTime}
+                    onChange={(event) => setEndTime(event.target.value)}
                   />
-                  <TimePicker selectedTime={endTime} onChange={setEndTime} />
+                  <Button
+                    className="service-time-link"
+                    variant="link"
+                    onClick={() => setEndTime("02:59")}
```
digitalcora, line 169: I think this should also be `03:00`, per [this thread](https://mbta.slack.com/archives/C06211Q5YAZ/p1721932270002629) and the updated [Start/End of Service](https://www.notion.so/mbta-downtown-crossing/Start-of-Service-End-of-Service-change-dcf3f99e296e4f02aeaeab362b495119) guidance.

digitalcora (approved):
Looking good! The native calendar picker experience is surprisingly nice, in Firefox anyway.

assets/js/components/Dashboard/NewPaMessagePage.tsx
cmaddox5: Ahh, thanks for the link!

PR #410, opened 2024-07-30
feat: Select Zones Page

cmaddox5:
This PR adds the Select Zones Page. Design is [here](https://www.figma.com/design/P6RZ1AJWk7P98D6IFDK9Ik/PA%2FESS-Modernization?node-id=8569-31857&t=pmjL4MhDj3NyW4Rq-4). Logic breakdown is [here](https://miro.com/app/board/uXjVKEDY9Hw=/). 

The labels for the buttons depends on some local data being updated. Problem there is that custom labels for signs lives in `mbta.ts` in `signs_ui`. I'm trying to think through how to automate pulling out those labels because manually extracting those labels is a huge pain. We may need to address the issue with PA/ESS configs living in multiple files across multiple sources now that Screenplay needs to read more of the files.

---
  - https://app.asana.com/0/0/1207624689363634


cmaddox5:
@sloanelybutsurely I'm not really sure what happened there. I tried redoing the merge and for some reason it keeps coming back. Seems easier to just do another one-off revert.

assets/js/components/Dashboard/NewPaMessage/SelectedSignsByRouteTags.tsx
```diff
@@ -0,0 +1,114 @@
...
+                screen.route_ids?.some((r) => r.startsWith(routeId)),
+              ),
+            ).length
+          }
+          routeId={routeId}
+          onClick={() => {
+            removeSelectedScreens(
+              (screen) =>
+                screen.route_ids?.some((r) => r.startsWith(routeId)) ?? false,
+            );
```
sloanelybutsurely, line 73: **suggestion:** i feel like `removeSelectScreens` needs a different name. it took me a while to figure out what was going on here. probably something that describes the action it takes in relation to the arguments passed to it.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectStationsAndZones.tsx
```diff
@@ -0,0 +1,56 @@
...
+  onError,
+}: Props) => {
+  const dispatch = useScreenplayDispatchContext();
+  const [signIds, setSignIds] = useState(value);
+
+  useEffect(() => {
+    dispatch({ type: "SHOW_SIDEBAR", showSidebar: false });
+
+    return () => dispatch({ type: "SHOW_SIDEBAR", showSidebar: true });
+  }, []);
```
sloanelybutsurely, line 34: **question:** this is kind of odd to me. is this because the select stations and zones stuff is inside a modal now? what's the reason for the modal? (heads up: i have a bias against modals 🙂)

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+import { Page } from "../types";
+import { Button } from "react-bootstrap";
+import { Place } from "Models/place";
+import fp from "lodash/fp";
+import { Screen } from "Models/screen";
+import { getZoneLabel, signIDs, sortByStationOrder } from "../../../../util";
+import cx from "classnames";
+import { ArrowLeftShort, ArrowRightShort, Dot } from "react-bootstrap-icons";
+import { useRouteToRouteIDsMap } from "../hooks";
+import { Dictionary } from "lodash";
```
sloanelybutsurely, line 17: **suggestion:** is this just a type import? can we make it a type import then?

```suggestion
import type { Dictionary } from "lodash";
```

or

```suggestion
import { type Dictionary } from "lodash";
```

should both work

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+      case "Blue":
+        return {
+          left: "Westbound",
+          middle: "Westbound & Eastbound",
+          right: "Eastbound",
+        };
+      case "Silver":
+        return { left: "To South Station/Downtown", right: "Eastbound" };
+      case "Bus":
+        return { left: "", middle: "", right: "" };
```
sloanelybutsurely, line 73: **question:** does this leave the directions unlabeled if you select only bus on the select stations page?

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+    if (selectedRouteFilter === "Silver") {
+      return {
+        left: "Both",
+        middle: "South Station/Downtown",
+        right: "Outbound Content",
+      };
+    }
+
+    return {
+      left: "All Zones (except bus)",
```
sloanelybutsurely, line 93: **question:** is this still accurate if you select only bus from the select stations page?

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+    setSelectedRouteFilter(Object.keys(routes)[0]);
+    setNewPlaces(
+      places.filter((p) =>
+        places
+          .filter((p) => p.screens.some((s) => value.includes(s.id)))
+          .map((p) => p.id)
+          .includes(p.id),
+      ),
+    );
+  }, [places]);
```
sloanelybutsurely, line 133: **note:** a `useEffect` that has no effects beyond calling state setter functions usually means that those state values should instead be derived from the values in the dependencies array of `useEffect`.

if you want these to be default values you can compute them in the calls to `useState` ([lazily even!](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state)).

and finally, if your concern is that `places` isn't the value you need during the first render, you can wrap this component in another component that simply retrieves `places` and waits to render `<SelectZonesPage />` when `places` has the value you're looking for.

happy to walk through this with you if that's helpful!

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+
+  const filteredPlaces = useMemo(() => {
+    return newPlaces.filter((p) =>
+      p.screens.some((s) =>
+        s.route_ids?.some((r) =>
+          routeToRouteIDMap[selectedRouteFilter]?.some((a) => r === a),
```
sloanelybutsurely, line 139: **note:** i feel like i am seeing this pattern _a lot_. might be a good opportunity for a helper function.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+      "Green",
+      "Green-B",
+      "Green-C",
+      "Green-D",
+      "Green-E",
+      "Blue",
+      "Red",
+      "Orange",
+      "Mattapan",
+    ].includes(selectedRouteFilter);
```
sloanelybutsurely, line 198: **question:** why is this a thunk?

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+    const newScreens = groups.flatMap((group) =>
+      group.flatMap((zone) => signIDs(screensByZone[zone] ?? [])),
+    );
+
+    onChange(fp.uniq([...signsWithoutSelectedRoute, ...newScreens]));
+  };
+
+  const isAllSelected =
+    value.length > 0 &&
+    areAllGroupsSelected([LEFT_ZONES, MIDDLE_ZONES, RIGHT_ZONES]);
```
sloanelybutsurely, line 181: **suggestion:** if feels like the contents of `areAllGroupsSelected` could be in-lined here

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+      "Green-D",
+      "Green-E",
+      "Blue",
+      "Red",
+      "Orange",
+      "Mattapan",
+    ].includes(selectedRouteFilter);
+
+  const massSelectClassName = (selected: boolean) =>
+    selected ? "button-primary" : "button-secondary-outline";
```
sloanelybutsurely, line 201: **quibble:** i would prefer something that acted as an `active` class on a button that only included the properties required to make the given button appear active instead of having to swap out two different kinds of button classes.

meaning you could have a class list like `button-primary` and when it was active you would have `button-primary button-active`.

then you could use the `classnames` library to conditionally set the active class only like so

```javascript
className={classnames('button-primary', { ['button-active']: selected })}
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+
+  const getFilterPillLabel = () => {
+    if (selectedRouteFilter.startsWith("Green-")) {
+      return `Green Line ${selectedRouteFilter.split("-")[1]}`;
+    } else {
+      return `${selectedRouteFilter} Line`;
+    }
+  };
```
sloanelybutsurely, line 209: **note:** another thunk here...

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+            Done
+          </Button>
+        </div>
+      </div>
+      <div className="zone-selection">
+        <div className="filters-container">
+          <div>Service type</div>
+          <div className="filters">
+            {Object.keys(selectedRoutes).map((routeID) => {
+              let branchFilters;
```
sloanelybutsurely, line 248: **suggestion:** no need for a mutable reference here. a ternary would allow this to be a `const` and colocate the value of `branchFilters` with its declaration.

```javascript
const branchFilters = routeID === "Green" ? selectedRoutes["Green"].sort().etc().etc() : null
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+
+                    return (
+                      <Button
+                        key={`branch-filter-${branch}`}
+                        onClick={() => setSelectedRouteFilter(branchID)}
+                        className={cx(
+                          "filter-button",
+                          selectedRouteFilter === branchID
+                            ? ROUTE_TO_CLASS_NAMES_MAP["Green"]
+                            : "filter-button-default",
```
sloanelybutsurely, line 263: **quibble:** i would prefer to use the cascade here so that you could write something like this instead:

```javascript
className={cx('filter-button', {
  [[ROUTE_TO_CLASS_NAMES_MAP["Green"]]: selectedRouteFilter === branchID
})}
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+              );
+            })}
+          </div>
+        </div>
+        <div className="zones-table-container">
+          <div className="container-header">
+            <div className="title-and-route-container">
+              <div className="title h3">Zones</div>
+              <div
+                className={`route ${ROUTE_TO_CLASS_NAMES_MAP[selectedRouteFilter.split("-")[0]]}`}
```
sloanelybutsurely, line 296: **quibble:** i would prefer using `cx` to build this class list.

```suggestion
                className={cx("route", ROUTE_TO_CLASS_NAMES_MAP[selectedRouteFilter.split("-")[0]])}
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+  allSelectedSigns: string[];
+  setSignIds: (signs: string[]) => void;
+  route: string;
+  routeToRouteIDMap: { [key: string]: string[] };
+}) => {
+  const allSignsForRouteAtPlace = place.screens.filter((screen) =>
+    fp.any(
+      (r) => routeToRouteIDMap[route]?.some((a) => r.startsWith(a)),
+      screen.route_ids,
+    ),
```
sloanelybutsurely, line 386: **quibble:** could be replaced with `screen.route_ids.some(...)`

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+  );
```
sloanelybutsurely, line 387: **quibble:** it feels strange to me that this component is calculating these values instead of being passed them. i feel like this component should be passed lists of left, middle, and right zones maybe?

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+                  onSignButtonClick(sign.id);
+                }}
+                isSelected={allSelectedSigns.includes(sign.id)}
+              >
+                {sign.label ?? getZoneLabel(sign.zone ?? "")}{" "}
+                <ArrowRightShort />
+              </SelectSignButton>
+            );
+          })}
+        </div>
```
sloanelybutsurely, line 514: **suggestion:** this "sign button group" is duplicated for each cell. feels like a good opportunity to extract a "dumb" component.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+      </td>
+    </tr>
+  );
+};
+
+interface SelectSignButtonProps {
+  isSelected: boolean;
+  onClick: () => void;
+  children: ReactNode;
+}
```
sloanelybutsurely, line 524: **suggestion:** if you need a `children` prop i'd suggest using `React.PropsWithChildren`

```suggestion
type SelectSignButtonProps = React.PropsWithChildren<{
  isSelected: boolean;
  onClick: () => void;
}>;
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+    <Button
+      className={isSelected ? "button-primary" : "button-primary-outline"}
+      onClick={onClick}
+    >
+      {children}
+    </Button>
+  );
+};
+
+export default SelectZonesPage;
```
sloanelybutsurely, line 541: 541 lines feels like too many to me. what do you think?

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectStationsAndZones.tsx
cmaddox5: No, they actually aren't in modals anymore. The problem is the conditional rendering of the Sidebar involves looking at the current route. Since both pages rendered here do not have a unique route, we can't hide the Sidebar in the same way we do everywhere else. This was kind of a "best way to make the conditional work without changing the structure of the app" way of tackling the problem. Very much open to hearing other suggestions.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: It does, yes, and that's by design: https://miro.com/app/board/uXjVKEDY9Hw=/?moveToWidget=3458764592365724172&cot=14

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: These buttons are not visible when the selected route filter is `Bus`.

sloanelybutsurely:
sorry i meant to include this in my review.

i've noticed there's a mix of `useMemo`, direct computation of composite values, and some thunks. i think we should probably start with direction computation over thunks and only if we really think a value is particularly expensive to calculate _or_ the value is more complicated than a single expression, should we reach for `useMemo`.

assets/js/components/Dashboard/NewPaMessage/hooks.tsx
```diff
@@ -0,0 +1,49 @@
...
+      fp.map((place: Place) => {
+        return {
+          ...place,
+          screens: place.screens.filter((screen) =>
+            fp.includes(screen.id, signs),
+          ),
+        };
+      }),
+      fp.filter((place) => place.screens.length > 0),
+    )(places);
```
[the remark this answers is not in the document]
cmaddox5: Good idea. I feel like I tunneled on lodash a bit too much after I read the docs...

assets/js/components/Dashboard/NewPaMessage/SelectedSignsByRouteTags.tsx
cmaddox5: Something like `removeFilteredScreens` maybe? Or `removeScreensInFilter`? I'll commit something and see what you think.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: A bit unnecessary. Going to remove it.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: When this was just an if with an else like it is here, I agree that this was unnecessary. I need to add an additional conditional so a simple ternary won't cut it. I could do a nested ternary but that feels harder to read. Do you have a preference for a conditional like this?

sloanelybutsurely: You could use an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) or `useMemo`.

assets/js/components/Dashboard/NewPaMessage/SelectedSignsByRouteTags.tsx
sloanelybutsurely: filterSelectedScreens? rejectSelectedScreens? How would you describe what this function does?

assets/js/components/Dashboard/NewPaMessage/hooks.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const useRouteToRouteIDsMap = (): { [key: string]: string[] } => {
+  const places = usePlacesWithPaEss();
+  return useMemo(() => {
+    return {
+      ...BASE_ROUTE_NAME_TO_ROUTE_IDS,
+      Bus: busRouteIdsAtPlaces(places),
+    };
+  }, []);
```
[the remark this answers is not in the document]
cmaddox5: Looks like CI passed so I would say no. Can look into adding that.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: I think my original goal of this hook was to bring together all values that used the same dependency array. In hindsight, that seems unnecessary considering `selectedRoutes` and `newPlaces` (terrible name so will think of a better one) can use `useMemo` and `selectedRouteFilter` can initial to whatever `selectedRoutes` evaluates to in the `useMemo`.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: Maybe the function name is a bit misleading because "All" here refers to all parameters instead of all groups. I can rename for better clarity.

assets/css/dashboard/new-pa-message/new-pa-message.scss
```diff
@@ -14,14 +15,14 @@
...
-    color: $text-link-primary;
+    margin-right: 16px;
     border: none;
-  }
 
-  .submit-button {
-    background-color: $button-primary;
-    color: $button-secondary;
+    &:hover {
+      background-color: #c1e4ff14;
```
sloanelybutsurely, line 24: **question:** i see there are several in-line color values throughout the SCSS changes. do these colors have names we could use instead?

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/PlaceZonesRow.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const onSignButtonClick = (signID: string) => {
+    if (allSelectedSigns.includes(signID)) {
+      setSignIds(fp.without([signID], allSelectedSigns));
+    } else {
+      setSignIds(fp.uniq([...allSelectedSigns, signID]));
+    }
+  };
+
+  const allSignsSelected =
+    selectedSignsAtPlace.length === allSignsForRouteAtPlace.length;
```
sloanelybutsurely, line 48: **question:** is this always true?

sloanelybutsurely (approved):

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/PlaceZonesRow.tsx
cmaddox5: No, `allSignsForRouteAtPlace` contains selected and unselected signs that match the current route filter. `selectedSignsAtPlace` is all of the selected signs from `allSignsForRouteAtPlace`. So when a sign is unselected, it becomes `false`.

assets/css/dashboard/new-pa-message/new-pa-message.scss
cmaddox5: Some colors do not have names, but I can create variables based on the button class defined by Mary.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/PlaceZonesRow.tsx
sloanelybutsurely: oh sorry my question was unclear. i meant it is always true that if these have the same length all signs are selected?

cmaddox5: It is, but I can see how the variable implies something different. `selectedSignsAtPlace` is a bit misleading because it's only the selected signs for the filtered route. I can change that.

PR #412, opened 2024-08-01
fix: Script updates for route filtering and formatting

cmaddox5:
**Asana task**: ad-hoc

I noticed a new row in screenplay-dev-green for Haverhill:
Screenshot 2024-08-01 at 5 08 30 PM

My investigation into this lead to me finding that Haverhill CR is being replaced by a shuttle next rating: https://api-dev-green.mbtace.com/routes?filter[stop]=place-WR-0329. This pushed me to track down how we filter out places in the script and I realized how I did it originally was... unstable. Instead of relying on whether the route ID is a number (why would I do that?), we now look at either route type or route ID. 

I also found that we were removing CR/SL/Bus only places way too early. Because they were removed after only Showtime screens were added to them, we were actually neglected to show a lot of places that have countdown clocks.

### DIFF
--- a/scripts/build_places.exs
+++ b/scripts/build_places.exs
@@ -620,7 +620,7 @@ merged_paess =
   end)
   # Get rid of CR and Bus stops with no screens
   |> Enum.reject(fn %{routes: routes, screens: screens} ->
-    cr_or_bus_only?.(routes) and length(screens) == 0
+    cr_or_bus_only?.(routes) and Enum.empty?(screens)
   end)
 
 Logger.info("Writing result to priv/config/places_and_screens.json")
--- a/scripts/build_places.exs
+++ b/scripts/build_places.exs
@@ -51,6 +51,8 @@ get_config = fn bucket, path ->
   end
 end
 
+sl_route_ids = ~w[741 742 743 746 749 751]
+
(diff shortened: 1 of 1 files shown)


scripts/build_places.exs
```diff
@@ -598,9 +614,14 @@ pa_ess_screens =
...
 # Merge screens and pa/ess
 merged_paess =
-  Enum.map(contents, fn %{id: id, screens: screens} = place ->
+  contents
+  |> Enum.map(fn %{id: id, screens: screens} = place ->
     Map.put(place, :screens, screens ++ (pa_ess_screens[id] || []))
   end)
+  # Get rid of CR and Bus stops with no screens
+  |> Enum.reject(fn %{routes: routes, screens: screens} ->
+    cr_or_bus_only?.(routes) and length(screens) == 0
```
sloanelybutsurely, line 623: **quibble (non-blocking):** i find `Enum.empty?/1` a bit easier to follow. bonus! it will be slightly faster because it will check for `value == []` instead of iterating over the entire length of the list.

```suggestion
    cr_or_bus_only?.(routes) and Enum.empty?(screens)
```

sloanelybutsurely (approved):

PR #415, opened 2024-08-05
chore: remove "Search" and "Advanced Search" buttons

sloanelybutsurely:
Comments out these buttons because they don't do anything yet. They can be shown again when the features they are intended link to are built.

### DIFF
--- a/assets/js/components/Dashboard/PaMessagesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessagesPage.tsx
@@ -33,12 +33,13 @@ const PaMessagesPage: ComponentType = () => {
                   Add New
                 </Link>
               </Col>
-              <Col className="pa-message-table-action-bar__search">
+              {/* Can remove for now because this will not be implemented until after MVP launch */}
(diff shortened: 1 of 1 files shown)


assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -33,12 +33,13 @@ const PaMessagesPage: ComponentType = () => {
                   Add New
                 </Link>
               </Col>
-              <Col className="pa-message-table-action-bar__search">
+              {/* Disabled as these features are not yet implemented */}
```
cmaddox5, line 36: suggestion: Can remove for now because this will not be implemented until after MVP launch.

cmaddox5 (approved):

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: Are you suggesting I reword this comment or that I remove the code completely?

PR #418, opened 2024-08-05
chore: Split functions in `PaMessage` model into separate modules

sloanelybutsurely:
Moves functions from `Screenplay.PaMessages.PaMessage` schema module into `Screenplay.PaMessages` context module. Moves query into `*.Queries` module.

Also adds a typespec type for `%PaMessage{}` structs


lib/screenplay/pa_messages/pa_message.ex
```diff
@@ -2,16 +2,27 @@ defmodule Screenplay.PaMessages.PaMessage do
...
-  alias Screenplay.{Repo, Util}
-
   use Ecto.Schema
 
   @derive {Jason.Encoder, except: [:__meta__]}
 
+  @type t() :: %__MODULE__{
+          alert_id: String.t() | nil,
+          start_time: DateTime.t(),
+          end_time: DateTime.t(),
```
cmaddox5, line 12: This can also be `nil` as long as `alert_id` is not `nil`.

cmaddox5 (approved):
Perfect timing. Was just about to start on the insert logic here.

PR #421, opened 2024-08-07
feat: New Messages backend

cmaddox5:
This PR adds an endpoint that can be used to create new PA messages. Although the client should catch any validation errors before the DB insert, I thought it was appropriate to add some changeset validation.

### DIFF
--- a/lib/screenplay/pa_messages/pa_message.ex
+++ b/lib/screenplay/pa_messages/pa_message.ex
@@ -75,7 +75,7 @@ defmodule Screenplay.PaMessages.PaMessage do
     alert_id = get_field(changeset, :alert_id)
 
     if is_nil(end_time) and is_nil(alert_id) do
-      add_error(changeset, :end_time, "can't be nil if alert_id is nil")
+      add_error(changeset, :end_time, "cannot have an end time if associated with an alert")
     else
       changeset
     end
--- a/lib/screenplay/pa_messages/pa_message.ex
+++ b/lib/screenplay/pa_messages/pa_message.ex
@@ -3,6 +3,7 @@ defmodule Screenplay.PaMessages.PaMessage do
   Represents a PA Message that will be retrieved by RTS to play audio in stations.
   """
   use Ecto.Schema
+  import Ecto.Changeset
 
   @derive {Jason.Encoder, except: [:__meta__]}
 
@@ -38,5 +39,33 @@ defmodule Screenplay.PaMessages.PaMessage do
     field(:message_type, :string)
 
     timestamps(type: :utc_datetime)
+  end
+
+  def changeset(message, attrs \\ %{}) do
+    message
+    |> cast(attrs, [
+      :alert_id,
+      :start_time,
+      :end_time,
+      :days_of_week,
+      :sign_ids,
+      :priority,
+      :interval_in_minutes,
+      :visual_text,
+      :audio_text,
+      :paused,
+      :saved
+    ])
+    |> validate_required([
+      :start_time,
+      :days_of_week,
+      :sign_ids,
+      :priority,
+      :interval_in_minutes,
+      :visual_text,
+      :audio_text
+    ])
+    |> validate_length(:sign_ids, min: 1)
+    |> validate_subset(:days_of_week, 1..7)
   end
 end
--- a/lib/screenplay/pa_messages/pa_message.ex
+++ b/lib/screenplay/pa_messages/pa_message.ex
@@ -67,5 +67,17 @@ defmodule Screenplay.PaMessages.PaMessage do
     ])
     |> validate_length(:sign_ids, min: 1)
     |> validate_subset(:days_of_week, 1..7)
+    |> validate_end_date()
+  end
+
+  defp validate_end_date(changeset) do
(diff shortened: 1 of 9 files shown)


lib/screenplay/pa_messages/pa_message.ex
```diff
@@ -39,4 +40,32 @@ defmodule Screenplay.PaMessages.PaMessage do
...
+      :days_of_week,
+      :sign_ids,
+      :priority,
+      :interval_in_minutes,
+      :visual_text,
+      :audio_text
+    ])
+    |> validate_length(:sign_ids, min: 1)
+    |> validate_subset(:days_of_week, 1..7)
+  end
```
sloanelybutsurely, line 70: **question:** should we validate that `end_time` is `nil` if `alert_id` is not `nil`?

lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -20,4 +22,15 @@ defmodule ScreenplayWeb.PaMessagesApiController do
...
     end
   end
+
+  def create(conn, params) do
+    case PaMessages.create_message(params) do
+      {:ok, _} ->
+        json(conn, %{success: true})
+
+      {:error, changeset} ->
+        Logger.error("[pa_message create] #{Util.format_changeset_errors(changeset)}")
```
sloanelybutsurely, line 32: **note:** i'm pretty sure sentry will capture things logged at the `:error` level. we probably don't want any validation error at the API level to create a sentry issue

lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -20,4 +22,15 @@ defmodule ScreenplayWeb.PaMessagesApiController do
...
+        send_resp(conn, 500, "Could not create message")
```
sloanelybutsurely, line 33: **issue:** a `500` status code implies an internal server error. we probably want something in the 400-499 range (i'd go for `422`).

there's also a great guide in the phoenix docs that covers setting up the fallback controller to render out changeset errors that i think we should do: https://hexdocs.pm/phoenix/json_and_apis.html#fallbackcontroller-and-changesetjson

lib/screenplay_web/router.ex
```diff
@@ -89,6 +89,7 @@ defmodule ScreenplayWeb.Router do
     get("/pa-messages/new", PaMessagesController, :index)
     get("/api/pa-messages", PaMessagesApiController, :index)
     get("/api/pa-messages/preview_audio", PaMessagesApiController, :preview_audio)
+    post("/api/pa-messages/create", PaMessagesApiController, :create)
```
sloanelybutsurely, line 92: **quibble:** i would expect the endpoint to be `POST  /api/pa-messages`

lib/screenplay/pa_messages/pa_message.ex
cmaddox5: Good call. Will add.

lib/screenplay_web/controllers/pa_messages_api_controller.ex
cmaddox5: Ahh, didn't think about that. Will change to `:warning`.

lib/screenplay_web/controllers/pa_messages_api_controller.ex
cmaddox5: Very interesting. Thanks for the link. Took me down a nice educational rabbit hole. I think I managed to get this working for us.

lib/screenplay_web/controllers/pa_messages_api_controller.ex
cmaddox5: Scratch that. With the new fallback controller, we shouldn't need to log anything.

lib/screenplay/pa_messages/pa_message.ex
```diff
@@ -39,4 +40,44 @@ defmodule Screenplay.PaMessages.PaMessage do
...
+    |> validate_subset(:days_of_week, 1..7)
+    |> validate_end_date()
+  end
+
+  defp validate_end_date(changeset) do
+    end_time = get_field(changeset, :end_time)
+    alert_id = get_field(changeset, :alert_id)
+
+    if is_nil(end_time) and is_nil(alert_id) do
+      add_error(changeset, :end_time, "can't be nil if alert_id is nil")
```
sloanelybutsurely, line 78: **suggestion:** can we make this a more human friendly message? how about this?

```suggestion
      add_error(changeset, :end_time, "cannot have an end time if associated with an alert")
```

lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -1,6 +1,10 @@
 defmodule ScreenplayWeb.PaMessagesApiController do
   use ScreenplayWeb, :controller
 
+  require Logger
+
```
sloanelybutsurely, line 5: **quibble:** doesn't look like you're using the Logger in here. can we remove this?

```suggestion
```

sloanelybutsurely (approved):
a couple of small things. no need for another pass

PR #422, opened 2024-08-07
feat: support filtering by message state (past, active, future)

sloanelybutsurely:
Updates the `GET /api/pa-messages` endpoint to filter the returned PA Messages based on provided query parameters.

So far the only supported parameters are `state` and `now`. `state` can be used to limit the returned PA Messages to only those which are "past", "active", or "future". `now` is mostly a testing convience but allows passing in a timestamp to be treated as the current time when determining which state a PA Message is currently in.


cmaddox5 (approved):
Very cool solution. Haven't seen this pattern before. Looks great!

PR #423, opened 2024-08-08
chore: update phoenix view / template structure

sloanelybutsurely:
Moves views and templates out of `templates/` and `views/` into `controllers/*_html/*`. This is how new Phoenix applications are generated. Updating to use this format means that following Phoenix guides is easier and also keeps up up-to-date with the latest Phoenix practices.


cmaddox5 (approved):
Very cool, thanks for taking care of this!

PR #427, opened 2024-08-08
feat: support filtering PaMessages by route

sloanelybutsurely:
Derives a mapping of route id to pa/ess sign ids from the "places and screens" data.

Uses the mapping of routes to PA/ESS sign ids to expand the the sign filtering to include all signs for all of the routes included in the options passed.

Required adding a `route_ids` field to the `PaEssScreen` struct.

### DIFF
--- a/lib/screenplay/pa_messages.ex
+++ b/lib/screenplay/pa_messages.ex
@@ -7,6 +7,7 @@ defmodule Screenplay.PaMessages do
   import Ecto.Query
 
   alias Screenplay.Alerts.Cache, as: AlertsCache
+  alias Screenplay.Config.RoutesToSigns
   alias Screenplay.PaMessages.PaMessage
   alias Screenplay.Repo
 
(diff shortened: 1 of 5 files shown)


lib/screenplay/pa_messages.ex
```diff
@@ -19,7 +20,8 @@ defmodule Screenplay.PaMessages do
     @type t :: %{
             optional(:state) => :all | :active | :future | :past,
             optional(:now) => DateTime.t(),
-            optional(:signs) => [String.t(), ...]
+            optional(:signs) => [String.t(), ...],
+            optional(:routes) => [String.t(), ...]
```
cmaddox5, line 24: Just to clarify, it will be the client's responsibility to break down the labels `Busway` and `Silver` into their specific route IDs?

cmaddox5 (approved):
Looks good! Just one non-blocking question.

lib/screenplay/pa_messages.ex
sloanelybutsurely: Probably! We could add some aliases on the server if we find that's helpful but probably its own thing imo.

PR #429, opened 2024-08-09
feat: adds UI to filter PA messages on their current state

sloanelybutsurely:
- Installs `swr` for data fetching
- Wires up page query param to local state for persisted filters
- Moves some button and button group styles into own stylesheets
- Adds filters to PA Messages page to select which state to limit the displayed PA messages to. Clicking a filter that is already applied clears the filter.

https://github.com/user-attachments/assets/3d3c7a59-2527-4e1a-8db6-ee8808f3a689

<details>
<summary>Loading state 📹 </summary>

https://github.com/user-attachments/assets/fb93e139-8802-4f9c-a57d-31203367ac14

</details>

---

https://swr.vercel.app/

### DIFF
--- a/assets/js/components/Dashboard/PaMessagesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessagesPage.tsx
@@ -1,20 +1,75 @@
-import React, { ComponentType, useState, useEffect } from "react";
-import { Container, Row, Col, FormCheck } from "react-bootstrap";
+import React, { ComponentType, useState, useMemo, useEffect } from "react";
+import {
+  Container,
+  Row,
+  Col,
+  FormCheck,
+  ButtonGroup,
+  Button,
+  Spinner,
+} from "react-bootstrap";
 import { PlusCircleFill } from "react-bootstrap-icons";
-import { fetchPaMessages } from "Utils/api";
 import { PaMessage } from "Models/pa_message";
-import { Link } from "react-router-dom";
+import { Link, useSearchParams } from "react-router-dom";
+import cx from "classnames";
+import useSWR from "swr";
+
+type StateFilter = "active" | "future" | "past";
+
+const fetcher = (...args: Parameters<typeof fetch>) =>
+  fetch(...args).then((resp) => resp.json());
+const usePaMessages = ({
+  stateFilter,
+}: {
+  stateFilter?: StateFilter | null;
+}) => {
+  const url = useMemo(() => {
+    const params = new URLSearchParams();
+
+    if (stateFilter) params.set("state", stateFilter);
+
+    return `/api/pa-messages?${params.toString()}`;
+  }, [stateFilter]);
+
+  return useSWR<PaMessage[]>(url, fetcher, { keepPreviousData: true });
+};
+
+const useDelayedLoadingState = (value: boolean, delay = 250) => {
+  const [delayedLoadingState, setDelayedLoadingState] =
+    useState<boolean>(value);
+  useEffect(() => {
+    if (value) {
+      const timeout = setTimeout(() => {
+        setDelayedLoadingState(value);
+      }, delay);
+
+      return () => clearTimeout(timeout);
+    } else {
+      setDelayedLoadingState(value);
+    }
+  }, [delay, value]);
+
+  return delayedLoadingState;
+};
 
 const PaMessagesPage: ComponentType = () => {
-  const [paMessages, setPaMessages] = useState<PaMessage[]>([]);
+  const [params, setParams] = useSearchParams();
+  const [stateFilter, setStateFilter] = useState<StateFilter>(
+    () => (params.get("state") as StateFilter) ?? "active",
+  );
 
   useEffect(() => {
-    const fetchAndSetPaMessages = async () => {
-      const allPaMessages = await fetchPaMessages();
-      setPaMessages(allPaMessages);
-    };
-    fetchAndSetPaMessages();
-  }, []);
+    setParams(() => {
+      const newParams = new URLSearchParams();
+
+      if (stateFilter) newParams.set("state", stateFilter);
+
+      return newParams;
+    });
+  }, [stateFilter]);
+
+  const { data, isLoading } = usePaMessages({ stateFilter });
+  const shouldShowLoadingState = useDelayedLoadingState(isLoading);
 
   return (
     <>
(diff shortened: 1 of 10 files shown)


assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -1,29 +1,84 @@
...
-    const fetchAndSetPaMessages = async () => {
-      const allPaMessages = await fetchPaMessages();
-      setPaMessages(allPaMessages);
-    };
-    fetchAndSetPaMessages();
-  }, []);
+    setParams(() => {
+      const newParams = new URLSearchParams();
+
+      if (state) newParams.set("state", state);
```
digitalcora, line 42: We're building the URL params twice in the exact same way here (L22~ and L40~). Should we de-duplicate so these will be easier to keep in sync?

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -1,29 +1,84 @@
...
+} from "react-bootstrap";
 import { PlusCircleFill } from "react-bootstrap-icons";
-import { fetchPaMessages } from "Utils/api";
 import { PaMessage } from "Models/pa_message";
-import { Link } from "react-router-dom";
+import { Link, useSearchParams } from "react-router-dom";
+import cx from "classnames";
+import useSWR from "swr";
+
+type State = "active" | "future" | "past";
```
digitalcora, line 16: `State` almost seems like too generic a term here considering it could refer to the overall "state" of the page/component. idk if I'd go so far as to change the name of the API query param, but maybe in the context of this frontend code it could be more clearly-named, like "time filter". (Though maybe once we have more filters and presumably those are defined together in some way, it will become more clear that this is a "state filter".)

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -78,7 +133,7 @@ const PaMessageTable: ComponentType<PaMessageTableProps> = ({
       </table>
       {paMessages.length == 0 && (
         <div className="pa-message-table__empty">
-          There are no active PA/ESS Messages.
+          There are no PA/ESS Messages matching the current filters.
```
digitalcora, line 136: Should we add a simple text message for the error/loading states of SWR as well? In the video it does look a bit odd that the table briefly disappears when loading new messages.

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -1,29 +1,84 @@
...
+import {
+  Container,
+  Row,
+  Col,
+  FormCheck,
+  ButtonGroup,
+  Button,
+} from "react-bootstrap";
 import { PlusCircleFill } from "react-bootstrap-icons";
-import { fetchPaMessages } from "Utils/api";
```
digitalcora, line 4: Is this function still used elsewhere or can we remove it?

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -42,7 +97,7 @@ const PaMessagesPage: ComponentType = () => {
               </Col> */}
             </Row>
             <Row>
-              <PaMessageTable paMessages={paMessages} />
+              {Array.isArray(data) && <PaMessageTable paMessages={data} />}
```
digitalcora, line 100: It seems like we never actually narrow `data` from what I assume is a generic JSON value type returned from the fetcher. How does this work with the prop type needing to be a `PaMessage[]`?

digitalcora (commented):
Liking how straightforward this is!

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: Yeah I started with `stateFilter`. I'll put it back that way. You're right, this is a bit too generic.

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: Consider it gone!

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: This "works" because things like `Response.prototype.json()` and `JSON.parse()` return `any`.

We could cast the value in the `usePaMessages` but that's effectively what happens we by the time the data is consumed in `<PaMessageTable />`

Getting type safe external data would require first parsing any external data to ensure it conforms the the types we're expecting. It would look something like this: https://github.com/mbta/screenplay/pull/433

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: These will diverge when we add filters on service type / route

- https://github.com/mbta/screenplay/pull/431/files#diff-63017d34d99e9b41261c38d09103b8c5eb957968e3ac83d6824c7cfc809ce03fR61-R69

- https://github.com/mbta/screenplay/pull/431/files#diff-63017d34d99e9b41261c38d09103b8c5eb957968e3ac83d6824c7cfc809ce03fR88-R92

assets/js/components/Dashboard/PaMessagesPage.tsx
digitalcora: Huh. I kinda expected those to be properly typed (in which case we'd have gotten an error due to accessing possibly undefined properties on the data, and would have to explicitly type-cast it). [It's not like JSON is that complicated.](https://github.com/Microsoft/TypeScript/issues/15225#issuecomment-294718709) 🫠

I think I'd prefer an explicit type-cast somewhere close to where the data is fetched, but enforcing the type conformance would be even better.

sloanelybutsurely: [Done!](https://github.com/mbta/screenplay/compare/75744043a2d6993fa135508113d577b85d850798..9fa47b418fd34e5fb7a4eee0c70417279c2a2cf7)

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: I've asked Mary how she'd like these loading states to work [[slack](https://mbta.slack.com/archives/C039ARUM48H/p1723237320826709)]

sloanelybutsurely: Okay Mary asked for a spinner in the table when loading new data. You could still see a pop of that loading state whenever you clicked on a new filter though. So I've also added a 250ms delay to showing the loading state at all. This is something I've done in the past and in my opinion makes the application feel snappier.

<details>
<summary>Fast network</summary>

https://github.com/user-attachments/assets/b6aa6baf-7238-4de7-90bb-4abfb4fc2e95

</details>

<details>
<summary>Slow network</summary>

https://github.com/user-attachments/assets/d74f23b2-da42-4026-89fd-00c12322dac1

</details>

digitalcora (approved):
I had one last question about maybe using the "slow loading" options in SWR, but whether or not we use that, looks good!

assets/js/components/Dashboard/PaMessagesPage.tsx
sloanelybutsurely: I do kind of prefer my callback-less approach but it would totally work either way. Not a hard thing to change later on if we feel like we need to for some reason.

PR #435, opened 2024-08-12
feat: start_time changeset validator

cmaddox5:
Added a changeset validator for `start_time`. The client should catch this, but in case it doesn't we will prevent it from being inserted.


sloanelybutsurely (approved):
lgtm. unrelated to this change specifically but the mixing of `state_date` and `start_time` is kind of confusing 😕

cmaddox5:
> lgtm. unrelated to this change specifically but the mixing of `state_date` and `start_time` is kind of confusing 😕

I've been thinking about that too. I can make a migration for changing the column name.

PR #437, opened 2024-08-12
feat: Import signs from alert informed_entities

cmaddox5:
This PR adds the logic needed to select signs informed by an alert. It filters down by stop > route > direction (if provided in `informed_entities`).

---
  - https://app.asana.com/0/0/1208009646310782

### DIFF
--- a/assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
+++ b/assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
@@ -53,11 +53,10 @@ const NewPaMessage = () => {
         : places;
 
       if (entity.route) {
+        const entityRoute = entity.route;
         let signsToAdd = informedPlaces
           .flatMap((place) => place.screens)
-          .filter((screen) =>
-            getRouteIdsForSign(screen).includes(entity.route ?? ""),
-          );
+          .filter((screen) => getRouteIdsForSign(screen).includes(entityRoute));
 
         const directionId = entity.direction_id;
         if (directionId !== null) {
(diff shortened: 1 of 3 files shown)


assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
```diff
@@ -45,7 +46,36 @@ const NewPaMessage = () => {
     setVisualText(alertMessage);
   };
 
-  const onImportLocations = () => {};
+  const onImportLocations = (informedEntities: InformedEntity[]) => {
+    const importedSigns = informedEntities.reduce((acc, entity) => {
```
sloanelybutsurely, line 50: **suggestion:** I'm pretty sure this is a [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
```diff
@@ -45,7 +46,36 @@ const NewPaMessage = () => {
...
+      const informedPlaces = entity.stop
+        ? places.filter((place) => place.id === entity.stop)
+        : places;
+
+      if (entity.route) {
+        let signsToAdd = informedPlaces
+          .flatMap((place) => place.screens)
+          .filter((screen) =>
+            getRouteIdsForSign(screen).includes(entity.route ?? ""),
```
sloanelybutsurely, line 59: **suggestion:** If you create a variable of the narrowed `entity.route` value you won't need that nullish coalescing operator.

```suggestion
      if (entity.route) {
        const entityRoute = entity.route
        let signsToAdd = informedPlaces
          .flatMap((place) => place.screens)
          .filter((screen) =>
            getRouteIdsForSign(screen).includes(entityRoute),
```

sloanelybutsurely (approved):

assets/js/components/Dashboard/NewPaMessage/NewPaMessage.tsx
cmaddox5: Ahh yes. Thanks!

PR #440, opened 2024-08-12
fix: Zone filter sorting

cmaddox5:
This PR addresses a big ole 🤦 . Accidentally changed a variable [here](https://github.com/mbta/screenplay/blob/a15c3ff18fc48fd705b07b53e02c74a727b36d63/assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx#L225) that I shouldn't have. Was causing an error when a GL station was not selected.


sloanelybutsurely (approved):
:shipit:

PR #442, opened 2024-08-13
chore: Read PA/ESS custom labels at runtime

cmaddox5:
This PR moves `paess_labels.json` so it can be read at runtime. This will make it easier to change custom labels on the fly. This required changing the way we hide lab signs in the script. I also changed the keys in the label file to be the sign IDs instead of `{station_code}-{zone}`.

Here is a good starting file for [paess_labels.json](https://github.com/user-attachments/files/16601175/paess_labels.json).


lib/screenplay/config/local_fetch.ex
```diff
@@ -116,4 +116,30 @@ defmodule Screenplay.Config.LocalFetch do
...
+              place["screens"],
+              [],
+              fn screen, acc ->
+                [Map.put(screen, "label", Map.get(labels, screen["id"])) | acc]
+              end
+            )
+
+          [Map.put(place, "screens", new_screens) | acc]
+        end
+      )
```
sloanelybutsurely, line 138: **question:** Do we need `reduce` here? These look like straight-forward `Enum.map`s to me (with the exception that they also reverse the lists processed but I imagine that's just a side effect of using `reduce` in the first place).

```elixir
Enum.map(config, fn place ->
  new_screens =
    Enum.map(place["screens"], fn screen ->
      Map.put(screen, "label", Map.get(labels, screen["id"]))
    end)

  Map.put(place, "screens", new_screens)
end)
```

Additionally, this might be simplified further with [`update_in/2`](https://hexdocs.pm/elixir/Kernel.html#update_in/2) (or [`update_in/3`](https://hexdocs.pm/elixir/Kernel.html#put_in/3))

lib/screenplay/config/s3_fetch.ex
```diff
@@ -121,4 +124,28 @@ defmodule Screenplay.Config.S3Fetch do
...
+              end
+            )
+
+          [Map.put(place, "screens", new_screens) | acc]
+        end
+      )
+    else
+      _ -> config
+    end
+  end
```
sloanelybutsurely, line 150: **suggestion:** The contents of this `with` clause are the same as above. Can we pull that into a shared function?

lib/screenplay/config/local_fetch.ex
cmaddox5: No idea why I've been so `reduce`-happy lately. Will work on changing it.

lib/screenplay/config/s3_fetch.ex
cmaddox5: Yeah good idea. I feel like the best place for this is in `Screenplay.Config.Fetch` but wasn't sure if implementation and callbacks should live in the same module. What do you think?

sloanelybutsurely: I think that's totally fine!

lib/screenplay/config/fetch.ex
```diff
@@ -12,4 +12,12 @@ defmodule Screenplay.Config.Fetch do
   @callback put_config(list(PlaceAndScreens.t())) :: :ok | :error
   @callback commit() :: :ok
   @callback revert(version_id()) :: any()
+
+  def add_labels_to_config(config, labels) do
+    update_in(
+      config,
+      [Access.all(), "screens", Access.all()],
+      &Map.put(&1, "label", Map.get(labels, &1["id"]))
+    )
```
sloanelybutsurely, line 21: it's beautiful 😍

sloanelybutsurely (approved):

PR #443, opened 2024-08-14
Cm/remove label from showtime

cmaddox5:
This new `label` field doesn't apply to Showtime so need to filter those out when adding.


sloanelybutsurely (approved):
Thanks! I've tested my code with this change and things are working again.

PR #446, opened 2024-08-16
chore: Add dependabot groups

cmaddox5:
I want to try out Dependabot's `groups` option so we can reduce the number of PRs it opens. Feel free to call out groups that should be added or groups I added that don't look right.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,11 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-minor"]
+    groups:
+      mix-patches:
+        applies-to: version-updates
+        update-types:
+          - "patch"
   - package-ecosystem: npm
     directory: "/assets"
     schedule:
@@ -19,10 +24,8 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-minor"]
     groups:
-      patches:
+      npm-patches:
         applies-to: version-updates
-        patterns:
-          - "*"
         update-types:
           - "patch"
       eslint:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,6 @@ updates:
         update-types: ["version-update:semver-minor"]
     groups:
       mix-patches:
-        applies-to: version-updates
         update-types:
           - "patch"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -18,3 +18,35 @@ updates:
...
+          - "@babel*"
+     webpack:
+        applies-to: version-updates
+        patterns:
+          - "webpack*"
+          - "*webpack-plugin"
+     moment:
+        applies-to: version-updates
+        patterns:
+          - "moment*"
```
sloanelybutsurely, line 52: **suggestion:**

```yaml
sentry:
  applies-to: version-updates
  patterns:
    - "@sentry/*"
```

sloanelybutsurely (approved):

digitalcora:
Could we make a group that just has all patch-level updates (of all libraries) in it, using `update-types`? I feel like these both make up a majority of the "commit noise" and are a kind of update we can be reasonably confident about making "in bulk", assuming libraries are following semver conventions.

.github/dependabot.yml
```diff
@@ -18,3 +18,45 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-minor"]
+    groups:
+      patches:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - "patch"
```
digitalcora, line 27: We could do this group for `mix` packages too.

.github/dependabot.yml
```diff
@@ -18,3 +18,45 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-minor"]
+    groups:
+      patches:
+        applies-to: version-updates
+        patterns:
+          - "*"
```
digitalcora, line 25: I don't think this is required, based on the examples in the docs — omitting `patterns` should be the same as having `*` as the only pattern.

digitalcora (approved):
Looking forward to seeing the results!

.github/dependabot.yml
cmaddox5: O gotcha. Was being a bit too cautious.

PR #451, opened 2024-08-16
chore: Upgrade React

cmaddox5:
Upgraded React and all packages that depend on specific React versions. This essentially combines two Dependabot PRs into one. The `testing-tools` PR was failing because of a dependency on React 18. This required small changes to some client tests, but how the tests function should not have changed.

This branch is currently running on dev-green and not seeing any errors or unexpected changes.


digitalcora:
> The `testing-tools` PR was failing because of a dependency on React 18.

Does this suggest a reconsideration of the group structure? Ideally the grouping would try to avoid issues where upgrading one dependency fails because another one needs to be updated in tandem.

cmaddox5:
> > The `testing-tools` PR was failing because of a dependency on React 18.
> 
> Does this suggest a reconsideration of the group structure? Ideally the grouping would try to avoid issues where upgrading one dependency fails because another one needs to be updated in tandem.

I did think about that but this being a major version upgrade is what made me hesitate. I feel like the grouping is ok for minor and patches, but maybe that's being a bit too optimistic. Happy to move the groups around if so.

digitalcora (approved):
Thanks for doing the legwork on these major updates!

PR #454, opened 2024-08-19
chore: Fix sign columns

cmaddox5:
In the original implementation, I designed the page to organize the signs by zone in the same way SignsUI does. This was a misunderstanding and the signs should actually be organized by direction ID. I changed up the logic so the left column is direction 0, middle is 0 and 1, right is 1. 

---
  - https://app.asana.com/0/0/1208030889228580

### DIFF
--- a/assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
+++ b/assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
@@ -146,15 +146,11 @@ const SelectZonesPage = ({
       ),
     ).every(isSelected);
 
-  const selectDirectionGroup = (directionIds: (0 | 1)[]) => {
+  const selectDirectionGroup = (directionId: 0 | 1) => {
     const signsWithoutSelectedRoute = fp.without(signIDs(allScreens), value);
 
-    const newScreens = allScreens.filter(
-      (s) =>
-        fp.intersection(
-          s.routes?.map((r) => r.direction_id),
-          directionIds,
-        ).length > 0,
+    const newScreens = allScreens.filter((s) =>
+      s.routes?.map((r) => r.direction_id).includes(directionId),
     );
 
     onChange(fp.uniq([...signsWithoutSelectedRoute, ...signIDs(newScreens)]));
@@ -274,7 +270,9 @@ const SelectZonesPage = ({
                   className={cx("button-secondary-outline", {
                     "button-active": isAllSelected,
                   })}
-                  onClick={() => selectDirectionGroup([0, 1])}
+                  onClick={() =>
+                    onChange(fp.uniq([...value, ...signIDs(allScreens)]))
+                  }
                 >
                   {massSelectButtonLabels.left}
                 </Button>
@@ -282,7 +280,7 @@ const SelectZonesPage = ({
                   className={cx("button-secondary-outline", {
                     "button-active": isLeftSelected,
                   })}
-                  onClick={() => selectDirectionGroup([0])}
+                  onClick={() => selectDirectionGroup(0)}
                 >
                   {massSelectButtonLabels.middle}
                 </Button>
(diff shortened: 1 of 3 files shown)


assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -140,35 +139,31 @@ const SelectZonesPage = ({
...
     const signsWithoutSelectedRoute = fp.without(signIDs(allScreens), value);
 
-    const newScreens = groups.flatMap((group) =>
-      group.flatMap((zone) => signIDs(screensByZone[zone] ?? [])),
+    const newScreens = allScreens.filter(
+      (s) =>
+        fp.intersection(
+          s.routes?.map((r) => r.direction_id),
+          directionIds,
+        ).length > 0,
```
digitalcora, line 157: I think this might be easier to understand, and possibly more efficient, as:

```js
allScreens.filter((screen) =>
  screen.routes?.some((route) => directionIds.includes(route.direction_id))
)
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -279,25 +274,23 @@ const SelectZonesPage = ({
                   className={cx("button-secondary-outline", {
                     "button-active": isAllSelected,
                   })}
-                  onClick={() =>
-                    selectGroups([LEFT_ZONES, MIDDLE_ZONES, RIGHT_ZONES])
-                  }
+                  onClick={() => selectDirectionGroup([0, 1])}
```
digitalcora, line 277: Would this be equivalent to "select everything"? (i.o.w. is there even a need for `selectDirectionGroup` to support more than one direction?)

sloanelybutsurely (approved):

assets/js/util.ts
```diff
@@ -185,12 +181,13 @@ export const placesWithSelectedAlert = (
...
+export const signsByDirection = (signs: Screen[]) => {
   const groupedSigns = fp.groupBy((sign) => {
-    if (fp.includes(sign.zone, LEFT_ZONES)) {
-      return "left";
-    } else if (fp.includes(sign.zone, MIDDLE_ZONES)) {
+    const directionIds = sign.routes?.map((r) => r.direction_id);
+    if (fp.intersection(directionIds, [0, 1]).length === 2) {
       return "middle";
+    } else if (directionIds?.includes(0)) {
+      return "left";
```
digitalcora, line 190: I'm a little confused here. Under this logic if a sign has direction `0` but not direction `1`, it goes in the `left` group. But elsewhere, the `left` selection button appears to select all signs with both directions.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: Good point. This case doesn't need this function.

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
cmaddox5: This was simplified as a result of the change [here](https://github.com/mbta/screenplay/pull/454#discussion_r1722245143).

assets/js/util.ts
cmaddox5: Your description sounds right and is by design. The grouping buttons select signs with the labeled direction so `middle` signs by default will always be selected when either button is clicked.

digitalcora: Oh, duh, the grouping of the columns is [0 / both / 1], but the arrangement of the buttons is [both / 0 / 1]. Not sure why I was confused there 😅

digitalcora (approved):
↔️ ☑️

PR #455, opened 2024-08-19
Clean up message management functions

PaulJKim:
**Asana task**: [Clean up Message Management Functions](https://app.asana.com/0/1185117109217413/1207983141001734/f)

Description
This PR removes two columns that are now out of scope for the first release

### DIFF
--- a/assets/css/dashboard/pa-messages-page.scss
+++ b/assets/css/dashboard/pa-messages-page.scss
@@ -99,17 +99,6 @@
     padding-right: 0px;
   }
 
-  &__actions {
-    padding-top: 0px !important;
-    min-width: 102px;
-
-    a {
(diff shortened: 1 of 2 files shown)


sloanelybutsurely (approved):
i would prefer we delete this code as some of these features are still being iterated on by design but i won't block on that

digitalcora:
I would also advocate for deleting rather than commenting-out. If we wanted to bring it back later, we could do so by reverting this commit — but even if we don't remember that, this change is small enough that it could be recreated from scratch pretty quickly. (If the surrounding code and/or the design has changed, it might be _easier_ to do from scratch than to start from the old version.)

PaulJKim:
@sloanelybutsurely @digitalcora works for me!

PR #456, opened 2024-08-20
fix: disappearing places on SelectZonesPage

cmaddox5:
Hoping to squash this bug once and for all. Tl;dr, unselecting all signs for a place would cause the place to disappear when changing the route filter. I think the problem was some of the logic in the original `useMemo` needs to change much less frequently than what is returned. I split out that piece into its own and places are no longer disappearing.

### DIFF
--- a/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, useMemo, useRef, useState } from "react";
+import React, { Fragment, useMemo, useState } from "react";
 import { Page } from "../types";
 import { Button } from "react-bootstrap";
 import { Place } from "Models/place";
@@ -42,9 +42,11 @@ const SelectZonesPage = ({
   places,
 }: Props) => {
   const routeToRouteIDMap = useRouteToRouteIDsMap();
+  const getInitialPlacesWithSelectedSigns = () =>
+    places.filter((p) => p.screens.some((s) => value.includes(s.id)));
 
-  const { current: initialPlacesWithSelectedSigns } = useRef(
-    places.filter((p) => p.screens.some((s) => value.includes(s.id))),
+  const [initialPlacesWithSelectedSigns] = useState<Place[]>(
+    getInitialPlacesWithSelectedSigns,
   );
 
   const getInitialRoutes = () => {
--- a/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
@@ -43,12 +43,7 @@ const SelectZonesPage = ({
 }: Props) => {
   const routeToRouteIDMap = useRouteToRouteIDsMap();
   const getInitialPlacesWithSelectedSigns = () =>
-    places.filter((p) =>
-      places
-        .filter((p) => p.screens.some((s) => value.includes(s.id)))
-        .map((p) => p.id)
-        .includes(p.id),
-    );
+    places.filter((p) => p.screens.some((s) => value.includes(s.id)));
 
   const getInitialRoutes = () => {
     return fp.flow(
--- a/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
@@ -45,12 +45,14 @@ const SelectZonesPage = ({
   const getInitialPlacesWithSelectedSigns = () =>
     places.filter((p) => p.screens.some((s) => value.includes(s.id)));
 
+  // eslint-disable-next-line react/hook-use-state
+  const [initialPlacesWithSelectedSigns] = useState<Place[]>(
+    getInitialPlacesWithSelectedSigns,
+  );
+
   const getInitialRoutes = () => {
     return fp.flow(
-      fp.flatMap((place: Place) =>
-        place.screens.filter((s) => value.includes(s.id)),
-      ),
-      fp.flatMap(getRouteIdsForSign),
+      fp.flatMap((place: Place) => place.screens.flatMap(getRouteIdsForSign)),
       fp.uniq,
       fp.groupBy((routeID: string) => {
         if (routeID.startsWith("Green")) {
@@ -67,16 +69,10 @@ const SelectZonesPage = ({
 
         return routeID;
       }),
-    )(places);
+    )(initialPlacesWithSelectedSigns);
   };
 
-  // eslint-disable-next-line react/hook-use-state
-  const [initialPlacesWithSelectedSigns] = useState<Place[]>(
-    getInitialPlacesWithSelectedSigns,
-  );
-
-  // eslint-disable-next-line react/hook-use-state
-  const [initialRoutes] = useState(getInitialRoutes);
+  const initialRoutes = getInitialRoutes();
 
   const [selectedRouteFilter, setSelectedRouteFilter] = useState(
     Object.keys(initialRoutes)[0],
--- a/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, useMemo, useState } from "react";
+import React, { Fragment, useMemo, useRef, useState } from "react";
 import { Page } from "../types";
 import { Button } from "react-bootstrap";
 import { Place } from "Models/place";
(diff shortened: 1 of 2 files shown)


sloanelybutsurely (commented):
This looks like an issue caused by a ["rules of hooks"](https://react.dev/reference/rules/rules-of-hooks) violation. There's an eslint plugin that will call out issues like this. Here's a PR to add that https://github.com/mbta/screenplay/pull/459.

And here's the output of that rule on this branch:

```
screenplay/assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
  126:6  warning  React Hook useMemo has a missing dependency: 'value'. Either include it or remove the dependency array                           react-hooks/exhaustive-deps
  132:6  warning  React Hook useMemo has a missing dependency: 'initialPlacesWithSelectedSigns'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

✖ 2 problems (0 errors, 2 warnings)
```

assets/js/components/Dashboard/NewPaMessage/StationsAndZones/SelectZonesPage.tsx
```diff
@@ -117,15 +117,17 @@ const SelectZonesPage = ({
     };
   }, [selectedRouteFilter]);
 
-  const filteredPlaces = useMemo(() => {
-    const placesWithSelectedSigns = places.filter((p) =>
+  const initialPlacesWithSelectedSigns = useMemo(() => {
```
panentheos, line 120: Just a note, `useMemo` is a good tool for fixing specific kinds of performance issues (measurably slow re-renders after frequent, unrelated state changes, e.g. keyboard events). However, it adds a bunch of extra complexity and statefulness in the form of the cache, so it's best avoided unless we know we need it.

cmaddox5: Thanks for the tip. I feel like I've been a bit too eager to use `useMemo` just because I don't need the setter from `useState`. Changed this and another variable into `useState` without a setter instead.

digitalcora: If you don't ever need to set the state and don't want the setter cluttering things up, `const [value] = useState(...)` works. Though it really feels like it shouldn't if you're used to Elixir's destructuring.

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
```diff
@@ -42,8 +42,15 @@ const SelectZonesPage = ({
   places,
 }: Props) => {
   const routeToRouteIDMap = useRouteToRouteIDsMap();
+  const getInitialPlacesWithSelectedSigns = () =>
+    places.filter((p) =>
+      places
+        .filter((p) => p.screens.some((s) => value.includes(s.id)))
+        .map((p) => p.id)
+        .includes(p.id),
+    );
```
panentheos, line 51: Could this just be:
```suggestion
    places.filter((p) => p.screens.some((s) => value.includes(s.id)))
```

cmaddox5: O yikes, good catch.

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
```diff
@@ -66,10 +68,18 @@ const SelectZonesPage = ({
         return routeID;
       }),
     )(places);
-  }, [places, routeToRouteIDMap, value]);
+  };
+
+  // eslint-disable-next-line react/hook-use-state
+  const [initialPlacesWithSelectedSigns] = useState<Place[]>(
+    getInitialPlacesWithSelectedSigns,
+  );
```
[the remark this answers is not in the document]
digitalcora: oops, this was [my suggestion](https://github.com/mbta/screenplay/pull/456#discussion_r1731347011) in the first place (omitting the setter assignment entirely to reduce clutter), and I didn't realize a lint rule would complain about it. [Relevant discussion here](https://github.com/jsx-eslint/eslint-plugin-react/issues/3268) — in particular it seems like this rule doesn't have a special case for underscores in the name, nor is it configurable in a way that would allow them. The name just has to be exactly the same but with `set` prepended.

On a second reading I'm not sure why this can't be simplified to:

```ts
const initialPlacesWithSelectedSigns = getInitialPlacesWithSelectedSigns()
```

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
```diff
@@ -66,10 +68,18 @@ const SelectZonesPage = ({
...
+
+  // eslint-disable-next-line react/hook-use-state
+  const [initialRoutes] = useState(getInitialRoutes);
```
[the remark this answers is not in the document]
cmaddox5: Was able to rework that function so it looked at the new unchanging state value instead of the ever-changing `value` prop. Keeps the filter list from having the same problem as the place rows.

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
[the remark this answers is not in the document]
cmaddox5: And to Cora's point, it seems like I need to either disable the rule or convert it to a `useMemo` (which seems like it would have the same problem as before due to reevaluation on rerenders).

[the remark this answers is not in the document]
digitalcora: Does this work? It's a slightly weird way to use a ref, but like state, the initial value is only ever assigned once.

```ts
const { current: initialPlacesWithSelectedSigns } = useRef(
  places.filter((p) => p.screens.some((s) => value.includes(s.id))),
);
```

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
```diff
@@ -42,13 +42,17 @@ const SelectZonesPage = ({
...
 }: Props) => {
   const routeToRouteIDMap = useRouteToRouteIDsMap();
+  const getInitialPlacesWithSelectedSigns = () =>
+    places.filter((p) => p.screens.some((s) => value.includes(s.id)));
 
-  const selectedRoutes = useMemo(() => {
+  // eslint-disable-next-line react/hook-use-state
+  const [initialPlacesWithSelectedSigns] = useState<Place[]>(
+    getInitialPlacesWithSelectedSigns,
+  );
```
sloanelybutsurely, line 51: I'm sorry for the churn on this. I'm asking for `getInitialPlacesWithSelectedSigns` to be inlined into the call to `setState` (or `useRef` if you end up going that route) because it:

- is not used elsewhere
- is misleadingly named because it actually gets places with selected signs _at any point_ but you are currently only using to get the initial places with selected signs

```typescript
  // eslint-disable-next-line react/hook-use-state
  const [initialPlacesWithSelectedSigns] = useState<Place[]>(() =>
    places.filter((p) => p.screens.some((s) => value.includes(s.id))),
  );
```

digitalcora: Agree with this — that's exactly why I inlined the function in my proposed change.

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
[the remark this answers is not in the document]
sloanelybutsurely: Can I suggest that we instead remove `react/hook-use-state` from our eslint config?

https://github.com/mbta/screenplay/blob/6441ed4b513cb7d73e333437ab03fa0c148bfcce/assets/.eslintrc.js#L44

It isn't part of the recommended settings for react hooks which we do have enabled

https://github.com/mbta/screenplay/blob/6441ed4b513cb7d73e333437ab03fa0c148bfcce/assets/.eslintrc.js#L10

and I think through this discussion we've demonstrated a thorough understanding of what we are doing and why.

---

As an aside, I'm being picky about this because "ignore" comments for linting rules feel like a code smell to me. I feel like we should be all in any linting rule we adopt. Anything that requires more nuance than the linter can provide should be handled in code review. And I really like @digitalcora's suggestion to just destructure out the value and only suggested the `_` prefix because I thought eslint was complaining about the unused variable.

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
cmaddox5: O I see. I misunderstood.

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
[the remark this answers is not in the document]
cmaddox5: The `useRef` is interesting. I think I'll go with that. Won't leave us confused why we threw away the setter in `useState`.

And I'm all for removing this rule. I foresee it doing more harm than good.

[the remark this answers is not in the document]
digitalcora: If we're removing the rule I'd go with the `useState`. If anything the `useRef` approach is more confusing, the only advantage being that it keeps to the letter of the rule if we wanted to retain it. (I'd certainly like it if this rule could check setter names only _if_ you destructure them, and allow a setter to be entirely absent, but it doesn't seem like there's a way to set that up.)

sloanelybutsurely (approved):

PR #457, opened 2024-08-20
dev: update `pull_configs` to include new file

digitalcora:
This file was moved from static to S3 configuration in 478f1b9.


cmaddox5 (approved):
Woops, forgot about this. Sorry about that.

PR #458, opened 2024-08-20
chore: generalize PaMessageForm

sloanelybutsurely:
Moves form component for PA messages from `NewPaMessage` into `PaMessageForm` so that it can be used to make an `EditPaMessage` component later on.

Just moving this separately so the diffs are easier to read later on.


assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
sloanelybutsurely, line 1: This is just the old contents of `NewPaMessage.tsx`

PR #459, opened 2024-08-20
chore: install eslint-plugin-react-hooks, fix issues

sloanelybutsurely:
I noticed a few "rules of hooks" violations and realized we weren't using the very helpful `eslint-plugin-react-hooks` module to catch these for us.

This installs that package and corrects any outstanding violations of the rules of hooks

https://react.dev/reference/rules/rules-of-hooks
https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks

### DIFF
--- a/assets/js/hooks/useUpdateAnimation.tsx
+++ b/assets/js/hooks/useUpdateAnimation.tsx
@@ -21,7 +21,11 @@ export const useUpdateAnimation = (
       timer.current = undefined;
       setShowAnimation(false);
     }, 2000);
-  }, deps);
+    // Disabling this because this code predates our usage of the `react-hooks`
+    // eslint plugin. There's likely a better way to achieve this effect but
+    // we are disabling this check for now.
+    // - sloane 2024-08-20
+  }, deps); // eslint-disable-line react-hooks/exhaustive-deps
 
   // Cancel timer if component unmounts.
   useEffect(() => {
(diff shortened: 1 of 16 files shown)


assets/js/hooks/useUpdateAnimation.tsx
```diff
@@ -21,7 +21,7 @@ export const useUpdateAnimation = (
       timer.current = undefined;
       setShowAnimation(false);
     }, 2000);
-  }, deps);
+  }, deps); // eslint-disable-line react-hooks/exhaustive-deps
```
digitalcora, line 24: Can we do something like `[prevValue, showAnimationOnMount].concat(deps)` here? Or else add a comment explaining why we have to ignore the warning / why it's safe to do this.

digitalcora (approved):
Nice, looks like this caught quite a haul of potential issues! We also aren't using this on `screens` yet so I'll go see if it's feasible to add.

assets/js/hooks/useUpdateAnimation.tsx
sloanelybutsurely: So there will be a warning about the dependencies array not being statically analyzable either way. I feel like there's probably a better way to achieve this though. I'll add a comment for now.

digitalcora: Hm, ok. I did just notice a [remark](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks#advanced-configuration) in this package's docs that _"we recommend most custom Hooks to not use the dependencies argument, and instead provide a higher-level API that is more focused around a specific use case."_ So there probably is some better way we could be doing this, but that doesn't have to be investigated here.

PR #469, opened 2024-08-28
chore: Improve Screenplay migrations on startup

cmaddox5:
Tying a couple of different things in hopes of eliminating the startup errors we see occasionally. 

- Replace our migrator with the out-of-the-box `Ecto.Migrator` in the supervision tree
- Increase `pool_size` to 15
    - I noticed on deploy, RDS shows 12 DB connections but the default for `pool_size` is 10. 

I'm not 100% confident this is the actual solution to our problem, but I felt it was a good start. During testing yesterday, I regularly saw the issue and see it less with these changes. Please do point out any other ideas that I missed. 

Note, this is actually a very similar solution that MyCharlie used for the same issue last year: https://github.com/mbta/my_charlie/pull/30.

---
  - https://app.asana.com/0/0/1207647895639263


digitalcora (approved):
Being able to delete our custom migrator is certainly an improvement in any case.

PR #479, opened 2024-09-16
chore: Remove all Triptych code.

cmaddox5:
The Triptych app is dead. Removing all code.


digitalcora (approved):
Huh. I would have expected Screenplay to use the `shows_alerts?` check that was extracted into `screens-config-lib`, but I guess it didn't? Why was that moved, then? 🤔

cmaddox5:
I was surprised too. That's a good question. My guess is that someone anticipated Screenplay would use it but never did.

PR #481, opened 2024-09-16
fix: SelectZonesPage initial filter value

cmaddox5:
When rendering the `SelectZonesPage`, `selectedRouteFilter` was being assigned to the first route ID in `initialRoutes` before it was sorted. Initializing the labels earlier and using the same variable to initialize and render helps solve this.

I also fixed an issue with all routes showing in the filter list instead of just the routes for the selected screens.

---
  - https://app.asana.com/0/0/1208169473836920

### DIFF
--- a/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
@@ -49,9 +49,15 @@ const SelectZonesPage = ({
     getInitialPlacesWithSelectedSigns,
   );
 
+  const isSelected = (id: string) => value.includes(id);
+
   const getInitialRoutes = () => {
     return fp.flow(
-      fp.flatMap((place: Place) => place.screens.flatMap(getRouteIdsForSign)),
+      fp.flatMap((place: Place) =>
+        place.screens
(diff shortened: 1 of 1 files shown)


assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
```diff
@@ -211,7 +216,7 @@ const SelectZonesPage = ({
         <div className="filters-container">
           <div>Service type</div>
           <div className="filters">
-            {sortRoutes(Object.keys(initialRoutes)).map((routeID) => {
+            {routeFilterLabels.map((routeID) => {
```
digitalcora, line 219: We're calling this "route labels" here but the individual items a "route ID" — the latter term seems to also be used in other places so maybe we should use it here instead of "labels".

digitalcora (approved):
🔃 ✅

assets/js/components/Dashboard/PaMessageForm/SelectStationsAndZones/SelectZonesPage.tsx
cmaddox5: Good call. I also feel like `initialRoutes` didn't really mean much so I tried improving that too.

PR #497, opened 2024-09-26
Cm/automate places and screens

cmaddox5:
**Asana task**: [title](url)

Description


test/screenplay/permanent_config_test.exs
```diff
@@ -19,6 +20,8 @@ defmodule Screenplay.PermanentConfigTest do
   end
 
   setup_all do
+    _ = start_supervised(Screenplay.Places.Cache)
```
digitalcora, line 23: There is a `start_supervised!` if we don't care about the result (or rather, if we want to crash on failure instead of returning an `:error` that we then ignore).

test/screenplay/permanent_config_test.exs
```diff
@@ -373,45 +376,56 @@ defmodule Screenplay.PermanentConfigTest do
...
+      |> Enum.map(&{&1.id, &1})
+      |> Cache.put_all()
 
-        File.write(get_fixture_path("places_and_screens.json"), Jason.encode!(fresh_config))
+      on_exit(fn ->
+        Cache.delete_all()
       end)
     end
 
     test "publishes pending screens and adds screen to places_and_screens" do
```
digitalcora, line 396: The "adds screen to places_and_screens" and "does not add to places_and_screens" parts of these tests appear to no longer be tested. I think that's fine as that's no longer part of this module's responsibility, but it does mean the second test might be redundant now, and the descriptions should be updated.

test/screenplay/places/builder_test.exs
```diff
@@ -0,0 +1,307 @@
...
+  use ExUnit.Case
+
+  import Mox
+
+  alias Screenplay.Places.Cache, as: PlacesCache
+  alias Screenplay.Places.{Builder, Place}
+  alias Screenplay.ScreensConfig.Cache, as: ScreensConfigCache
+
+  setup_all do
+    _ = start_supervised(Screenplay.Places.Cache)
```
digitalcora, line 11: Could also use `start_supervised!` here.

test/screenplay/places/builder_test.exs
```diff
@@ -0,0 +1,307 @@
...
+                     location: "",
+                     type: :bus_eink_v2
+                   }
+                 ],
+                 description: nil
+               }
+             ] = PlacesCache.all(nil, return: :value)
+    end
+
+    test "splits mutil-place screens", %{state: state} do
```
digitalcora, line 105: typo: `mutil`

test/screenplay/places/builder_test.exs
```diff
@@ -0,0 +1,307 @@
...
+        ScreensConfigCache.delete_all()
+      end)
+
+      %{state: %{stops_mod: Screenplay.Stops.Mock, routes_mod: Screenplay.Routes.Mock}}
+    end
+
+    test "adds all parent stations as places", %{state: state} do
+      expect(Screenplay.Stops.Mock, :fetch_parent_stops, 2, fn _stop_ids ->
+        []
+      end)
```
digitalcora, line 30: nitpick: Some of these empty `expect`s seem like they should fit on one line, which could improve readability.

test/screenplay/places/builder_test.exs
```diff
@@ -0,0 +1,307 @@
...
+      assert {:noreply, _} = Builder.handle_info(:build, state)
+
+      assert [
+               %Place{
+                 description: nil,
+                 id: "place-ogmnl",
+                 name: "Oak Grove",
+                 routes: ["Orange"],
+                 screens: [
+                   %Screenplay.Places.Place.PaEssScreen{
```
digitalcora, line 255: `Place` should be aliased already in this context so we don't have to use the full path. (ditto for L266, L271)

lib/screenplay/stops/stop.ex
```diff
@@ -0,0 +1,29 @@
...
+  @moduledoc """
+  Functions used to fetch stop data from the V3 API.
+  """
+
+  @behaviour Screenplay.Stops.Behaviour
+
+  alias Screenplay.V3Api
+
+  @impl true
+  def fetch_parent_stops(stop_ids) do
```
digitalcora, line 11: In this function name we call them "parent stops", but in the other one it's "parent stations". I'd vote to make these consistent. ("parent stations" might be more aligned with how they're called elsewhere)

test/screenplay_web/controllers/pa_messages_api_controller_test.exs
```diff
@@ -9,6 +12,7 @@ defmodule ScreenplayWeb.PaMessagesApiControllerTest do
     end
 
     _ = start_supervised({Screenplay.Alerts.Cache, get_json_fn: get_json_fn})
+    _ = start_supervised(Screenplay.Places.Cache)
```
digitalcora, line 15: ...and one more use case for `start_supervised!` here. (actually two?)

lib/screenplay/stops/behaviour.ex
```diff
@@ -0,0 +1,6 @@
+defmodule Screenplay.Stops.Behaviour do
```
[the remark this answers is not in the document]
digitalcora: I can't take the credit here, it was Sloane who surfaced this to my awareness (screens in peace 🥲)

assets/js/utils/api.ts
```diff
@@ -93,6 +93,11 @@ export const putPendingScreens = async (
   });
 };
 
+interface PublishScreensForPlaceResponse {
+  message: string;
+  new_config?: Place[];
+}
+
 export const publishScreensForPlace = async (
```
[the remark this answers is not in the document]
cmaddox5: Looks like it's able to infer the return type which is why eslint isn't complaing, but an explicit one would still be helpful. Can add to the other functions as well.

lib/screenplay/stops/stop.ex
```diff
@@ -0,0 +1,29 @@
...
+    case V3Api.get_json("/stops", %{
+           "filter[id]" => Enum.join(stop_ids, ",")
+         }) do
+      {:ok, %{"data" => data}} -> data
+      _ -> []
```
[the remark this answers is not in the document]
cmaddox5: Very good point. Much better to let it restart than wait 15 minutes for the next refresh.

lib/screenplay/stops/stop.ex
digitalcora: Actually, this is my bad, I thought this was specifically fetching parent stations, but now I see it's just fetching stops by ID. I think a better name for this would be `fetch_by_ids` or something.

digitalcora: To clarify the terminology as I understand it:

* A "stop" is one of the following depending on context:
  * a "stopping location" a.k.a. a stop with `location_type=0`
  * any entry in `stops.txt`, which includes "stopping locations", parent stations, entrances/exits, and pathway nodes
* A "parent station" is a stop with `location_type=1`. Stops that are _not_ parent stations may "have" or "belong to" a parent station. For entrances/exits and pathway nodes this is required, for "stopping locations" it is optional.
* A "child stop" is a stop that has a parent station, or depending on context, possibly just a "stopping location" that has a parent station.
  * Sometimes we use "platform" or "berth" to specifically refer to this latter case, but this is not a formal GTFS term.

lib/screenplay/permanent_config.ex
```diff
@@ -415,46 +409,22 @@ defmodule Screenplay.PermanentConfig do
...
-
-        [%{place_and_screens | screens: existing_screens ++ new_screens} | acc]
+        end)
+        |> Enum.map(fn {screen_id, %Screen{app_id: app_id, disabled: disabled}} ->
+          %ShowtimeScreen{id: screen_id, type: app_id, disabled: disabled}
+        end)
 
-      map, acc ->
-        [map | acc]
+      %{place | screens: screens ++ new_screens_at_place}
```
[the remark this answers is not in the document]
cmaddox5: Yeah I like that. Need to start remembering `update_in` when building stuff like this.

lib/screenplay_web/controllers/dashboard_api_controller.ex
```diff
@@ -17,28 +20,21 @@ defmodule ScreenplayWeb.DashboardApiController do
...
-        &Map.put(&1, "location", location)
-      )
+    Enum.map(config, fn %Place{screens: screens} = place ->
+      new_screens =
+        Enum.map(screens, fn %{id: id} = screen ->
+          location = Enum.find(locations, "", &match?(%{"id" => ^id}, &1))
+          %{screen | location: location}
+        end)
+
+      %{place | screens: new_screens}
```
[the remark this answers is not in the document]
cmaddox5: The client is able to handle `nil` and `""` in the same way so it isn't an issue. I did find a bug while looking at this though. The result of `Enum.find` is a map so I need to access the correct property before adding it to the data.

lib/screenplay/github_api/client.ex
```diff
@@ -0,0 +1,66 @@
...
+  Client used to fetch signs.json from the RTS GitHub repo
+  """
+
+  require Logger
+
+  @behaviour Screenplay.GithubApi.ClientBehaviour
+
+  @impl true
+  def get_file_contents_from_repo(name, file_path) do
+    url = "https://api.github.com/repos/mbta/#{name}/contents/#{file_path}"
```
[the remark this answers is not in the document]
cmaddox5: I agree keeping a copy in S3 would be a better solution. I'll create a task for us to talk through on Monday.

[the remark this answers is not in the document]
cmaddox5: https://app.asana.com/0/1185117109217422/1208468578245474/f

lib/screenplay/places/builder.ex
```diff
@@ -0,0 +1,518 @@
...
+    GlEink,
+    Header,
+    PreFare
+  }
+
+  use GenServer
+
+  @sl_route_ids ~w[741 742 743 746 749 751]
+
+  @polling_interval :timer.minutes(15)
```
digitalcora, line 33: suggestion: 5 minutes? Much more responsive but hopefully still easy on resources/API limits.

This also makes me think perhaps we should check the modification timestamps of the config files this pulls, keep a record (in the state of this module maybe) of when they were last changed, and only go through the whole process if there's been a change. That could be a future improvement though.

digitalcora (approved):
Some small comments, looking good! This turned out to be a pretty big change so thanks for taking it on!

lib/screenplay_web/controllers/dashboard_api_controller.ex
```diff
@@ -23,7 +23,9 @@ defmodule ScreenplayWeb.DashboardApiController do
     Enum.map(config, fn %Place{screens: screens} = place ->
       new_screens =
         Enum.map(screens, fn %{id: id} = screen ->
-          location = Enum.find(locations, "", &match?(%{"id" => ^id}, &1))
+          location =
+            Enum.find(locations, %{"location" => ""}, &match?(%{"id" => ^id}, &1))["location"]
```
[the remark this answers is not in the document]
cmaddox5: Sure, makes sense.

lib/screenplay/places/builder.ex
cmaddox5: Yeah 5 minutes would feel less frustrating. 

That's a good idea for an improvement. Could probably go a step further and implement API response caching in Screenplay so we can catch changes to stops and routes. Obviously those change much less frequently, but would hate for us to miss a change just cause our screen data has been changed.

PR #501, opened 2024-09-27
refactor: Simplify ScreensConfig cache

cmaddox5:
While working on the automate Screenplay data task, I found it very difficult to work with the cache in order to write tests. In my opinion, it's a bit over-engineered given that the architecture was written to reduce duplicated code across multiple data sources. 

This PR is definitely meant to present an option to my original problem. Push-back is totally welcome.


lib/screenplay/screens_config/cache.ex
```diff
@@ -1,125 +1,12 @@
...
-      @table
-      |> :ets.select(match_spec)
-      |> Map.new()
-    end
-  end
-
-  @doc """
-  Gets the entire config struct.
-
-  👉 WARNING: This function is expensive to run and returns a large amount of data.
```
digitalcora, line 112: Given the new solution just always does this, unless fetching a specific screen ID, did this comment turn out to be incorrect?

lib/screenplay/screens_config.ex
```diff
@@ -0,0 +1,45 @@
...
+    Cache.all(nil, return: :key)
+  end
+
+  def screens do
+    Cache.all(nil, return: {:key, :value})
+  end
+
+  @spec update_cache(list({String.t(), Screen.t()})) :: :ok
+  def update_cache(screens) do
+    Cache.put_all(screens)
```
digitalcora, line 43: If I'm understanding the docs correctly, this would _not_ remove screens whose IDs are no longer present in the input, which is a perhaps undesirable change in behavior from the original version.

digitalcora (commented):
Thanks for this; I'd agree the original version was over-engineered.

I am mulling over whether Nebulex is really the best tool for this job — in both this case and the draft PlacesAndScreens "cache", we're basically using it as a thin layer over ETS, and explicitly _don't_ want the stored values to ever expire, be limited to a certain size, be lazily fetched from the data source, or other behaviors you'd associate with a "cache". Nevertheless, I've searched around and it doesn't really seem like a _bad_ option short of using ETS directly. It just feels vaguely like we're using a screwdriver to hammer a nail.

lib/screenplay/screens_config.ex
```diff
@@ -0,0 +1,45 @@
...
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  @spec screen(String.t()) :: Screen.t() | nil
+  def screen(screen_id) do
+    Cache.get(screen_id)
+  end
+
+  @spec screen_ids() :: list(String.t())
+  def screen_ids do
```
digitalcora, line 33: This and `screen` don't appear to be used anywhere.

cmaddox5:
> I am mulling over whether Nebulex is really the best tool for this job

I did think about this when adding it. I honestly don't have any good reason for choosing Nebulex so if switching back to our own ETS implementation is desired, I'm happy to do that.

lib/screenplay/screens_config/fetch/fetcher.ex
```diff
@@ -0,0 +1,44 @@
+defmodule Screenplay.ScreensConfig.Fetch.Fetcher do
+  @moduledoc """
+  GenServer that fetches and caches the Screens app config
+  """
+  use GenServer
+
+  alias Screenplay.ScreensConfig, as: ScreensConfigUpdater
```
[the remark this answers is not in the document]
cmaddox5: This is a similar issue to the first comment above. Alias names are conflicting. Starting to think we should reconsider the name of the module.

lib/screenplay/screens_config/cache.ex
```diff
@@ -1,125 +1,12 @@
...
+    otp_app: :screenplay,
+    adapter: Nebulex.Adapters.Local
 
-  use Screenplay.Cache.Client, table: :screens_config
-
-  @type table_contents :: list(table_entry)
-
-  @type table_entry ::
-          {{:screen, screen_id :: String.t()}, ScreensConfig.Screen.t()}
-          | {:devops, ScreensConfig.Devops.t()}
```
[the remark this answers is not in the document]
cmaddox5: I removed it because I can't think of a reason Screenplay would need it anyway. It was likely brought over when we changed the caching in Screenplay to match Screens and didn't think to remove it.

lib/screenplay/screens_config.ex
cmaddox5: Ahh yes thank you. I should throw a `delete_all` in there first.

lib/screenplay/config/permanent_config.ex
```diff
@@ -7,7 +7,7 @@ defmodule Screenplay.Config.PermanentConfig do
   alias Screenplay.Config.{Fetch, PlaceAndScreens}
   alias Screenplay.PendingScreensConfig.Fetch, as: PendingScreensFetch
   alias Screenplay.RoutePatterns.RoutePattern
-  alias Screenplay.ScreensConfig.Cache, as: ScreensConfigCache
+  alias Screenplay.ScreensConfig, as: ScreensConfigCache
```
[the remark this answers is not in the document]
digitalcora: You're absolutely right. I spaced on the fact that `ScreensConfig` is exactly the module name of screens-config-lib. In that case I think this alias name is okay despite the potential confusion... maybe ScreensConfig...Store? Just `Screens`?

lib/screenplay/screens_config.ex
digitalcora: I don't think that would be a good way of handling this, since it would result in the store being empty for a brief interval — if another process accessed it during that time, it would see no screens in the config. This same pattern was behind some bugs we used to have (/maybe still have) in the V3 API where requests would very occasionally return `[]` instead of the correct list of resources. A better approach would be to determine the set of keys that are present in the existing data but not in the new data, and delete only those.

cmaddox5: Very good point. Can definitely make that happen.

lib/screenplay/config/permanent_config.ex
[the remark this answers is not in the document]
cmaddox5: I'm a fan of `Store`. I'll run with that.

lib/screenplay/screens_config/cache.ex
cmaddox5: That was my conclusion. May have been an overly cautious warning. Each call to the function is nearly instant so there is no need for concern when retrieving the full cache.

digitalcora: My intuition is also that this shouldn't have been much of a performance concern, so thanks for verifying.

digitalcora (approved):
One comment still outstanding about a missing typespec, but looks good.

lib/screenplay/screens_config.ex
```diff
@@ -24,22 +24,14 @@ defmodule Screenplay.ScreensConfig do
...
-  def screen(screen_id) do
-    Cache.get(screen_id)
-  end
-
-  @spec screen_ids() :: list(String.t())
-  def screen_ids do
-    Cache.all(nil, return: :key)
-  end
-
+  @spec screens() :: list(Screen.t())
```
cmaddox5, line 27: @digitalcora Flagging this just in case it was missed. For some reason, GH is not flagging the comment as outdated after I pushed up the change.

digitalcora: Oh, weird. I know this has happened to me before but it seems to happen _just_ infrequently enough that I've forgotten by the next time. (like once a year?)

PR #502, opened 2024-09-27
refactor: Improve directory structure

cmaddox5:
More ad-hoc refactoring! Because we use the term `PlacesAndScreens` by default when describing the Screenplay config, I changed the folder and module names in hopes that they more clearly describe what they contain. The `permanent_config.ex` file also had no good reason to be in that folder so moved it up a level.


digitalcora (approved):
Sensible. I wonder if we could make the naming less awkward by just using the term `Place`, and saying that a "place" by definition includes all the screens associated with it. Do we ever deal with a _standalone_ "place", no screens? It seems like the concepts are intertwined.

cmaddox5:
> It looks like there are still some `places_and_screens` or `place_and_screens` function and variable names hanging around following the rename. I don't think tackling them right now is essential, just wanted to flag in case you weren't aware.

Ugh, sorry about that. Will address in a new PR soon.

PR #510, opened 2024-10-04
chore: Add credo config

cmaddox5:
I was going to disable a rule and realized we didn't have a credo config file. Generated one using `mix credo gen.config` and disabled the `Credo.Check.Refactor.Nesting` check. This PR can also be a forum to disagree/suggest more checks to disable.


digitalcora:
What do you think of using only `disabled` and `extra` to tweak the config from the defaults, instead of having a "pinned" copy of all the checks? I think I'd like it if we could automatically inherit changes to the defaults with new Credo versions — for example I'm looking at that _"Checks scheduled for next check update (opt-in for now)"_ section and thinking that if we do things the "pinned" way, we're opting out of these checks indefinitely, not just for now.

cmaddox5:
Yeah that's a good idea. Didn't think about how the config kinda forces you into a specific set of defaults out of the box. Cleans up the file quite a bit too.

.credo.exs
```diff
@@ -0,0 +1,35 @@
...
+      #
+      name: "default",
+      #
+      # These are the files included in the analysis:
+      files: %{
+        #
+        # You can give explicit globs or simply directories.
+        # In the latter case `**/*.{ex,exs}` will be used.
+        #
+        included: ["lib/"],
```
digitalcora, line 24: Interesting that it doesn't include `test/` by default. Inter-testing.

digitalcora (approved):

.credo.exs
cmaddox5: Woops, it actually does. It also comes with a bunch of directories it thinks you might have. When deleting those, I accidentally removed `test/`. Added it back.

PR #519, opened 2024-10-10
suggestion: Retry http_fetch_errors

cmaddox5:
About once a day, the `Builder` throws an error due to one of the API calls timing out. One possible solution I'm suggesting is retrying the fetch whenever `HTTPoison.get` returns an error.


digitalcora:
I'll "yes, and" this with a suggestion that if we're doing this for all API requests, we might consider adding a backoff, to avoid the retries themselves making the problem (whatever it is) worse. Screens already does this using the [Retry](https://hex.pm/packages/retry) package.

cmaddox5:
Yeah good idea. I kept `constant_backoff` but we can also use `exponential` or `linear` if those are preferred.

lib/screenplay/v3_api.ex
```diff
@@ -51,6 +47,23 @@ defmodule Screenplay.V3Api do
...
 
+  defp do_fetch(url, headers, opts) do
+    retry with: Stream.take(constant_backoff(500), 3) do
+      HTTPoison.get(
+        url,
+        headers,
+        Keyword.merge(@default_opts, opts)
+      )
+    after
+      result -> result
```
digitalcora, line 58: It seems like this is the default and can be omitted.

https://hexdocs.pm/retry/Retry.html#retry/2
> The `after` and `else` clauses are optional. By default, a successful value is just returned.

digitalcora (approved):
One minor comment. Let's deploy this to dev-green first and look for (absence of) errors, but otherwise seems good!

cmaddox5:
@digitalcora Unfortunately this did not seem to work in dev-green overnight. Sentry error [here](https://mbtace.sentry.io/issues/5983309607/events/c25c0ac423454b61bf47578d772f815c/). It is a new error but that's from me trusting dialyzer and messing up the pattern of the error (It can't figure out what will be returned). Even this looks like a new error, it only occurs for the same reason as before. Do you think this change would still be nice to have or should we explore other solutions first?

digitalcora:
Given this runs on a timer, we should be able to figure out how many of the rebuilds we'd expect to see in a given day are failing, both with this solution and without. Do the retries make an appreciable dent? If before it was failing 50% of the time and now it's 10%, that would sound like a great start, even if we could improve it later. If before it was 50% and now it's 40%, maybe we should investigate further. If before it was 10% and now it's 5%, maybe we can live with the failures for now since it's not a major issue.

digitalcora:
That doesn't sound bad at all — considering also the apparent complexity of addressing this and the fact that it's an out-of-band task, I'm fine with ignoring the crashes as long as they don't escalate. I went ahead and did an "Archive => Until escalating" on these, which should notify us again only if the error rate exceeds the historical weekly rate.

cmaddox5:
Sounds good. I can close this then. Was kind of a knee-jerk reaction to being annoyed by the error 😅.

PR #520, opened 2024-10-10
chore: Refresh session on request

cmaddox5:
Screenplay is very often left open in people's browser. This will eventually lead to an outdated session failing in our request pipeline and cause a CORS error due to a redirect. I followed [this](https://www.notion.so/mbta-downtown-crossing/Session-Management-ad53a774fd39462c82146f60700db5fc) Notion doc to get a sense of what our session best practices are and that helped me find [this](https://github.com/mbta/arrow/pull/967) PR. I followed in that PR's footsteps for the solution to our problem. We now set an idle time of 30 minutes (although the plug will prevent users from exceeding that) and a max session age of 12 hours.


lib/screenplay_web/auth_manager/auth_manager.ex
```diff
@@ -12,6 +12,15 @@ defmodule ScreenplayWeb.AuthManager do
     "pa-message-admin" => :pa_message_admin
   }
 
+  def max_session_time do
+    Application.get_env(:screenplay, __MODULE__)[:max_session_time]
+  end
+
+  def idle_time do
+    Application.get_env(:screenplay, __MODULE__)[:idle_time]
```
digitalcora, line 20: Since these are configured at compile time there is not much point in reading them at runtime. We could instead define e.g.

```elixir
@idle_time Application.fetch_env!(:screenplay, [__MODULE__, :idle_time])
```

And this would ensure at compile time that we are, in fact, configuring them as expected.

lib/screenplay_web/controllers/auth_controller.ex
```diff
@@ -18,11 +24,12 @@ defmodule ScreenplayWeb.AuthController do
...
 
     conn
+    |> configure_session(drop: true)
     |> Guardian.Plug.sign_in(
       ScreenplayWeb.AuthManager,
       username,
-      %{roles: roles},
-      ttl: {session_ttl_hours, :hours}
+      %{auth_time: auth_time, roles: roles},
+      ttl: {1, :minute}
```
digitalcora, line 32: Does this only work because we know our frontend always makes a request to the server at least once a minute? (just trying to get an idea of how this would look under different conditions, like if this were a plain server-rendered app)

cmaddox5: Correct. My understanding is that the value here can be anything as long as it's <= the idle time we configure. Because we are guaranteed to refresh every 4 seconds, 1 minute will always be enough.

lib/screenplay_web/auth_manager/auth_manager.ex
cmaddox5: You mean `compile_env!`? Either way, will make that change.

digitalcora (approved):
If this seems to work correctly on dev-green, looks good to me!

digitalcora:
If you express it this way it may work better:

```diff
- @idle_time Application.compile_env!(:screenplay, __MODULE__)[:idle_time]
+ @idle_time Application.compile_env!(:screenplay, [__MODULE__, :idle_time])
```

The difference should be that with the latter, `compile_env!` can "see" the specific key you're accessing (and that it doesn't change at runtime), instead of only knowing that you're accessing the `__MODULE__` group of keys (and then warning you because something in that group is set at runtime).

PR #525, opened 2024-10-15
chore: Update to eslint v9

cmaddox5:
Dependabot closed the PR I fixed so making my own. This updates eslint et al. There was a config update required. I made the update by converting our existing `.eslintrc.js` to a JSON config and followed [these](https://eslint.org/docs/latest/use/configure/migration-guide#migrate-your-config-file) instructions.


assets/eslint.config.mjs
```diff
@@ -0,0 +1,85 @@
...
+import tsParser from "@typescript-eslint/parser";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import js from "@eslint/js";
+import { FlatCompat } from "@eslint/eslintrc";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const compat = new FlatCompat({
+    baseDirectory: __dirname,
```
digitalcora, line 14: This file seems to be 4-spaced, rather than our usual 2-spaced.

digitalcora (approved):
🧹 Nice that we can delete some old ignores here.

assets/eslint.config.mjs
cmaddox5: Hmm yeah thank you. I guess the devs of the migrator tool are 4-spacers 😠

PR #526, opened 2024-10-15
suggestion: Handle session expiration gracefully

cmaddox5:
Even after #520, we still need to handle errors when someone has Screenplay open during a deploy. I think this can best be handled by looking for `api` in the previous path and returning a 403 instead of a redirect. This allows us to throw/catch an exception on the client and display a modal letting the user know they need to refresh the page. 

Screenshot 2024-10-15 at 1 51 35 PM


lib/screenplay_web/auth_manager/error_handler.ex
```diff
@@ -9,8 +9,12 @@ defmodule ScreenplayWeb.AuthManager.ErrorHandler do
 
   @impl Guardian.Plug.ErrorHandler
   def auth_error(conn, error, _opts) do
-    auth_params = auth_params_for_error(error)
-    Phoenix.Controller.redirect(conn, to: ~p"/auth/keycloak?#{auth_params}")
+    if Plug.Conn.get_session(conn, :previous_path) =~ "api" do
```
digitalcora, line 12: Shouldn't we be looking at the _current_ request path rather than the previous one? I'm surprised we set a "previous path" for API requests at all — I'd expect that value to always be some browser-navigable route.

assets/js/components/Dashboard/Dashboard.tsx
```diff
@@ -52,23 +55,35 @@ const Dashboard: ComponentType = () => {
...
+              alerts: newAlerts,
+              allAPIAlertIds: allAPIalertIds,
+              screensByAlertMap: screensByAlertMap,
+            });
+          },
+        )
+        .catch((response: Response) => {
+          if (response.status === 403) {
+            setIsAlertsIntervalRunning(false);
+            setShowModal(true);
```
digitalcora, line 79: It seems we'd only show this modal if, specifically, a background alerts request fails. How is the experience if some other kind of "foreground" request fails? Should we try to integrate this into all instances where we fetch data from the server?

lib/screenplay_web/auth_manager/error_handler.ex
cmaddox5: I think the main reason for holding onto the previous path is for the sake of auth redirects. When you log in to Keycloak, you're temporarily not at the URL you want to be at so we help the user get there by holding on to their original request path. I used it here just in case it landed in this function from a Keycloak request. Because `save_previous_path/2` is part of the request pipeline, it would be the same URL regardless of where the error surfaced.

assets/js/components/Dashboard/Dashboard.tsx
cmaddox5: Because this task was already kind of scope creep-y, I focused specifically on what I knew to be the problem we see in Sentry.

You do raise a good point though. We've had a habit of assuming a request will be successful so we unintentionally fail silently in some areas. For example, I hardcoded a 500 for the `Associate with alert` page and the only indicator that something went wrong is that the table is empty. Because this would be a bigger task, I'll get it added to the backlog for us to talk about in the future.

lib/screenplay_web/auth_manager/error_handler.ex
digitalcora: I think we should check the current request path in any case, given that sometimes we may not have a previous path or it may be a non-API page path (like for the first API request that happens after doing a full page load). "If current path is `api` _or_ previous path is `api`" in other words.

assets/js/components/Dashboard/Dashboard.tsx
digitalcora: Did you end up writing this task? (sorry for letting this review drop off)

cmaddox5: Just wrote it up [here](https://app.asana.com/0/1185117109217422/1208769995914082/f).

digitalcora (approved):
This is a good improvement to the UX as-is, so let's :shipit:

PR #529, opened 2024-10-17
feat: Elevator Screens

cmaddox5:
Now that we can make elevator screens, we need to make them work in Screenplay. This involves a few things:

- Adding a way to determine `stop_id` in `Builder`. Did this with a `/facilities` lookup
- Correcting the ID in some existing client code (was added a long time ago but we guessed ID incorrectly)
- Add styles so `iframe`s display correctly

Deploying this to dev-green so it can be seen in a deployed environment.

---
  - https://app.asana.com/0/0/1208310396563617


lib/screenplay/places/builder.ex
```diff
@@ -228,6 +234,11 @@ defmodule Screenplay.Places.Builder do
     stop_id
   end
 
+  defp get_stop_id({_id, %{app_params: %Elevator{elevator_id: elevator_id}}}) do
+    {:ok, facility} = @facilities_mod.fetch(elevator_id)
+    get_in(facility, ~w[relationships stop data id])
```
digitalcora, line 239: Is this allowed to return `nil`? If not, we should fetch the field in some other way or guard against it — `get_in` returns `nil` if any part of the path doesn't exist.

One way to do this would be destructuring:

```elixir
%{"relationships" => %{"stop" => %{"data" => %{"id" => stop_id}}}} = facility
stop_id
```

test/fixtures/builder/screens_config.json
```diff
@@ -345,7 +345,7 @@
           "medium_asset_path": "images/bus-shelter/survey/10413-weld-sb-medium.png"
         }
       },
-      "vendor": "lg-mri",
+      "vendor": "lg_mri",
```
digitalcora, line 348: This is now already fixed on `main`; it should disappear with a rebase.

lib/screenplay/places/builder.ex
cmaddox5: Good catch. I neglected to think about bad data/errors occurrence in the fetch.

assets/js/components/Dashboard/PlaceRowAccordion.tsx
```diff
@@ -23,11 +23,12 @@ type ScreenGroup = {
 
 const groupScreens = (screens: Screen[]): ScreenGroup[] => {
   const inlineScreens = screens.filter((screen) =>
-    ["busway_v2", "solari"].includes(screen.type),
+    ["busway_v2", "solari", "elevator_v2"].includes(screen.type),
```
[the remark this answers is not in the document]
cmaddox5: Going the flexbox route should be easy. Just need to add `flex-wrap`. Wondering if it would be helpful to group screens of the same type into each inlined group. Without it, two sections and two elevator screens leads to 3 screens in one row and one in another. Grouping first would split them evenly. What do you think?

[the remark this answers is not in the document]
digitalcora: I think screens of the same type should be grouped, yes. It feels like an oversight that they aren't (maybe we only had one "inline" screen type at first).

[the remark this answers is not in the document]
cmaddox5: Yeah only Solari/Sectionals were inline before this. Fixed them up so they are grouped by type.

assets/js/components/Dashboard/PlaceRowAccordion.tsx
```diff
@@ -15,20 +15,22 @@ import ScreenDetail from "Components/ScreenDetail";
 import { sortScreens } from "../../util";
 import { useUpdateAnimation } from "Hooks/useUpdateAnimation";
 import classNames from "classnames";
+import _ from "lodash";
```
digitalcora, line 18: I thought we were using `lodash/fp`?

digitalcora (approved):
One comment, looks good otherwise!

assets/js/components/Dashboard/PlaceRowAccordion.tsx
cmaddox5: Ahh yes thank you. (Although I dislike how bad it is at inferring types 😅)

digitalcora: I agree this is a notable downside of FP and maybe something we should have considered (if we'd had a better sense for it at the time) in our conversations about which "utility library" to use. I tend to avoid Lodash entirely and use vanilla JS where feasible, since built-in classes and functions have generally excellent types (except for you, `JSON.parse`, you're not invited to the party).

PR #536, opened 2024-10-22
chore: Fake Watts Client

cmaddox5:
I got annoyed by not being able to create PA messages locally without running Watts. For local dev, we don't care what the text sounds like so just going to return a static mp3 instead.


lib/screenplay/watts/client.ex
```diff
@@ -41,3 +49,26 @@ defmodule Screenplay.Watts do
...
+  @behaviour Screenplay.Watts.ClientBehaviour
+
+  @impl true
+  # sobelow_skip ["Traversal.FileModule"]
+  def fetch_tts(_) do
+    path = Path.join(:code.priv_dir(:screenplay), "static.mp3")
+
+    case File.read(path) do
+      {:ok, file} ->
+        {:ok, Jason.decode!(file)}
```
panentheos, line 67: This `Jason.decode!` doesn't seem right

cmaddox5: Thank you. Muscle memory kicked in.

panentheos (approved):
Looks good, tested locally

PR #539, opened 2024-10-23
feat: PA Messages End Now

cmaddox5:
Added the ability to end an active message. Did this by setting the `end_datetime` of the message to now minus 1 second (w/o the minus 1 second, the active query still considers the message as active and doesn't move the message to Done).

---
  - https://app.asana.com/0/0/1208255736020795


assets/js/components/Dashboard/ThreeDotsDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </OverlayTrigger>
+  ),
+);
+
+// Expects children to be a group of `Dropdown.Item`s
+interface Props extends PropsWithChildren {
+  tooltipText?: string;
+}
+
+const ThreeDotsDropdown = ({ children, tooltipText }: Props) => {
```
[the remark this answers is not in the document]
cmaddox5: Totally forgot about Kebabs! I think that fits best here considering that is the UX term for a component like this.

assets/css/dashboard/pa-messages-page.scss
```diff
@@ -121,4 +121,55 @@
     align-items: center;
     margin-top: 48px;
   }
+
+  .three-dots-vertical-dropdown {
```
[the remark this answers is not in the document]
cmaddox5: Hmm yeah good point. I think I'll let the style of the menu on this page be the default and place them in a new `kebab-menu.scss`. Will then only include bg colors overrides in `screen-detail-action-bar`.

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -14,11 +15,13 @@ import { Link, useNavigate, useSearchParams } from "react-router-dom";
 import cx from "classnames";
 import useSWR, { mutate } from "swr";
 import { useRouteToRouteIDsMap } from "Hooks/useRouteToRouteIDsMap";
+import KebabMenu from "Components/KebabMenu";
+import moment from "moment";
```
digitalcora, line 19: minor suggestion: I find it helps readability of imports when third-party packages are grouped together, then imports within our own app, instead of being mixed.

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -14,11 +15,13 @@ import { Link, useNavigate, useSearchParams } from "react-router-dom";
...
 import { updateExistingPaMessage } from "Utils/api";
 import { UpdatePaMessageBody } from "Models/pa_message";
-import Toast from "Components/Toast";
+import Toast, { type ToastProps } from "Components/Toast";
 
-type StateFilter = "current" | "future" | "past";
+type StateFilter = "current" | "future" | "done";
```
digitalcora, line 24: Because these values are exposed in the URL, this will break any existing open tabs/bookmarks to the "Past/Done" page (as we just recently experienced with renaming `active` to `current`). Can we keep these values the same, and just change how the UI refers to it? I think it's probably fine if the URL still says "past", and `past` as an internal name might actually make more sense for us due to the symmetry with `future`.

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -261,14 +262,36 @@ const PaMessageTable: ComponentType<PaMessageTableProps> = ({
               <PaMessageRow
                 key={paMessage.id}
                 paMessage={paMessage}
-                stateFilter={stateFilter}
-                updatePaMessage={updatePaMessage}
+                onEndNow={() => {
+                  setToastProps({
+                    variant: "info",
+                    message: "PA/ESS message has ended, and moved to “Done.“",
```
digitalcora, line 268: Since we've used an opening quote `“` here, the second quote should be a closing one `”`.

assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -116,20 +119,12 @@ const PaMessagesPage: ComponentType = () => {
...
+    stateFilter,
+  });
 
-    const updatedData = data.map((paMessage) => {
-      return paMessage.id === updatedPaMessage.id
-        ? updatedPaMessage
-        : paMessage;
-    });
-
-    mutate(`/api/pa-messages?${params.toString()}`, updatedData, false);
```
digitalcora, line 131: I don't quite understand the implications of how you've changed the state tracking here. Are there any changes to how the UI behaves in practice?

assets/js/components/Dashboard/PaMessagesPage.tsx
cmaddox5: Woops, thanks.

assets/js/components/Dashboard/PaMessagesPage.tsx
cmaddox5: I'd say the key difference is ending the message or switching the toggle will now invalidate the client state and trigger a refetch. The UI behavior should be unchanged outside of maybe outside changes showing as a result of the refetch. I felt going the new direction simplified callbacks a bit.

assets/js/components/Dashboard/PaMessagesPage.tsx
cmaddox5: Good callback. I can change these back.

digitalcora (approved):
🔚 ⌚ ✅

PR #541, opened 2024-10-24
chore: Upgrade to eslint v9

cmaddox5:
All of our eslint plugins support v9 now so moving forward with updating. Biggest change was to the config. `.eslintrc.js` and `.eslintignore` are no longer supported. They provided a [migrator](https://eslint.org/docs/latest/use/configure/migration-guide#migrate-your-config-file) that I used to convert to the new format.


cmaddox5:
Duplicate of https://github.com/mbta/screenplay/pull/525 😅

PR #542, opened 2024-10-25
chore: Replace Quantum w/ Oban for cron job.

cmaddox5:
After a conversation with Ian around Screenplay uptime, he recommended bumping up our minimum number of instances to 2, confirming what we originally thought we needed to do. To prepare for this, I glanced over the code to find anything that could be impacted by this change. The only thing I found was the `TakeoverToolTestingJob`. 

This job does not need to run more than once and might actually cause issues if it does. To resolve this, I'm proposing we replace Quantum with Oban for processing this job. This will allow us to keep the original scheduling of the job while ensuring it only runs a single time regardless of the number of instances.


config/runtime.exs
```diff
@@ -60,11 +60,14 @@ if sentry_dsn not in [nil, ""] do
...
-    do: [{"0 7 * * *", {Screenplay.Jobs.TakeoverToolTestingJob, :run, []}}],
-    else: []
-
-config :screenplay, Screenplay.Scheduler, jobs: scheduler_jobs
+if env == "prod" do
+  config :screenplay, Oban,
+    plugins: [
+      {Oban.Plugins.Cron,
+       crontab: [
+         {"* * * * *", Screenplay.Jobs.TakeoverToolTestingJob}
```
digitalcora, line 68: I'm sure we could stand to run this more frequently, but running it every minute of every hour of every day seems excessive. Why the schedule change?

config/runtime.exs
```diff
@@ -60,11 +60,14 @@ if sentry_dsn not in [nil, ""] do
...
 
-scheduler_jobs =
-  if env == "prod",
-    do: [{"0 7 * * *", {Screenplay.Jobs.TakeoverToolTestingJob, :run, []}}],
-    else: []
-
-config :screenplay, Screenplay.Scheduler, jobs: scheduler_jobs
+if env == "prod" do
+  config :screenplay, Oban,
+    plugins: [
```
digitalcora, line 65: We should probably also use `Pruner`. Per [the docs](https://hexdocs.pm/oban/Oban.Plugins.Pruner.html):

> Pruning is critical for maintaining table size and continued responsive job processing. It is recommended for all production applications.

It also seems like we'd want the default behaviors of `Lifeline` and `Reindexer`.

config/runtime.exs
cmaddox5: Very much unintentional. Set it to that for the sake of testing and forgot to revert. Thank you.

config/runtime.exs
cmaddox5: Ahh, missed those in the docs. Thanks.

digitalcora (approved):
I see this hasn't yet been tested on dev-green, so let's do that first. Otherwise looks good!

cmaddox5:
> I see this hasn't yet been tested on dev-green, so let's do that first. Otherwise looks good!

Because this job is only set up to run in prod, do you think it's enough to see dev-green work as usual or would a small sample job that always runs give more confidence?

digitalcora:
Let's temporarily modify the code so it also runs the real test in dev-green? Maybe at a different time.

cmaddox5:
The infra is not quite set up to allow a successful connection. I can spin up a super quick task that just logs something unique and look for it in Splunk.

cmaddox5:
Will let it run for a bit but seems to be working great. Splunk query: `index="screenplay-dev-green-application" "oban test"`.

PR #544, opened 2024-10-29
chore: remove support for v1 screen types

digitalcora:
We no longer have any "v1" screens configured. Since mbta/screens-config-lib#40 the config structs/types are removed entirely, so there should be no references remaining in this change.


cmaddox5 (approved):
Love it.

PR #550, opened 2024-11-01
validate PA message interval field

panentheos:
**Asana task**: [bug: PA messages accept negative interval](https://app.asana.com/0/1185117109217413/1208634633982716/f)

This fixes an issue where PA messages would allow setting a negative interval. Enforces validation on the front and back end. Also validates the `priority` field for completeness.


digitalcora (approved):
1️⃣➕

PR #551, opened 2024-11-01
fix current PA message filter

panentheos:
**Asana task**: [bug: Can't see PA messages that don't play on the current day](https://app.asana.com/0/1185117109217422/1208664128012797/f)

This fixes an issue where the `current` filter wouldn't show messages that don't play on the current day. Moves the day condition to the `active` filter so it continues to apply there.


digitalcora (approved):
📅 ✅

PR #555, opened 2024-11-05
Store start/end date and times as strings

PaulJKim:
**Asana task**: [[Bug] PA/ESS message Screenplay: Minute and AM/PM fields unexpectedly update in Chrome browser](https://app.asana.com/0/1185117109217413/1208573903710847/f)

Description
Store these values as strings in state to prevent the form input from defaulting to midnight when zero is entered as the hours on google chrome.


PaulJKim:
@panentheos I think this is the approach you mentioned in yesterdays meeting. I noticed though that a double zero press is still handled a bit differently in Chrome. Chrome sees it as `12` and doesn't change the AM/PM whereas Firefox sees it as `00` and changes to AM, which seems correct to me. Do you think that's just a quirk between the two browsers or is there a point I might've missed in the approach you outlined?

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "screenplay",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}
```
digitalcora, line 6: Accidentally ran `npm` in the root of the project instead of `assets`?

panentheos:
This is looking good. We'll also need some client-side validation to check that the inputs are in a good state, and print an appropriate error if they're not.

PaulJKim:
@panentheos It looks to me like there's already some client-side validation built into the bootstrap component. Only numerical inputs are allowed and any time there is a valid input, the value passed to the onChange callback is always in the format `HH:mm`.

panentheos:
Right, the component behaves well now, but if you try to submit with it in an invalid state, there's no field-level error message that indicates what's wrong.

PaulJKim:
@panentheos I added some error messaging for when the time field is invalid. I guess I'm curious though what exactly you mean by submitting with the time in an invalid state? From my POV, the component doesn't allow the field to ever get into an invalid state. Or do you mean that we should have the error messaging there just in case the component changes at some point and does allow it?

PaulJKim:
@panentheos gotcha, I added similar treatment for the date fields. I also added a bit to the validation logic to make the messaging a little clearer to the user.

panentheos (approved):
Looks good!

PR #556, opened 2024-11-06
fix: Actions column styles

cmaddox5:
Mary expressed some concerns over styles [here](https://mbta.slack.com/archives/C039ARUM48H/p1730912294881359?thread_ts=1730904072.858599&cid=C039ARUM48H). Made some adjustments to the layout so what she called out is implemented correctly.


digitalcora (approved):
🧹 ✨

PR #557, opened 2024-11-07
chore: OFM Reminders Job Processing Change

cmaddox5:
While fixing the webhook URL for the Slack app, I remembered this being another good candidate for Oban processing. It still runs once a minute but should only run a single time each minute now. Should prevent the reminder from going off twice when it needs to.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -65,7 +65,8 @@ if env == "prod" do
     plugins: [
       {Oban.Plugins.Cron,
        crontab: [
-         {"0 7 * * *", Screenplay.Jobs.TakeoverToolTestingJob}
+         {"0 7 * * *", Screenplay.Jobs.TakeoverToolTestingJob},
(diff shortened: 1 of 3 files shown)


digitalcora (approved):
If there's a feasible way for us to test this on dev-green before merging, I'd recommend we do so — I don't have access to the channel these alerts go to, so I'm not sure if it already has been tested. But approving the change in principle.

cmaddox5:
> If there's a feasible way for us to test this on dev-green before merging, I'd recommend we do so — I don't have access to the channel these alerts go to, so I'm not sure if it already has been tested. But approving the change in principle.

Will probably need to hardcode the webhook URL temporarily and allow the job to run on dev-green. Should be easy though.

cmaddox5:
Was able to successfully send a message to #tid-oio. Should be good to merge after CI is good.

PR #561, opened 2024-11-14
introduce CSS modules

panentheos:
This converts one of the PA message CSS files to use CSS modules. The goal here was to maintain the current styling, with a few minor exceptions (a handful of hover and focus colors). While I think there are some improvements we could make to this page, I didn't want to tie those to this purely technical change. Notes:

* Due to the explicit scoping of CSS modules, there is no possibility for accidental leakage. That means we can flatten out the selector structure and use less verbose names.
* This leverages Bootstrap variables to modify the styling of buttons, etc. Given that we're bought into Bootstrap, this tends to encourage more consistent treatment, compared to overriding the styles explicitly.
* Also uses builtin Bootstrap utility classes to apply common styling patterns, especially for flexbox layout and margins/padding. This reduces the need for bespoke classes in many cases.
* Bootstrap also has the ability to generate additional utility classes, which we could use to replace some one-off classes. For example, `.alertEndedText`, which just sets `color` to a standard value, could be replaced with a generic utility if we had it.
* Some one-off classes might be better expressed via inline styles. For example, `.startEndItem`, which just sets a specific `flex` value, could probably be an inline style.


assets/css/pa-messages.module.scss
```diff
@@ -0,0 +1,204 @@
...
+@mixin flat-button {
+  --bs-btn-active-color: var(--bs-btn-color);
+  --bs-btn-hover-color: var(--bs-btn-color);
+  --bs-btn-active-bg: var(--bs-btn-bg);
+  --bs-btn-hover-bg: var(--bs-btn-bg);
+  --bs-btn-active-border-color: var(--bs-btn-border-color);
+  --bs-btn-hover-border-color: var(--bs-btn-border-color);
+}
+
+.editPage {
```
digitalcora, line 12: If we'd like to keep using `kebab-case` within CSS, but use the more convenient `camelCase` when referencing styles in JS, it looks like css-loader has an [exportLocalsConvention](https://webpack.js.org/loaders/css-loader/#exportlocalsconvention) option that can accommodate this. I'd slightly prefer this, especially given we're still using kebab-case names for mixins and variables. But it may also be convenient to have the names match up exactly for cross-referencing.

panentheos: Interesting. I'd be hesitant to pay the cognitive overhead of having to translate between the two, especially since doing so would make grep-based exploration of the code more difficult. I'm curious to head other opinions on this.

As for the generated names, yes, we can tweak that so it's more amenable to debugging.

cmaddox5 (approved):
This is great. Assuming the team is all in on this, do we want to introduce a "New pages use CSS modules" policy? Then we would only have to worry about migrating existing styles.

PR #562, opened 2024-11-15
chore: Remove AWS Cert hash check

cmaddox5:
Hash is outdated. Got the new one using `curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem | shasum -a 256`.

Is this check useful? It being outdated results in deployment errors and we end up needing to blindly follow whatever the `curl` command returns anyway.


digitalcora (approved):
> Is this check useful?

I'm not entirely convinced it is; it seems some other teams have already dropped it. The scenarios it would protect against are fairly esoteric, and I haven't found any recommendation from AWS itself that users of the bundle should be doing this check.

PR #566, opened 2024-11-18
fix: PaMessagesPage accessibility bug

cmaddox5:
Looks like the recent ESLint update introduced a new rule that we were breaking. Fixed that by moving the `onClick` from a `td` to the kebab menu itself.

I also noticed that the rows are not part of tab navigation. This causes a mouse click to be the only way to navigate to the edit page. Resolved that by making the `Message` column a link.


digitalcora (approved):
🧹

PR #567, opened 2024-11-19
feat: Static Templates

cmaddox5:
This PR adds a static templates feature. We will keep a list of static templates in `priv/config/static_templates.json`. When a message is created using a template:
- visual and phonetic text cannot change
- priority and interval default to a value determined by the type of template selected
    - These values can change after defaults are applied
- message cannot be associated with an alert

I have also added an `archived` field to each static template in order to prevent IDs from being reused by mistake. If a template is deleted and the ID is used again later, any message linked to the deleted template would then be incorrectly linked to the new one. This is unlikely to happen, but wanted to prevent it nonetheless.


lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -49,11 +52,28 @@ defmodule ScreenplayWeb.PaMessagesApiController do
...
+        alert: if(is_nil(alert), do: alert, else: Alert.to_full_map(alert))
+      })
     else
       conn
       |> put_status(404)
       |> json(%{error: "not_found"})
     end
   end
+
+  def static_templates(conn, _) do
```
panentheos, line 70: Apologies if this was already discussed, but do we need to serve these from the backend? I was imagining just compiling the template text into the frontend bundle and accessing it statically, since that's the only place we need it.

cmaddox5: I don't think server vs. client was ever discussed, but that's an interesting idea. It didn't click until now that the server has no reason to care about the contents of the file. That should simplify some things so I can give that a shot and see how we feel.

assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
```diff
@@ -502,53 +511,91 @@ const NewPaMessageHeader = ({
...
+      case "psa":
+        return "PSA";
+      case "emergency":
+        return "Emergency";
+      default:
+        return "";
+    }
+  };
+
+  let content;
```
panentheos, line 525: nit: Since we're just returning the content immediately, we could put `return` statements inside each `if` and avoid a `let` here.

assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
```diff
@@ -230,6 +262,19 @@ const PaMessageForm = ({
           onCancel={() => setPage(Page.MAIN)}
         />
       )}
+      {page === Page.TEMPLATES && (
+        <StaticTemplatePage
+          onCancel={() => setPage(Page.MAIN)}
+          onSelect={(template) => {
+            setSelectedTemplate(template);
+            setVisualText(template.audio_text);
```
panentheos, line 270: This should be:
```suggestion
            setVisualText(template.visual_text);
```

panentheos (approved):
See comments, then looks good.

PR #569, opened 2024-11-25
validate that PA messages specify at least one day

panentheos:
**Asana task**: [bug: Can create a PA message with no days selected](https://app.asana.com/0/1185117109217413/1208664128012801/f)

This adds client-side validation for the day-of-week picker, and also server-side for completeness.


digitalcora (approved):
📅1️⃣➕

PR #570, opened 2024-11-25
log PA message create and update

panentheos:
**Asana task**: [Add log entry to capture PA message create/update](https://app.asana.com/0/1185117109217413/1208670991547802/f)

This adds log entries for PA message create/update actions, as specified

### DIFF
--- a/lib/screenplay_web/controllers/pa_messages_api_controller.ex
+++ b/lib/screenplay_web/controllers/pa_messages_api_controller.ex
@@ -30,7 +30,8 @@ defmodule ScreenplayWeb.PaMessagesApiController do
   end
 
   def create(conn, params) do
-    with {:ok, _} <- PaMessages.create_message(params) do
+    with {:ok, message} <- PaMessages.create_message(params) do
+      log_pa_message("pa_message_created", message, conn)
       json(conn, %{success: true})
     end
   end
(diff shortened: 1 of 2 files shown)


lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -56,4 +58,17 @@ defmodule ScreenplayWeb.PaMessagesApiController do
       |> json(%{error: "not_found"})
     end
   end
+
+  defp log_pa_message(event, message, conn) do
+    Screenplay.Util.log(event,
```
cmaddox5, line 63: I know it isn't called out in the task, but should we include `alert_id`? Just want to mention it in case it was an oversight.

lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -30,14 +30,16 @@ defmodule ScreenplayWeb.PaMessagesApiController do
   end
 
   def create(conn, params) do
-    with {:ok, _} <- PaMessages.create_message(params) do
+    with {:ok, message} <- PaMessages.create_message(params) do
```
cmaddox5, line 33: Do you think it's worth it to log failures that make it to the [FallbackController](https://github.com/mbta/screenplay/blob/main/lib/screenplay_web/controllers/fallback_controller.ex)? We may not specifically know which even failed but would be notified when they happen.

cmaddox5 (approved):
Looks good. Just one non-blocking question and a potential suggestion.

PR #573, opened 2024-11-27
add per-environment styling

panentheos:
**Asana task**: [Indicator to distinguish Screenplay environment](https://app.asana.com/0/1185117109217413/1208717090153970/f)

This adds custom colors and favicons based on the configured `ENVIRONMENT_NAME`. The SVG favicons should work in modern browsers, and older browsers will fall back to the ico file (which I also updated). Also converts the CSS file to use modules. 

- [X] For features with a design/UX component, deployed branch to dev-green and let product know it's ready for review.


lib/screenplay_web/templates/layout/app.html.heex
```diff
@@ -34,6 +34,8 @@
     <% end %>
     <%= csrf_meta_tag() %>
     <title>Screenplay</title>
+    <link rel="icon" href={Routes.static_path(@conn, "/favicon.ico")} sizes="32x32">
+    <link rel="icon" href={Routes.static_path(@conn, "/images/favicon-#{@environment_name || "prod"}.svg")} type="image/svg+xml">
```
PaulJKim, line 38: I'm not certain if it's just a local bug for me, but my browser can't seem to load the prod favicon when running the branch. Dev and dev-green work though

PaulJKim: Yeah nevermind, maybe a weird caching thing or something? Works now after directly loading the path to the favicon once

PR #595, opened 2025-01-24
PA message flow read only

PaulJKim:
**Asana task**: [Read-only mode for PA Message workflow](https://app.asana.com/0/1185117109217413/1208981851846577/f)

Description
Prior to this change, only PA Message admins were able to access the PA Message flow. With this change, all users will now be able to access the PA Message flow, but non-admin users will get a read-only experience.

### DIFF
--- a/assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
@@ -64,6 +64,7 @@ interface Props {
   paused: boolean;
   selectedTemplate: StaticTemplate | null;
   onClearSelectedTemplate: () => void;
+  isReadOnly?: boolean;
 }
 
 const MainForm = ({
@@ -103,6 +104,7 @@ const MainForm = ({
   paused,
   selectedTemplate,
   onClearSelectedTemplate,
+  isReadOnly = false,
 }: Props) => {
   const navigate = useNavigate();
   const [validated, setValidated] = useState(false);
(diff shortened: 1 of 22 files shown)


assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
```diff
@@ -172,6 +174,7 @@ const MainForm = ({
             setEndWithEffectPeriod={setEndWithEffectPeriod}
             selectedTemplate={selectedTemplate}
             onClearSelectedTemplate={onClearSelectedTemplate}
+            isReadOnly={isReadOnly}
```
panentheos, line 177: This is totally fine as implemented, but as a future note, using Context for global state that's used widely can save a bunch of boilerplate. In this case, if this value was passed via Context and accessed with a custom hook (e.g. `useIsPaMessageAdmin`), it could eliminate a bunch of intermediate wiring.

panentheos (approved):
Looks good! Tested by clicking around locally.

assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
PaulJKim: Ah yeah, I did consider that approach as well! But I wasn't certain if using context was overkill somehow. I'd be happy to modify the approach to use context rather than passing it down as a prop!

panentheos: I'd say only if you think it's worth it. There's nothing wrong with explicitly passing values, and it's simpler and more explicit that way, so there are tradeoffs to consider. I just wanted to call out the alternative.

PaulJKim: Gotcha. I did ultimately go with this approach initially because I thought that it was more simplistic, so I'll keep it this way!

PR #602, opened 2025-02-03
Language changes to filter PA/ESS and alerts by time

deanshi:
**Asana task**: [Language changes to filter PA/ESS and Alerts by time, and describe PA/ESS message type](https://app.asana.com/0/1185117109217413/1208918421836698)

Description
- "Now" switched to "Live"
- "Done" switched to "Past"
- "Current Service Disruption" to "Ongoing service disruption"
- "Planned Service Disruption" to "Upcoming service disruption"

Screenshot 2025-02-03 at 12 01 20 PM
Screenshot 2025-02-03 at 12 01 28 PM
Screenshot 2025-02-03 at 12 01 39 PM

### DIFF
--- a/assets/js/components/Dashboard/PaMessagesPage.tsx
+++ b/assets/js/components/Dashboard/PaMessagesPage.tsx
@@ -161,7 +161,7 @@ const PaMessagesPage: ComponentType = () => {
             </section>
             <FilterGroup
               className="mb-5"
-              header="Message state"
+              header="Filter by message state"
               onFilterSelect={(selected) =>
                 setStateFilter(selected as StateFilter)
               }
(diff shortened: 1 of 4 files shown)


assets/js/components/Dashboard/PaMessagesPage.tsx
```diff
@@ -161,19 +161,19 @@ const PaMessagesPage: ComponentType = () => {
             </section>
             <FilterGroup
               className="mb-5"
-              header="Filter by message state"
+              header="Message state"
```
panentheos, line 164: I think this change is a step backwards in terms of accessibility (posted a comment in the Miro design as well). There aren't any section headers or other landmarks here, so without the words "filter by", there's no way for a screen reader to know that these controls are filters. (As a side note, the controls themselves aren't really marked up properly either, which we should think about fixing as a separate effort)

deanshi: That makes perfect sense. I'll make those changes from what we discussed in standup and update the ticket to make it consistent on the other page.

deanshi: Switched the language back to be "Filter by" but also added it on the static template page to be consistent (with approval from design).

panentheos (approved):

PR #604, opened 2025-02-03
fix: enable non-PA-admins to view PA messages

digitalcora:
**Asana task**: https://app.asana.com/0/1185117109217413/1209298740143904

This includes a significant routing cleanup in an attempt to make the overall module more readable — some things were scattered around when they were actually related, there was some unneeded duplication, etc. Best reviewed per-commit.


lib/screenplay_web/router.ex
```diff
@@ -1,156 +1,118 @@
...
-    ])
+  # PA Message Management
 
-    get("/pending", ConfigController, :index)
-    get("/configure-screens/*app_id", ConfigController, :index)
+  scope "/pa-messages", ScreenplayWeb do
+    pipe_through([:browser, :authenticate])
+
+    get("/", PaMessagesController, :index)
+    get("/new", PaMessagesController, :index)
```
PaulJKim, line 82: Should the `/new` route now also use the `:ensure_pa_message_admin` plug?

digitalcora: Since all of these route to the same action (`PaMessagesController.index`), I don't think it would make sense for one of them to be protected and the other not. The only thing these routes "do" is load the frontend app, which then takes over routing. On the backend, we only need to protect actions that will actually change data (assuming that _viewing_ things is always okay, which I think it is).

lib/screenplay_web/router.ex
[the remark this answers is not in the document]
deanshi: Me as well hahaha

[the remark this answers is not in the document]
digitalcora: Thanks for the reminder that I should make some updates to the tests we already have for these endpoints — they all still pass, of course, but some are now running with more permissions than they should (i.e. they're not properly testing that you can view PA messages even when you're not a PA admin), and we could stand to have a few more cases that verify you get a 403 when you _don't_ have permission to do something.

As far as testing that the overall reorganization hasn't affected the route structure of the app, I mentioned this in the commit message, but I diffed the output of `mix phx.routes` to check this. It wasn't 100% straightforward since the refactor changed the order in which some routes are defined (intentionally, to move related routes closer to each other), but this also affects the order in which they're matched, and we have some routes that "mask" others. For example we define a `/api/pa-messages/preview_audio` which, if it were below the `/api/pa-messages/:id` route for fetching a specific message, would never match. Having basically any meaningful tests for the route will catch this, though, and it did in this case. (It will also sometimes result in a "function clause will never match" warning due to how route macros are compiled into pattern matches, but seemingly not always.)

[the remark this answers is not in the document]
digitalcora: Oh, right, also, local testing: in `config/dev.exs` we have this

```elixir
config :ueberauth, Ueberauth,
  providers: [
    keycloak:
      {Screenplay.Ueberauth.Strategy.Fake,
       [roles: ["screenplay-emergency-admin", "screens-admin", "pa-message-admin"]]}
  ]
```

Changing the list of roles here changes your assumed roles when running the app locally. So for example I could remove `pa-message-admin` and that would switch me to the read-only PA message view.

lib/screenplay_web/router.ex
PaulJKim: Actually bringing this thread back (tangentially) - do you mind making an update to `NewPaMessage.tsx` on my behalf? Something this comment made me realize I overlooked was that an unauthenticated user could type in the url `https://screenplay.mbta.com/pa-messages/new` and be brought to an editable pa message form. Although they wouldn't be able to submit a new pa message now with this PR 🙏 , this still seems like an odd thing. We can just make another call to `isPaMessageAdmin()` and pass it to the PaMessageForm component.

digitalcora: Not totally convinced it makes sense to support the "use case" of someone manually typing in URLs to get places they shouldn't... but I went ahead and did this and saw some other opportunities to clean up the role-checking code while I was at it. See the two new commits on this branch.

robbie-sundstrom (approved):
The split up commits make this really easy to digest! Also appreciate the info on how to test screenplay with different permissions

PR #605, opened 2025-02-05
Create a common table component for use across the PA Messages, Static Template and Associate Alert pages

deanshi:
**Asana task**: [PA Message UI: standardize tables](https://app.asana.com/0/1209229552734401/1208918444026658)

Description
- Created a `MessageTable` component for use across the PA Messages, Static Template and Associate Alert pages
- Uses a common `messages-table.scss` in order to standardize most of the tables look and feel
- Rows seem to be slightly more custom so I kept them per page rather than trying to standardize them. If we want to potentially make them more standard, I can try pulling things out from there as well (or at the very least pull them into their own components with their own .scss files). It's mainly the PA Messages table causing a little bit of trouble with making things generic
- Pulled out empty state + loading into the new `MessagesTable`, although it's only being used in one place

QA/Compared to [this Miro link](https://miro.com/app/board/uXjVL7Upw18=/?moveToWidget=3458764609622855805&cot=14) and things look the same. Let me know if there are specific UI details I missed or even functionality that I missed, also naming changes that need to come through. 


assets/css/dashboard/new-pa-message/message-tables.scss
```diff
@@ -0,0 +1,83 @@
+.message-table {
```
panentheos, line 1: Sorry, should have mentioned this up front, but we're starting to migrate new CSS to use CSS modules as a way to simplify and reduce style leakage. This would be a good opportunity to use a module file for this new CSS, with explicitly imported classnames.

assets/js/components/Dashboard/PaMessageForm/AssociateAlert.tsx
```diff
@@ -79,11 +103,22 @@ const AssociateAlert = ({ onApply, onCancel }: AssociateAlertPageProps) => {
...
-            <AssociateAlertsTable
-              alerts={alerts}
-              onSelectAlert={setSelectedAlert}
-              messageStateFilter={selectedMessageState}
-              serviceTypeFilter={selectedServiceType}
+            <MessageTable
+              isLoading={false}
+              isReadOnly={false}
+              headers={["Alert message", "ID", "Start-End", "Last modified"]}
+              addSelectColumn={true}
```
panentheos, line 110: nit: If you omit the value in JSX, it passes `true`, so this line could just be:
```suggestion
              addSelectColumn
```

PR #610, opened 2025-03-04
ci: Updated action versions

deanshi:
Copying from https://github.com/mbta/screens-config-lib/pull/52 (insert Write That Down GIF here)

but more or less updating action versions from failing CI runs


digitalcora (approved):
📑📋

PR #611, opened 2025-03-04
Renamed Pre-Fare Duo to just be Pre-Fare

deanshi:
**Asana task**: [Fix announcement priority for migrated SCUs](https://app.asana.com/0/1185117109217413/1209272297071108)

Description
- Allow non-PA messages to take priority of 2
- Keep display numbers the same

### DIFF
--- a/assets/js/components/Dashboard/AlertsPage.tsx
+++ b/assets/js/components/Dashboard/AlertsPage.tsx
@@ -205,7 +205,7 @@ const AlertsList: ComponentType<AlertsListProps> = ({
 
       return screensWithAlert
         ? screensWithAlert.find((screen_id) =>
-            ids.includes(screenMetaData?.[screen_id]?.type),
+            ids.includes(screenMetaData[screen_id].type),
           )
         : false;
     });
--- a/assets/js/components/Dashboard/AlertsPage.tsx
+++ b/assets/js/components/Dashboard/AlertsPage.tsx
@@ -205,7 +205,7 @@ const AlertsList: ComponentType<AlertsListProps> = ({
 
       return screensWithAlert
         ? screensWithAlert.find((screen_id) =>
-            ids.includes(screenMetaData[screen_id].type),
+            ids.includes(screenMetaData?.[screen_id]?.type),
           )
         : false;
     });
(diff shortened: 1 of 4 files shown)


assets/js/components/Dashboard/AlertsPage.tsx
```diff
@@ -205,7 +205,7 @@ const AlertsList: ComponentType<AlertsListProps> = ({
 
       return screensWithAlert
         ? screensWithAlert.find((screen_id) =>
-            ids.includes(screenMetaData[screen_id].type),
+            ids.includes(screenMetaData?.[screen_id]?.type),
```
deanshi, line 208: When I looked a little deeper into this it seemed a little worrisome, I fixed it here so it doesn't straight up crash but it seems like some of the `screen_ids` might be missing. 

I don't have the full context of the underlying cause of this, but I can take a deeper peek if we feel it's necessary and it's not a known issue.

assets/js/constants/constants.ts
```diff
@@ -49,7 +49,7 @@ export const SCREEN_TYPES: { label: string; ids: string[] }[] = [
   { label: "E-Ink: Green Line", ids: ["gl_eink_v2"] },
   { label: "Elevator", ids: ["elevator_v2"] },
   { label: "PA ESS", ids: ["pa_ess"] },
-  { label: "Pre Fare Duo", ids: ["pre_fare_v2"] },
+  { label: "Pre Fare", ids: ["pre_fare_v2"] },
```
digitalcora, line 52: nit: we pretty consistently refer to these as "Pre-Fare" screens (with hyphen) in other contexts, so it might make sense to use that here too.

assets/js/components/Dashboard/AlertsPage.tsx
digitalcora: Was this related to the name change in some way? Did you run into the crash you mentioned while testing that change locally? Curious why TypeScript wouldn't have caught this.

deanshi: Yeah let me double check - I just checked `main` and it seemed like it was a problem, but dev/prod look fine. It might be my own outdated/incorrect local data. 

That being said, I'll revert this.

digitalcora (approved):
🖊️ ✅

PR #612, opened 2025-03-04
Change PA announcement priority from 1, 2, 3, 4 to 1, 3, 4, 5

deanshi:
**Asana task**: [Fix announcement priority for migrated SCUs](https://app.asana.com/0/1185117109217413/1209272297071108)

- Allow non-PA messages to take priority of 2, seems like we already do that in [realtime signs](https://github.com/mbta/realtime_signs/blob/5a5cbb9fea548cd5ab3cd69777ac74fc35432b60/lib/signs/utilities/audio.ex#L187)
- Keep display numbers the same, effectively everything will remain the same but if you look at the API calls underneath to `pa-messages`, priority will now be adjusted when you pick 2, 3, or 4. 

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -72,7 +72,7 @@ defmodule Random do
   end
 
   def priority do
-    Enum.random(1..4)
+    Enum.random([1, 3, 4, 5])
   end
 
   def text do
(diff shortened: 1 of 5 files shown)


priv/repo/seeds.exs
```diff
@@ -72,7 +72,7 @@ defmodule Random do
   end
 
   def priority do
-    Enum.random(1..4)
+    Enum.random(1..5)
```
robbie-sundstrom, line 75: This should exclude the number 2. Maybe just `Enum.random([1,3,4,5])`

robbie-sundstrom (approved):
nice 🚀

PR #615, opened 2025-03-20
Added data model and API for Prediction Suppression

deanshi:
**Asana task**: [Prediction suppression: Backend data model and API](https://app.asana.com/0/1185117109217413/1209637020443922/f)

- Added suppressed predictions migration for creating table
- Added `suppressed_prediction.ex` DataModel containing `location_id`, `direction_id` and `clear_at_end_of_day` boolean
- Added `suppressed_predictions.ex` Ecto.Repo bridge, with create, update, delete and get/get all functions
- Added `suppressed_predictions_api_controller` handling the calls from `router.ex`
- ~Tests coming soon~

- [x] For features with a design/UX component, deployed branch to dev-green and let product know it's ready for review.

### DIFF
--- a/lib/screenplay/suppressed_predictions/suppressed_prediction.ex
+++ b/lib/screenplay/suppressed_predictions/suppressed_prediction.ex
@@ -1,0 +1,160 @@
+defmodule Screenplay.SuppressedPredictions.SuppressedPrediction do
+  @moduledoc """
+  Represents a Suppressed Prediction for a given location
+  """
+  alias Screenplay.Places
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @jfk_umass_ashmont_place "jfk_umass_ashmont_platform"
+  @jfk_umass_braintree_place "jfk_umass_braintree_platform"
+  @green_line_routes ["Green-B", "Green-C", "Green-D", "Green-E"]
+  @valid_silver_line_routes ["741", "742", "743", "746"]
+
+  @derive {Jason.Encoder, except: [:__meta__]}
+
+  @type t() :: %__MODULE__{
+          location_id: String.t(),
+          route_id: String.t(),
+          direction_id: integer(),
+          clear_at_end_of_day: boolean()
+        }
+
+  @primary_key false
+  schema "suppressed_predictions" do
+    field(:location_id, :string, primary_key: true)
+    field(:route_id, :string, primary_key: true)
+    field(:direction_id, :integer, primary_key: true)
+    field(:clear_at_end_of_day, :boolean)
+
+    timestamps(type: :utc_datetime)
+  end
+
+  def changeset(suppressed_prediction, attrs \\ %{}) do
+    places = Places.get()
+
+    suppressed_prediction
+    |> cast(attrs, [
+      :location_id,
+      :route_id,
+      :direction_id,
+      :clear_at_end_of_day
+    ])
+    |> validate_required([
+      :location_id,
+      :route_id,
+      :direction_id,
+      :clear_at_end_of_day
+    ])
+    |> validate_number(:direction_id,
+      less_than_or_equal_to: 1,
+      greater_than_or_equal_to: 0,
+      message: "Location Direction ID must be 0 or 1"
+    )
+    |> validate_edge_cases()
+    |> validate_location(places)
+    |> validate_routes(places)
+  end
+
+  defp validate_edge_cases(changeset) do
+    location_id = get_field(changeset, :location_id)
+    route_id = get_field(changeset, :route_id)
+
+    cond do
+      # JFK/UMass has the Ashmont/Braintree split, we ask for those stop ids rather than the parent id
+      location_id == "place-jfk" ->
+        add_error(
+          changeset,
+          :location_id,
+          "Please provide either the Ashmont or Braintree platform location_id: (#{@jfk_umass_ashmont_place} or #{@jfk_umass_braintree_place})"
+        )
+
+      # We don't handle separate Green Line and Silver Line branching route_ids
+      # That is populated when we pass data to Transit Data
+      # For internal use we just use "Green" or "Silver" to show all green/silver line routes
+      route_id in @green_line_routes ->
+        add_error(
+          changeset,
+          :route_id,
+          "Please provide just `Green` as the route_id for handling all Green Line routes"
+        )
+
+      true ->
+        changeset
+    end
+  end
+
+  defp validate_location(changeset, places) do
+    location_id = get_field(changeset, :location_id)
+
+    if places |> Enum.any?(&(&1.id == location_id)) do
+      changeset
+    else
+      add_error(changeset, :location_id, "Location `#{location_id}` does not exist")
+    end
+  end
+
+  defp validate_routes(changeset, places) do
+    unchecked_location_id = get_field(changeset, :location_id)
+    route_id = get_field(changeset, :route_id)
+
+    location_id =
+      if unchecked_location_id in [@jfk_umass_ashmont_place, @jfk_umass_braintree_place] do
+        "place-jfk"
+      else
+        unchecked_location_id
+      end
+
+    places
+    |> Enum.find(fn place -> place.id == location_id end)
+    |> case do
+      nil ->
+        add_error(
+          changeset,
+          :route_id,
+          "Location `#{location_id}` does not exist"
+        )
+
+      place ->
+        valid_route? =
+          case route_id do
+            "Silver" ->
+              # For Silver Line, we make sure that the routes are all valid routes in Screenplay
+              # In this case, 741, 742, 743 and 746 within @valid_silver_line_routes
+              Enum.any?(
+                place.screens,
+                fn
+                  %Screenplay.Places.Place.PaEssScreen{routes: routes} ->
+                    Enum.any?(
+                      routes,
+                      fn route -> route.id in @valid_silver_line_routes end
+                    )
+
+                  _ ->
+                    false
+                end
+              )
+
+            "Green" ->
+              # For Green Line, double check that a branch exists within the routes
+              Enum.any?(
+                place.routes,
+                &String.starts_with?(&1, "Green")
+              )
+
+            _ ->
+              route_id in place.routes
+          end
+
+        if valid_route? do
+          changeset
+        else
+          add_error(
+            changeset,
+            :route_id,
+            "Route ID is invalid"
+          )
+        end
+    end
+  end
+end
--- a/lib/screenplay_web/controllers/suppressed_predictions_api_controller.ex
+++ b/lib/screenplay_web/controllers/suppressed_predictions_api_controller.ex
@@ -1,0 +1,68 @@
+defmodule ScreenplayWeb.SuppressedPredictionsApiController do
+  use ScreenplayWeb, :controller
+
+  action_fallback ScreenplayWeb.FallbackController
+
+  alias Screenplay.SuppressedPredictions
+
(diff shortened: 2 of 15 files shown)


lib/screenplay/suppressed_predictions/suppressed_prediction.ex
```diff
@@ -0,0 +1,59 @@
...
+      greater_than_or_equal_to: 0,
+      message: "Location Direction ID must be 0, 1 or 2 for both directions"
+    )
+    |> validate_location()
+  end
+
+  defp validate_location(changeset) do
+    location_id = get_field(changeset, :location_id)
+
+    place_ids =
```
deanshi, line 48: Just double checking here - is `location_id` and `place.id` supposed to match here?

lib/screenplay_web/controllers/suppressed_predictions_api_controller.ex
```diff
@@ -0,0 +1,60 @@
...
+
+  alias Screenplay.SuppressedPredictions
+
+  def index(conn, _params) do
+    json(conn, SuppressedPredictions.get_all_suppressed_predictions())
+  end
+
+  def create(conn, params) do
+    with {:ok, suppressed_prediction} <-
+           SuppressedPredictions.create_suppressed_prediction(params) do
```
deanshi, line 14: I initially here was trying to handle both a single suppression, and a list of suppressions (for bulk handling). 

I think I ran into some `changeset` issues with `insert_all`, and decided to drop it but more than happy to re-add it in if we think we'll be bulk inserting a fair bit. Looking at the miro board, it looks like we don't bulk modify suppressions so it might be okay?

lib/screenplay_web/router.ex
```diff
@@ -102,6 +102,20 @@ defmodule ScreenplayWeb.Router do
...
+  # Prediction Suppression
+
+  scope "/api/suppressed-predictions", ScreenplayWeb do
+    pipe_through([:api, :authenticate])
+
+    get("/", SuppressedPredictionsApiController, only: [:index])
+  end
+
+  scope "/api/suppressed-predictions", ScreenplayWeb do
+    pipe_through([:api, :authenticate, :ensure_screens_admin])
```
deanshi, line 114: Is the `screens_admin` role okay here? Should we have a separate role for dealing with suppressions?

priv/repo/migrations/20250317165509_create_suppressed_predictions_table.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Screenplay.Repo.Migrations.AddSuppressedPredictions do
```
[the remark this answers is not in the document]
digitalcora: `ecto.create` only creates the database itself. Tables are created when a migration is run that creates them — which can occur either manually when you run `mix ecto.migrate`, or automatically when the app starts up since we start `Ecto.Migrator` in our supervision tree. Which migrations have been run on a given database ("up") and which have not ("down") is stored in an extra table in the database itself, so the migration process can bring "up" only the ones that are currently "down". You can [read more about that here](https://hexdocs.pm/ecto_sql/Ecto.Migration.html).

lib/screenplay_web/router.ex
digitalcora: "Screens admin" here means someone with the role needed to configure screens using the Pending Configuration feature. I think in this case the permissions should mirror those required by the PA Messages feature: only OIOs can write, everyone can read. For that feature, we created a `pa-message-admin` role — here we might consider creating a new `suppression-admin` role since we have a [specified list of people](https://app.asana.com/0/1209029294173264/1209369662784542) who should have access, which I think is slightly different from the PA message group. This would be an [infrastructure request](https://form.asana.com/?d=15492006741476&k=yxvoxR2nofLwyXkOpuwrBg). (I think everyone listed under "OIOs" already has a group they're a part of, and roles can reference that group, so we should do that instead of duplicating them individually — when new OIOs are onboarded or existing OIOs leave, we want the role membership to automatically reflect that.)

lib/screenplay_web/controllers/suppressed_predictions_api_controller.ex
digitalcora: `insert_all` (and `update_all`/`delete_all`) intentionally doesn't support changesets — it's a low-level tool for efficiently inserting large amounts of data, leaving casting/validating/etc. up to the user. I don't think we need to worry about it here since, as you pointed out, we're only ever going to be inserting/updating/deleting a single record at a time.

test/screenplay/suppressed_predictions_test.exs
```diff
@@ -0,0 +1,324 @@
...
+      })
+
+      assert %SuppressedPrediction{
+               id: 3,
+               location_id: "place-two",
+               direction_id: 0,
+               clear_at_end_of_day: true
+             } =
+               SuppressedPredictions.get_suppressed_prediction(3)
+    end
```
panentheos, line 55: There's a fair bit of test code here that seems like it's just testing basic Ecto operations, so there's not many ways for this to fail due to a bug in our app code. Given that we have controller tests that are exercising this same functionality but in a more integrated, "real" usage pattern, we could probably slim down these low-level tests without losing much confidence.

priv/repo/migrations/20250317165509_create_suppressed_predictions_table.exs
```diff
@@ -0,0 +1,15 @@
...
+  def change do
+    create table("suppressed_predictions") do
+      add :location_id, :string
+      add :direction_id, :integer
+      add :clear_at_end_of_day, :boolean
+
+      timestamps(type: :utc_datetime)
+    end
+
+    create index("suppressed_predictions", :location_id)
```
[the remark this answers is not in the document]
deanshi: Sounds good! Swapped it out to be a composite key with location_id and direction_id as the primary keys.

For our operations of create + delete, do we not need to handle update if the user wants to change the "clear_at_end_of_day" flag?

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
deanshi: For JFK is this the Ashmont/Braintree split within Places? 

I'll add northbound/southbound extras if that's the case.

deanshi: Also, in terms of filtering down the places, is there a good way of doing that? Should it be filtered via some param based on Bus/Subway or something along those lines?

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
```diff
@@ -0,0 +1,145 @@
...
+          if route_id == "Silver" do
+            pa_ess_routes =
+              place.screens
+              |> Enum.filter(fn
+                %Screenplay.Places.Place.PaEssScreen{} -> true
+                _ -> false
+              end)
+              |> Enum.flat_map(& &1.routes)
+
+            if pa_ess_routes |> Enum.any?(fn route -> route in @valid_silver_line_routes end) do
```
deanshi, line 124: Just wanted to verify the logic here: do we need to worry about places that might have stops including than the 741, 742, 743, 746 but have other routes involved?

Right now I'm saying if there's a stop with _any_ route that contains one of those, allow that stop to be used. Only error out when a stop has _only_ routes outside of the of those approved ones.

panentheos (commented):
This is looking great. Just a few more things we need to handle, and some suggestions for tuning up the code.

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
```diff
@@ -0,0 +1,145 @@
...
+      less_than_or_equal_to: 1,
+      greater_than_or_equal_to: 0,
+      message: "Location Direction ID must be 0 or 1"
+    )
+    |> validate_edge_cases()
+    |> validate_location(places)
+    |> validate_routes(places)
+  end
+
+  defp validate_edge_cases(changeset) do
```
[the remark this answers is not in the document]
digitalcora: To Brett's point, it's probably fine if we don't have these be real validation errors, but if there are specific "this should never happen" scenarios that the data model otherwise allows, it might be good to assert (using pattern-matching or explicit raising) that we never encounter them, to guard against mistakes elsewhere in the implementation.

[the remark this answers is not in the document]
deanshi: I think that's what I was trying to go for - I'll see if adding an explicit error out/fail somewhere makes sense, otherwise this might be okay as a "fake" verification.

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
```diff
@@ -0,0 +1,172 @@
...
+
+  defp validate_routes_with_location_id(changeset, places) do
+    location_id = get_field(changeset, :location_id)
+
+    if location_id in [@jfk_umass_ashmont_place, @jfk_umass_braintree_place] do
+      "place-jfk"
+    else
+      location_id
+    end
+    |> validate_routes(changeset, places)
```
panentheos, line 105: nit: Since this function is small and single-purpose, we could just fold it into `validate_routes` by assigning a variable to the result of the `if`, if that seems clearer.

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
```diff
@@ -0,0 +1,172 @@
...
+    |> case do
+      nil ->
+        add_error(
+          changeset,
+          :route_id,
+          "Location `#{location_id}` does not exist"
+        )
+
+      place ->
+        if route_id in ["Green" | place.routes] do
```
panentheos, line 122: I had to read this logic block a couple times to convince myself it was correct. The front-end code has a similar bit of logic, and I found it helpful to branch first on `route_id`, to split out the Silver and Green cases, and go from there. FYI, we don't need to (or want to, really) support specifying branches, e.g. "Green-D", since that won't be an option in the UI. Also, since we'll design the UI to avoid any validation errors, we can afford to be less specific and just roll things up into a generic error message, which should simplify things here.

deanshi: Sounds good!

Assuming the front end UI can clear up some of the route_id handling, I basically was able to strip out the logic to be only the Silver Line handling (Green Line will just pass through and the other routes will pass through.)

I'm hoping I understood what you're saying correctly, but let me know if there's something I missed with my new simplified validation.

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
deanshi: Moved it into the validate_routes, makes sense. 

I initially just wanted the function flow to be super explicit about it passing in a potentially transformed location_id but it might just be a tad overkill.

lib/screenplay/suppressed_predictions/suppressed_prediction.ex
```diff
@@ -0,0 +1,150 @@
...
+            else
+              add_error(
+                changeset,
+                :route_id,
+                "Place `#{place.id}` currently has only Silver Line routes that are not supported in Screenplay"
+              )
+            end
+
+          _ ->
+            changeset
```
panentheos, line 146: The silver case looks well handled, but this is a bit too permissive. If we add a case for Green that checks if any of the place's routes starts with "Green", and a default case that checks if the route_id is in the place's routes, then we should be good.

panentheos (approved):
Looks great, thanks for all your work on this!

PR #617, opened 2025-03-27
add data structure for prediction suppression page

panentheos:
**Asana task**: [Prediction suppression: Stops data structure to drive UI](https://app.asana.com/0/1185117109217413/1209637020443919/f)

Description

This adds a new "service record" data structure, which is computed from the API and provides a basis for looking up a particular station/route/direction to determine whether it's the start, end, or middle of its trip. This info will be used to drive the UI for the prediction suppression page.

### DIFF
--- a/assets/js/models/service_record.ts
+++ b/assets/js/models/service_record.ts
@@ -1,0 +1,6 @@
+export interface ServiceRecord {
+  route: string;
+  place_id: string;
+  direction_id: 0 | 1;
+  type: "start" | "end" | "mid";
+}
--- a/assets/js/models/service_record.ts
+++ b/assets/js/models/service_record.ts
@@ -1,6 +1,0 @@
-export interface ServiceRecord {
-  route: string;
-  place_id: string;
-  direction_id: 0 | 1;
-  type: "start" | "end" | "mid";
-}
--- a/lib/screenplay/prediction_suppression.ex
+++ b/lib/screenplay/prediction_suppression.ex
@@ -1,4 +1,7 @@
 defmodule Screenplay.PredictionSuppression do
+  @moduledoc """
+  Generates data structures for driving the prediction suppression UI
+  """
   use GenServer
 
   @spec service_records() :: [
--- a/lib/screenplay/prediction_suppression.ex
+++ b/lib/screenplay/prediction_suppression.ex
@@ -1,0 +1,83 @@
+defmodule Screenplay.PredictionSuppression do
+  use GenServer
+
+  @spec service_records() :: [
+          %{
+            place_id: String.t(),
+            route: String.t(),
+            direction_id: 0 | 1,
+            type: :start | :end | :mid
+          }
+        ]
+  def service_records() do
+    case :ets.lookup(:service_records, :value) do
+      [{:value, data}] -> data
+      _ -> []
+    end
+  end
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+  end
+
+  @impl GenServer
+  def init(_) do
+    :ets.new(:service_records, [:protected, :named_table, read_concurrency: true])
+    send(self(), :update)
+    {:ok, %{}}
+  end
+
+  @impl GenServer
+  def handle_info(:update, state) do
+    Process.send_after(self(), :update, 60_000)
+
+    case Screenplay.V3Api.get_json("/route_patterns", %{
+           "filter[route]" => "Red,Orange,Blue,Green-B,Green-C,Green-D,Green-E,741,742,743,746",
+           "include" => "representative_trip.stops"
+         }) do
+      {:ok, %{"data" => data, "included" => included}} ->
+        trip_lookup =
+          for %{"type" => "trip"} = trip <- included, into: %{} do
+            {trip["id"], trip}
+          end
+
+        stop_lookup =
+          for %{"type" => "stop"} = stop <- included, into: %{} do
+            {stop["id"], stop}
+          end
+
+        new_service_records =
+          for %{"attributes" => %{"typicality" => 1}} = route_pattern <- data,
+              trip_id = route_pattern["relationships"]["representative_trip"]["data"]["id"],
+              trip = trip_lookup[trip_id],
+              stop_references = trip["relationships"]["stops"]["data"],
+              %{"id" => stop_id} <- stop_references,
+              uniq: true do
+            stop = stop_lookup[stop_id]
+
+            %{
+              place_id: stop["relationships"]["parent_station"]["data"]["id"],
+              route: trip["relationships"]["route"]["data"]["id"] |> route(),
+              direction_id: trip["attributes"]["direction_id"],
+              type:
+                cond do
+                  List.first(stop_references)["id"] == stop_id -> :start
+                  List.last(stop_references)["id"] == stop_id -> :end
+                  true -> :mid
+                end
+            }
+          end
+
+        :ets.insert(:service_records, {:value, new_service_records})
+
+      _ ->
+        nil
+    end
+
+    {:noreply, state}
+  end
+
+  defp route("Green-" <> _), do: "Green"
+  defp route(route_id) when route_id in ["741", "742", "743", "746"], do: "Silver"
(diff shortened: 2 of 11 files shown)


assets/js/models/service_record.ts
```diff
@@ -0,0 +1,6 @@
+export interface ServiceRecord {
```
digitalcora, line 1: * The name "service record" sounds a bit generic/non-indicative to my ears — almost any kind of transit data we use is a "record" that describes "service", in some sense. CanonicalStop?
* Is this sufficient to build the full proposed UI? It seems like we're missing an identifier for which pattern(s) the stop is on, to enable drawing the line in a sensible order and with appropriate visual "branches".

assets/js/models/service_record.ts
```diff
@@ -0,0 +1,6 @@
...
+  route: string;
```
digitalcora, line 2: I'd hesitate to call this `route` because it doesn't map directly onto the GTFS concept of a route. It _does_ map pretty directly onto our GTFS concept of a `line`, though (the Green Line routes together are `line-Green`, 741+742+743+746 are `line-SLWaterfront`). It would be nice if we could actually fetch the data based on Lines rather than Routes, but that would add a bit more complexity since it's not possible to directly filter on Lines, and the frequency of changes to the line/route mapping for these specific routes should be very low.

lib/screenplay/prediction_suppression.ex
```diff
@@ -0,0 +1,86 @@
...
+            stop = stop_lookup[stop_id]
+
+            %{
+              place_id: stop["relationships"]["parent_station"]["data"]["id"],
+              route: trip["relationships"]["route"]["data"]["id"] |> route(),
+              direction_id: trip["attributes"]["direction_id"],
+              type:
+                cond do
+                  List.first(stop_references)["id"] == stop_id -> :start
+                  List.last(stop_references)["id"] == stop_id -> :end
```
digitalcora, line 68: This is slightly inefficient since we have to re-iterate the entirety of `stop_references` for (nearly) every `stop_id` in it. Maybe we could pull this out to another clause of the `for`:

```elixir
stop_references = trip["relationships"]["stops"]["data"],
first_stop_id = List.first(stop_references)["id"],
last_stop_id = List.last(stop_references)["id"],
```

And that makes the condition like this:

```elixir
case stop_id do
  ^first_stop_id -> :start
  ^last_stop_id -> :end
  _other -> :mid
end
```

assets/js/models/service_record.ts
```diff
@@ -0,0 +1,6 @@
...
+  place_id: string;
```
digitalcora, line 3: Is it consistent within Screenplay that we use "place" to refer to a parent station, or can it mean something else? Here it seems to exactly mean "station" (stop with location type 1).

assets/js/models/service_record.ts
panentheos: Re: CanonicalStop, sure that seems fine. I couldn't think of a "good" name, so anything that's reasonable enough and unambiguous should work.

Re: sufficient: From what I could tell, the line map decoration is already broken out into a component that has the necessary information to draw the branches. And we already have a mechanism for sorting stations into the correct order, since we do that on a couple other pages.

assets/js/models/service_record.ts
panentheos: Yeah, given the extremely low data size and update frequency I'm inclined to just leave these hardcoded. I'll rename the field to `line` though.

assets/js/models/service_record.ts
panentheos: The full places list includes bus stops as well. But I guess this could be `stop_id` or `station_id`.

assets/js/models/service_record.ts
digitalcora: I see, though I don't love that it means this UI will ultimately be constructed by combining

1. Pattern data from the API, but with ordering/branching information discarded
2. Hard-coded ordering/branching information on the client

...as compared to getting everything from a single source of truth. Of course, if we did the latter, less of the code could be shared with existing components, though I'd argue that would be better in the long run if we could eventually move _those_ components to also use a single source of truth instead of the hard-coded data.

lib/screenplay/prediction_suppression.ex
```diff
@@ -0,0 +1,86 @@
...
+            {trip["id"], trip}
+          end
+
+        stop_lookup =
+          for %{"type" => "stop"} = stop <- included, into: %{} do
+            {stop["id"], stop}
+          end
+
+        new_service_records =
+          for %{"attributes" => %{"typicality" => 1}} = route_pattern <- data,
```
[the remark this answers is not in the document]
digitalcora: Ah, right, bus routes don't have canonical patterns defined... maybe we could _prefer_ `is_canonical` patterns but if none are available, use the most-typical ones?

assets/js/models/service_record.ts
panentheos: I agree it would be ideal if it could all get pulled from the API. The line map is split out into its own task, so I didn't do a comprehensive analysis of it, but when we get to implementing that we could decide whether to use what's there, augment this data structure, or do something else as appropriate.

assets/js/models/service_record.ts
digitalcora: Oh, gotcha, so it's "stop with no parent station" (a.k.a. either a parent station or standalone stop... it's annoying there's no more concise terminology for that). But yeah I think it's fine as `place_id`.

assets/js/models/service_record.ts
digitalcora: Sounds okay. It's just a little awkward that within the scope of this change, all of this is speculative code — nothing uses it yet, so it's not obvious that it's a good fit for whatever it will eventually be used for.

digitalcora: On the naming question, if we're actually not universally building these from `canonical` service, maybe the name `CanonicalStop` isn't fitting. `LineStop`? (like, a stop on a line) I'm not coming up with very good names at the moment it seems...

lib/screenplay/prediction_suppression.ex
panentheos: Thinking out loud: I made the requested change but I'd be surprised if this wasn't optimized by the compiler. One of the big benefits of immutable data structures is that operations can be cached/rearranged safely, so there are probably a lot of little optimizations like that under the hood.

lib/screenplay/prediction_suppression.ex
```diff
@@ -0,0 +1,96 @@
+defmodule Screenplay.PredictionSuppression do
+  @moduledoc """
+  Generates data structures for driving the prediction suppression UI
+  """
+  use GenServer
+
+  defguardp is_silver(route_id) when route_id in ["741", "742", "743", "746"]
```
digitalcora, line 7: `is_slw` or `is_sl_waterfront` might be a better name for this — it does not include the SL4 and SL5, which are also "the Silver Line". (alternatively we could actually include the SL4 and SL5, since it would make no difference to this logic)

lib/screenplay/prediction_suppression.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  @impl GenServer
+  def init(_) do
+    :ets.new(:line_stops, [:protected, :named_table, read_concurrency: true])
+    send(self(), :update)
+    {:ok, %{}}
+  end
+
+  @impl GenServer
+  def handle_info(:update, state) do
```
digitalcora, line 36: A simple test case for this behavior would be nice to have, especially since it has no "real" users at the moment. We could stub a cut-down version of the API response (which still includes relevant edge cases), run the update, and assert that `line_stops` returns what the data should have been parsed into.

digitalcora (approved):

lib/screenplay/prediction_suppression.ex
panentheos: I wanted to do this, but surprisingly the infrastructure for mocking HTTP responses isn't quite set up to make this straightforward. I may push that to a followup PR since it'll touch a bunch of things that aren't related. Also, the actual consumer of this data will be coming right after this.

PR #619, opened 2025-03-31
Added endpoint for Transit Data

deanshi:
**Asana task**: [Prediction Suppression: Transit Data API](https://app.asana.com/0/1185117109217413/1209646154006769/f)

- /api/suppressed_predictions/all will now return a list of suppressed predictions for transit data use
- Takes our current suppressed_predictions data and transforms them within the endpoint to get the correct format
- Added a few new test signs for verifying Green and Silver line logic

### DIFF
--- a/docs/tech_specs/suppressed_predictions_api.md
+++ b/docs/tech_specs/suppressed_predictions_api.md
@@ -1,0 +1,116 @@
+# List Suppressed suppression_type
+
+Lists all Suppressed suppression_type defined by the stop_id, route_id and direction_id
+
+_Note_: JFK/UMass (place-jfk) will use the child stop_ids.
+
+**URL** : `/api/suppressed_suppression_type/suppression_data`
+
+**Method** : `GET`
+
+**Parameters**: None
+
+**API key required** : YES
+
+## Success Responses
+
+**Code** : `200 OK`
+
+**Response**:
+
+```json
+[
+  {
+    "stop_id": "place-aqucl",
+    "route_id": "Blue",
+    "direction_id": 1,
+    "suppression_type": "none"
+  },
+  {
+    "stop_id": "place-aqucl",
+    "route_id": "Blue",
+    "direction_id": 0,
+    "suppression_type": "none"
+  },
+  {
+    "stop_id": "place-ogmnl",
+    "route_id": "Orange",
+    "direction_id": 0,
+    "suppression_type": "terminal"
+  },
+  {
+    "stop_id": "place-hymnl",
+    "route_id": "Green-B",
+    "direction_id": 1,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "place-hymnl",
+    "route_id": "Green-C",
+    "direction_id": 1,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "place-hymnl",
+    "route_id": "Green-D",
+    "direction_id": 1,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "place-crtst",
+    "route_id": "741",
+    "direction_id": 0,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "place-crtst",
+    "route_id": "742",
+    "direction_id": 0,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "place-crtst",
+    "route_id": "743",
+    "direction_id": 0,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "place-crtst",
+    "route_id": "746",
+    "direction_id": 0,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "70085",
+    "route_id": "Red",
+    "direction_id": 0,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "70086",
+    "route_id": "Red",
+    "direction_id": 1,
+    "suppression_type": "none"
+  },
+  {
+    "stop_id": "70095",
+    "route_id": "Red",
+    "direction_id": 0,
+    "suppression_type": "stop"
+  },
+  {
+    "stop_id": "70096",
+    "route_id": "Red",
+    "direction_id": 1,
+    "suppression_type": "none"
+  }
+]
+```
+
+## Failure Responses
+
+**Code** : `403 Forbidden`
+
+**Response**:
+
+`Invalid API key`
--- a/lib/screenplay/suppressed_predictions.ex
+++ b/lib/screenplay/suppressed_predictions.ex
@@ -2,6 +2,10 @@ defmodule Screenplay.SuppressedPredictions do
   @moduledoc """
   Module for functions dealing with `SuppressedPredictions` and the `suppressed_predictions` database 
   """
+
+  alias Screenplay.Places
+  alias Screenplay.PredictionSuppression
+  alias Screenplay.PredictionSuppressionUtils
   alias Screenplay.Repo
   alias Screenplay.SuppressedPredictions.SuppressedPrediction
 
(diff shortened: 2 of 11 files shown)


lib/screenplay/suppressed_predictions.ex
```diff
@@ -38,6 +41,76 @@ defmodule Screenplay.SuppressedPredictions do
     |> Repo.all()
   end
 
+  @doc """
+  Get all suppressed predictions modified for transit data
+  """
+  @spec get_all_suppressed_predictions_transit_data() :: [SuppressedPrediction.t()]
+  def get_all_suppressed_predictions_transit_data do
```
panentheos, line 48: nit: "Transit data" is an organizational detail, so probably better to avoid encoding it in the name

lib/screenplay/suppressed_predictions.ex
```diff
@@ -38,6 +41,76 @@ defmodule Screenplay.SuppressedPredictions do
...
+      } ->
+        # Go through Green Line route and repopulate the appropriate branches with the unique direction id
+        Enum.find(places, fn place -> place.id == location_id end).routes
+        |> Enum.filter(fn route -> String.contains?(route, "Green") end)
+        |> Enum.map(fn route ->
+          if String.contains?(route, "Green") do
+            %{
+              stop_id: location_id,
+              route_id: route,
+              direction_id: direction_id
```
panentheos, line 67: The current data feed from Signs UI generates all possible entries, and includes a flag specifying whether each one is suppressed. I think the goal there was to always make the payload the maximum size, so there wouldn't be any surprise performance issues (on either end) if OIOs suddenly suppressed a lot of things. I think we probably want to keep that approach for this new API.

lib/screenplay/suppressed_predictions.ex
```diff
@@ -38,6 +41,76 @@ defmodule Screenplay.SuppressedPredictions do
...
+        route_id: route_id,
+        location_id: location_id,
+        direction_id: direction_id
+      } ->
+        [
+          %{
+            stop_id: location_id,
+            route_id: route_id,
+            direction_id: direction_id
+          }
```
panentheos, line 108: There are actually two kinds of suppression flags here: stop-based, and terminal. We're kind of rolling them up in the database because each stop/line/direction can only be one or the other, but the handling of those by transit data will probably be quite different, so I'm guessing they'll want them split out somehow. Two options that come to mind: we could generate two separate lists, or we could specify a new value for the "flag" field mentioned above, which designates terminal suppression. We might want to check with transit data and see what would work best for them.

As for separating them out on our end, the data structure PR that's about to be merged should make that straightforward.

docs/tech_specs/suppressed_predictions_api.md
```diff
@@ -0,0 +1,115 @@
...
+    "stop_id": "place-aqucl",
+    "route_id": "Blue",
+    "direction_id": 1,
+    "predictions": "normal"
+  },
+  {
+    "stop_id": "place-aqucl",
+    "route_id": "Blue",
+    "direction_id": 0,
+    "predictions": "normal"
```
digitalcora, line 33: What is `normal`? It doesn't appear we actually generate this. It _does_ appear we can generate `null` for this field though — is that intentional? What would this mean?

deanshi: Sorry - jumped the gun slightly on this, I'm waiting for a response back from Transit Data to see if `null` is fine, and change the code accordingly.

lib/screenplay/suppressed_predictions.ex
```diff
@@ -37,6 +40,73 @@ defmodule Screenplay.SuppressedPredictions do
...
+                  suppressed_predictions,
+                  child_stop_id
+                )
+            }
+          end
+        )
+
+      %Screenplay.Places.Place{id: place_stop_id, screens: screens} ->
+        Enum.flat_map(screens, fn
+          %Screenplay.Places.Place.PaEssScreen{routes: routes} ->
```
panentheos, line 82: We shouldn't be limiting this to places with PA/ESS, with the exception of the SL which looks at only the stations currently shown in Signs UI. We can probably extract the logic in the SuppressedPrediction validation code for filtering the places list down to the ones we care about, since we want to use the same list here.

lib/screenplay/suppressed_predictions.ex
```diff
@@ -37,6 +40,73 @@ defmodule Screenplay.SuppressedPredictions do
...
+            routes
+            |> Enum.uniq_by(fn route -> {route.id, route.direction_id} end)
+            |> Enum.filter(&PredictionSuppressionUtils.valid_route?(&1.id))
+            |> Enum.map(fn route ->
```
panentheos, line 86: General note: Check out the special `for` syntax, which can be an elegant way of expressing chains of `flat_map`, `map`, `filter`, and `uniq`.

lib/screenplay/suppressed_predictions/prediction_suppression_utils.ex
```diff
@@ -0,0 +1,161 @@
...
+  def suppression_type(suppressed_predictions, location_id, route_id, direction_id) do
+    suppression_type_from_line_data(
+      suppressed_predictions,
+      location_id,
+      route_id,
+      direction_id
+    )
+  end
+
+  def suppression_type(suppressed_predictions, stop_id)
```
panentheos, line 99: General note: While it's possible to create two functions with the same name and different parameter counts, I think it should generally be avoided, since it makes the code a little harder to read and understand.

lib/screenplay/suppressed_predictions/prediction_suppression_utils.ex
```diff
@@ -0,0 +1,161 @@
...
+      when stop_id in @jfk_umass_child_stop_ids do
+    case Enum.find(jfk_umass_child_stop_data(), fn %{stop_id: id} -> id == stop_id end) do
```
panentheos, line 101: The caller is mapping over this data structure, but then we do a linear lookup here to recover the record again, so we're doing unnecessary extra work when we already had the full record to begin with.

lib/screenplay/suppressed_predictions/prediction_suppression_utils.ex
```diff
@@ -0,0 +1,161 @@
...
+        location_id: @jfk_umass_braintree_location_id,
+        route_id: route_id,
+        direction_id: direction_id
+      } ->
+        suppression_type_from_line_data(
+          suppressed_predictions,
+          @jfk_umass_braintree_location_id,
+          route_id,
+          direction_id
+        )
```
panentheos, line 124: These clauses look identical, and I don't think we could ever get to the `nil` clause below, so I think this whole function could be collapsed and inlined into the caller.

lib/screenplay/suppressed_predictions/prediction_suppression_utils.ex
```diff
@@ -0,0 +1,161 @@
...
+
+  defp suppression_type_from_line_data(
+         suppressed_predictions,
+         location_id,
+         route_id,
+         direction_id
+       ) do
+    Enum.find(suppressed_predictions, fn prediction ->
+      prediction.route_id == route_id && prediction.location_id == location_id &&
+        prediction.direction_id == direction_id
```
panentheos, line 139: Since we're doing this lookup repeatedly, it would be good to avoid the linear `Enum.find` to make sure the performance stays reasonable. We could pre-create a `MapSet` and then use a simple `in` to do the lookups.

lib/screenplay/suppressed_predictions/prediction_suppression_utils.ex
```diff
@@ -0,0 +1,161 @@
...
+    end)
+    |> case do
+      nil ->
+        :none
+
+      _found_prediction ->
+        Enum.find(PredictionSuppression.line_stops(), fn line_stop ->
```
panentheos, line 146: Same as above, we could create a `Map` for doing faster lookups with this data.

lib/screenplay/suppressed_predictions/prediction_suppression_utils.ex
```diff
@@ -0,0 +1,161 @@
...
+          line_stop.line == route_id &&
+            (line_stop.stop_id == location_id ||
+               (line_stop.stop_id == "place-jfk" &&
+                  location_id in jfk_umass_child_location_ids())) &&
+            line_stop.direction_id == direction_id
+        end)
+        |> case do
+          %{type: :start} -> :terminal
+          %{type: :end} -> :terminal
```
panentheos, line 155: We only make terminal predictions at the beginning of lines, not the end. We shouldn't generate any records at all for these.

lib/screenplay/suppressed_predictions.ex
deanshi: Got it - in this particular case, is it still safe to be mapping to all of the `routes` that exist under `screens` and validating based on those? (will this just effectively filter out to the PaEss screens again?) Or should we be validating on the routes that exist on the top level place and do some logic checking for Green Line branches and Silver line?

deanshi: I guess my question kind of boils down to - if there's no PA/ESS screen, is there a good way for us to get the different routes/direction ids? Or do we have to pull that data from elsewhere?

panentheos: Yeah, the we can use the list of `routes` on the place itself for everything except Silver line. For SL, we'll need to dig into the PA/ESS signs, since the `routes` field just rolls everything up as `Silver`

deanshi: Got it - and for direction id, will it _always_ have a corresponding 0/1 direction except at terminals?

panentheos (approved):
This looks good. There are a few opportunities for tightening up the logic that we could look at separately if you're interested, but the algorithm looks right so let's move forward with it!

PR #620, opened 2025-04-03
implement prediction suppression page

panentheos:
**Asana task**: [Prediction suppression: Build basic UI](https://app.asana.com/0/1185117109217413/1209637020443925/f)

This partially implements the prediction suppression page. Includes the station list with filtering, UI elements in the right place, and much of the styling. State and interactions are NOT wired up yet.

The sidebar link is hidden by default, and can be enabled in development by setting the `SHOW_PREDICTION_SUPPRESSION` environment variable.

### DIFF
--- a/assets/css/prediction-suppression.module.scss
+++ b/assets/css/prediction-suppression.module.scss
@@ -1,0 +1,35 @@
+@import "./variables.scss";
+
+.tableRow {
+  tbody > & {
+    border-bottom: 1px solid #6c757d;
+    border-top: 1px solid #6c757d;
+    &:nth-child(odd) {
+      background-color: $bg-elevation-3;
+    }
+    &:nth-child(even) {
+      background-color: #171f26;
+    }
+  }
+}
+
+.suppressionLabel {
+  cursor: pointer;
+  border-radius: var(--bs-border-radius);
+  padding: 0.375rem 0.75rem;
+  display: inline-flex;
+  align-items: center;
+  color: $text-link-primary;
+  &:has(:checked) {
+    color: $text-invert;
+    background-color: $button-primary;
+  }
+  &:has(:focus-visible) {
+    box-shadow: 0 0 0 0.25rem rgba(86, 100, 111, 0.5);
+  }
+}
+
+.suppressionCheckbox {
+  width: 0;
+  opacity: 0;
+}
--- a/assets/js/components/Dashboard/PredictionSuppressionPage.tsx
+++ b/assets/js/components/Dashboard/PredictionSuppressionPage.tsx
@@ -1,0 +1,238 @@
+import { useScreenplayContext } from "Hooks/useScreenplayContext";
+import { RadioList } from "Components/RadioList";
+import React, { useState } from "react";
+import fp from "lodash/fp";
+import * as styles from "Styles/prediction-suppression.module.scss";
+import { Dot, XCircleFill } from "react-bootstrap-icons";
+import { Form } from "react-bootstrap";
+import { sortByStationOrder } from "../../util";
+import { Place } from "Models/place";
+
+const lookupKey = (placeId: string, directionId: number) =>
+  `${placeId}:${directionId}`;
+
+const directionNames = (line: string) => {
+  switch (line.split("-")[0]) {
+    case "Green":
+    case "Blue":
+      return ["Westbound", "Eastbound"];
+    case "Silver":
+      return ["Eastbound", "To South Station/Downtown"];
+    default:
+      return ["Southbound", "Northbound"];
+  }
+};
+
+const PredictionSuppressionPage = () => {
+  const [line, setLine] = useState<string>("Green");
+  const { places, lineStops } = useScreenplayContext();
+  const filteredPlaces = sortByStationOrder(
+    fp.filter(({ routes, screens }) => {
+      switch (line) {
+        case "Green":
+          return fp.any(fp.startsWith("Green"), routes);
+        case "Silver":
+          return (
+            fp.includes(line, routes) &&
+            fp.any(
+              (screen) =>
+                screen.type === "pa_ess" &&
+                fp.any(
+                  ({ id }) => fp.includes(id, ["741", "742", "743", "746"]),
+                  screen.routes,
+                ),
+              screens,
+            )
+          );
+        default:
+          return fp.includes(line, routes);
+      }
+    }, places),
+    line,
+  );
+  const lineStopLookup = fp.flow([
+    fp.filter({ line: line.split("-")[0] }),
+    fp.map(({ stop_id, direction_id, type }) => [
+      lookupKey(stop_id, direction_id),
+      type,
+    ]),
+    fp.fromPairs,
+  ])(lineStops);
+
+  const noPredictions = (place: Place) => {
+    const serviceType = lineStopLookup[lookupKey(place.id, 0)];
+    return (
+      line.startsWith("Green") && fp.includes(serviceType, ["start", "end"])
+    );
+  };
+
+  const renderInput = (place: Place, directionId: number) => {
+    const serviceType = lineStopLookup[lookupKey(place.id, directionId)];
+    const suppressed = false;
+    const expire = false;
+    if (serviceType === "start" || serviceType === "mid") {
+      const predictionsText =
+        serviceType == "start" ? "terminal predictions" : "predictions";
+      return (
+        <>
+          <label className={styles.suppressionLabel}>
+            <input
+              type="checkbox"
+              checked={suppressed}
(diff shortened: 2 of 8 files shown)


assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+  ])(lineStops);
+
+  const noPredictions = (place: Place) => {
+    const serviceType = lineStopLookup[lookupKey(place.id, 0)];
+    return (
+      line.startsWith("Green") && fp.includes(serviceType, ["start", "end"])
+    );
+  };
+
+  const renderInput = (place: Place, directionId: number) => {
```
digitalcora, line 69: Reading through `PredictionSuppressionPage` it's sometimes not apparent to what extent the functions defined here are independent of the component's state and what data from the state they may depend on — but a lot of this logic seems "extractable" into pure functions or components defined outside of `PredictionSuppressionPage`, which accept as arguments/props the small amount of outside data they need. I'm sure you have better context on what the dependencies are but at a glance I'd suggest the filter implementation used to build `filteredPlaces`, `noPredictions`, and maybe `lineStopLookup` (does this need to depend on the state at all?) could be independent functions, and `renderInput`/`renderCell` could be independent components. What do you think?

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+      );
+    }
+    return renderInput(place, directionId);
+  };
+
+  return (
+    <div className="ms-5 mt-4">
+      <h1 className="mb-5">Suppress Predictions</h1>
+      <div className="d-flex gap-5">
+        <div style={{ flex: "0 0 200px" }}>
```
digitalcora, line 122: I'd like to hear the reasoning on using inline styles in this way — normally it's something I'd avoid. Does Bootstrap lack a utility that would do approximately this? Could we define our own, assuming this sidebar sizing is not purely a one-off?

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+                content: "Silver Line",
+              },
+            ]}
+          />
+        </div>
+        <table style={{ flex: "1" }}>
+          <colgroup>
+            <col style={{ width: 250 }} />
+            <col style={{ width: 300 }} />
+            <col style={{ width: 300 }} />
```
digitalcora, line 196: What are these inline widths doing? Looking at the rendered page, the table columns are certainly wider than 300px (as, I think, they should be!).

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+            </tr>
+          </thead>
+          <tbody>
+            {filteredPlaces.map((place) => (
+              <tr key={place.id} className={styles.tableRow}>
+                <td className="p-3">{place.name}</td>
+                {noPredictions(place) ? (
+                  <td colSpan={2} className="p-3">
+                    <em className="text-secondary">
+                      Predictions aren&apos;t generated for this location
```
digitalcora, line 218: When I replace this with `'` it still compiles and displays correctly (and in general I'd expect it to). The rationale for `eslint-plugin-react` rejecting this by default appears to be a [typographical one](https://github.com/jsx-eslint/eslint-plugin-react/issues/894) rather than a "real" limitation or pitfall of JSX — which is a bit silly given the suggested `&apos;` or `&#39;` are the same character as `'` so it's seemingly only requiring prose written in code to be less readable for no reason.

digitalcora (commented):
Pulled this down and tried it out, the functionality all looks good!

assets/css/prediction-suppression.module.scss
```diff
@@ -0,0 +1,35 @@
+@import "./variables.scss";
+
+.tableRow {
+  tbody > & {
```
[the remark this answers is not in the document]
panentheos: Good catch, that was left over from a previous iteration where I was trying to make this more general.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -0,0 +1,238 @@
+import { useScreenplayContext } from "Hooks/useScreenplayContext";
+import { RadioList } from "Components/RadioList";
+import React, { useState } from "react";
+import fp from "lodash/fp";
+import * as styles from "Styles/prediction-suppression.module.scss";
+import { Dot, XCircleFill } from "react-bootstrap-icons";
+import { Form } from "react-bootstrap";
+import { sortByStationOrder } from "../../util";
```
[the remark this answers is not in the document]
panentheos: Yep, let's do this in its own PR.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
panentheos: I started with that approach, but in practice all of these things are fairly tightly coupled, and are ultimately dependent on the state/context values (and that coupling will increase after the interactions are wired up). Breaking things out into top-level functions would mean having to pass all the dependencies and re-declare their types, which adds a fair bit of code that is mostly duplicative of information that's already here. Doing it this way, we get to leverage the lexical closure for most of the dependencies, but still break things up a bit.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
panentheos: There aren't generally helpers for specific pixel values like this, since the numbers could be anything. For the moment, this _is_ kind of a one-off, since the other pages with sidebars are all implemented a little differently, even though it seems like they're kind of aiming for 200px. That could very well be something we can standardize a bit at some point.

More broadly, I think inline styles have been overly maligned in the past, but they're only really a problem if you use them to introduce lots of duplication and coupling that becomes hard to manage. I worked on a large-ish project that used _only_ inline styles, and just leveraged React components to create abstractions and prevent duplication, and it worked fine. Utility frameworks like Tailwind and Bootstrap use tiny CSS classes, but the approach is thematically similar. For this particular case, we just need to apply `flex: 0 0 200px` to this div. I guess we could create a `.flex-0-0-200` helper, but that doesn't add much.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
panentheos: This is relying on the fact that table columns will expand proportionally to fill the space if the table is wider than the total. There may be a nicer way to do this, I'm a little rusty on table styling!

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
digitalcora: If you also don't see a clear reason for keeping it and dealing with the entity-escaping it requires, then yeah, I'd be in favor of either removing `'` from the rule or disabling it entirely.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
digitalcora: I think we should be able to use percentage widths if the goal is a certain proportion (30%, 35%, and 35% would be quite close to this). Also I don't know if it's necessary to define a `colgroup` at all — the widths could be defined on the header cells. (`col` seems intended for column-spanning situations?)

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
digitalcora: > the other pages with sidebars are all implemented a little differently, even though it seems like they're kind of aiming for 200px. That could very well be something we can standardize a bit at some point.

I'd go further and say this seems like a good opportunity to standardize — Mary has brought up on several occasions that we should be doing this proactively, to avoid accumulating small inconsistencies that then have to be written up as a separate task. Or if not right now, then at least proactively turning "at some point" into an extra task that someone can pick up.

> Utility frameworks like Tailwind and Bootstrap use tiny CSS classes, but the approach is thematically similar.

I suppose so! It just seems like a slippery slope that we'll have to be disciplined enough not to slide down (e.g. repeatedly adding "just one more property" to an inline style rather than extracting it into a descriptive class name).

panentheos: I created [a ticket](https://app.asana.com/0/1185117109217422/1209924206904036/f) to capture the desire to refactor the various page layouts to make them more consistent.

digitalcora (approved):
🚂

PR #621, opened 2025-04-03
add prediction suppression test

panentheos:
This sets up some infrastructure for mocking HTTP calls, and then adds a test for the new prediction suppression code.

### DIFF
--- a/lib/screenplay/alerts/screens_by_alert.ex
+++ b/lib/screenplay/alerts/screens_by_alert.ex
@@ -7,8 +7,10 @@ defmodule Screenplay.Alerts.ScreensByAlert do
   def get_screens_by_alert(alert_ids) do
     url = build_url(alert_ids)
 
+    http_client = Application.get_env(:screenplay, :http_client)
+
     with {:http_request, {:ok, response}} <-
-           {:http_request, HTTPoison.get(url)},
+           {:http_request, http_client.get(url)},
          {:response_success, %{status_code: 200, body: body}} <-
            {:response_success, response},
          {:parse, {:ok, parsed}} <- {:parse, Jason.decode(body)} do
--- a/test/screenplay/prediction_suppression_test.exs
+++ b/test/screenplay/prediction_suppression_test.exs
@@ -1,0 +1,121 @@
+defmodule Screenplay.PredictionSuppressionTest do
+  use ExUnit.Case
+
+  import Mox
+
+  @api_response %{
+    data: [
+      %{
+        "attributes" => %{"canonical" => true, "typicality" => 1},
+        "id" => "Blue-6-0",
+        "relationships" => %{
+          "representative_trip" => %{"data" => %{"id" => "canonical-Blue-C1-0", "type" => "trip"}},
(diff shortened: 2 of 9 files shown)


lib/screenplay/alerts/screens_by_alert.ex
```diff
@@ -7,8 +7,10 @@ defmodule Screenplay.Alerts.ScreensByAlert do
   def get_screens_by_alert(alert_ids) do
     url = build_url(alert_ids)
 
+    http_client = Application.get_env(:screenplay, :http_client)
```
digitalcora, line 10: Since this config is always set at compile time, this could be fetched using `compile_env`, e.g.

```elixir
@http Application.compile_env!(:screenplay, :http_client)

# ...

@http.get(url)
```

test/screenplay/prediction_suppression_test.exs
```diff
@@ -0,0 +1,121 @@
...
+
+  test "update" do
+    expect(HTTPoison.Mock, :get, fn _, _, _ ->
+      {:ok, %HTTPoison.Response{status_code: 200, body: Jason.encode!(@api_response)}}
+    end)
+
+    Screenplay.PredictionSuppression.init([])
+    Screenplay.PredictionSuppression.handle_info(:update, %{})
+
+    assert Screenplay.PredictionSuppression.line_stops() ==
```
digitalcora, line 111: nit: could use `alias Screenplay.PredictionSuppression` to reduce the repetition.

test/screenplay/prediction_suppression_test.exs
```diff
@@ -0,0 +1,121 @@
...
+        "relationships" => %{
+          "parent_station" => %{"data" => %{"id" => "place-wtcst", "type" => "stop"}}
+        },
+        "type" => "stop"
+      }
+    ]
+  }
+
+  test "update" do
+    expect(HTTPoison.Mock, :get, fn _, _, _ ->
```
digitalcora, line 104: When defining mocks like this I like to use pattern-matching to assert that the code is requesting the correct URL/path as well, though it maybe doesn't matter too much here since we are requesting a single fixed URL and there's no logic which changes that.

test/screenplay/prediction_suppression_test.exs
```diff
@@ -0,0 +1,121 @@
+defmodule Screenplay.PredictionSuppressionTest do
+  use ExUnit.Case
```
digitalcora, line 2: It feels possible for this to be `async: true` with perhaps some minor tweaks to the implementation — not always registering the ETS table under a fixed name, for example. Ideally with GenServers you can just start up your own "personal copy" for testing and it has no effect on the "main" one or anything other tests might be doing. In this case I don't really think it will slow anything down appreciably to have this one case be non-parallel, so take it with a grain of salt.

digitalcora (approved):

test/screenplay/prediction_suppression_test.exs
panentheos: Yeah, it would require making some non-trivial changes to the module just for the sake of this test, and test speed isn't anywhere close to a constraint right now.

PR #625, opened 2025-04-29
wire up prediction suppression UI

panentheos:
**Asana task**: [Prediction suppression: Wire up UI state](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209637020443931?focus=true)

This implements the stateful interactions on the prediction suppression page. Uses the SWR library to do "optimistic updates" for snappy interactions. The state lives in a top-level container because it will be needed on every page to drive the badge number on the left nav (coming in a followup PR)

### DIFF
--- a/assets/js/components/Dashboard/PredictionSuppressionPage.tsx
+++ b/assets/js/components/Dashboard/PredictionSuppressionPage.tsx
@@ -1,15 +1,25 @@
-import { useScreenplayContext } from "Hooks/useScreenplayContext";
+import {
+  usePredictionSuppressionState,
+  useScreenplayContext,
+} from "Hooks/useScreenplayContext";
 import { RadioList } from "Components/RadioList";
-import React, { useState } from "react";
+import React, { ReactNode, useEffect, useState } from "react";
 import fp from "lodash/fp";
 import * as styles from "Styles/prediction-suppression.module.scss";
 import { Dot, XCircleFill } from "react-bootstrap-icons";
 import { Form } from "react-bootstrap";
 import { sortByStationOrder } from "../../util";
 import { Place } from "Models/place";
+import {
+  createSuppressedPrediction,
+  deleteSuppressedPrediction,
+  updateSuppressedPrediction,
+} from "Utils/api";
+import { SuppressedPrediction } from "Models/suppressed_prediction";
+import { useUniqueId } from "Hooks/useUniqueId";
+import { useSearchParams } from "react-router-dom";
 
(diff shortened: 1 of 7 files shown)


assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -24,8 +34,18 @@ const directionNames = (line: string) => {
...
+  const [line, setLine] = useState<string>(params.get("line") || "Green");
   const { places, lineStops } = useScreenplayContext();
+  const { suppressedPredictions, mutateSuppressedPredictions } =
+    usePredictionSuppressionState();
+
+  useEffect(() => {
+    const newParams = new URLSearchParams();
+    newParams.set("line", line);
+    setParams(newParams);
+  }, [setParams, line]);
```
PaulJKim, line 47: I'm not sure I fully understand what's going on, but I think this is causing some weirdness with how this page handles browser history? But basically, after I select a different line to filter by, going back no longer works. It's almost like the url param is getting overwritten or something it seems.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
```diff
@@ -50,26 +70,93 @@ const PredictionSuppressionPage = () => {
...
+          await updateSuppressedPrediction(newRecord);
+          return newData;
+        },
+        { optimisticData: newData },
+      );
+    };
+
     if (serviceType === "start" || serviceType === "mid") {
       const predictionsText =
         serviceType == "start" ? "terminal predictions" : "predictions";
```
PaulJKim, line 162: Super nit since it wasn't even changed in this PR, but I noticed this check is using `==`. Should we use `===` instead to stay in-line with best practice?

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
panentheos: Hmm, strange. It looks like the PA messages page has this problem as well. I suspect it's related to the fact that the setter returned by `useSearchParams` "causes a navigation", according to the docs. I'll experiment with some alternate approaches.

assets/js/components/Dashboard/PredictionSuppressionPage.tsx
panentheos: Thanks, good catch. This is something we can have eslint enforce, so I'll get that set up.

PaulJKim (approved):
Looks good!

PR #627, opened 2025-05-06
replace reducers with regular state

panentheos:
This refactors the top-level application state to use regular state instead of reducers. Reducers tend to be more verbose, and they only provide extra value in a few narrow use cases (e.g. reducer composition and time-travel debugging), none of which we're using.

The revised code extends the passed Context values with setter callbacks to update the managed state. The result is much more compact and the names/types are more straightforward.

Also reworks the `AlertDetails` page to simplify the code by taking advantage of the new state container.


digitalcora (approved):
Incredible cleanup 🙌 This is far-ranging and I'm not as familiar with this codebase so perhaps it should get a second approval, but I could follow what's happening pretty easily.

PR #630, opened 2025-05-13
fix: URL in suppressed-predictions API documention

cmaddox5:
**Asana task**: ad-hoc

Hi friends! Working on consuming this new API endpoint and noticed a typo in the URL. Thought it might be easier to open a PR instead of DMing one of you on Slack about it.


deanshi (approved):
🤦 So sorry about this, thanks for the fix.

PR #631, opened 2025-05-13
use mbta asana workflow

panentheos:
**Asana task**: [Test task for GH integration](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210238619428086?focus=true)

This switches to the mbta version of the Asana integration, which has some additional capabilities and also ignores timeout errors when updating tickets (which seem to happen regularly)


digitalcora (approved):
Could also steal the [`screens` configuration](https://github.com/mbta/screens/blob/main/.github/workflows/asana.yml) which supports the auto-task-moving feature.

panentheos:
I chose to omit the auto-move stuff because it makes it difficult to do >1 PR per task, when appropriate.

digitalcora:
Good point — I've put a discussion item in for next engineering review so we can work out whether we generally want this or not.

PR #632, opened 2025-05-13
rework line map for prediction suppression page

panentheos:
**Asana task**: [Prediction suppression: Line diagram in stops lists](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209637020443928?focus=true)

This reworks the line diagram from the Places page so it can be used on the prediction suppression page. Notes:
* Segments are much longer, to accommodate the larger row heights on this page.
* The green line now shows branch letters for all stations, including trunk stations.


panentheos:
Good catch, fixed the places page. I'll handle adding Mattapan to prediction suppression in a separate PR, since that will probably require more extensive changes.

digitalcora (approved):
🔴 ✅

PR #633, opened 2025-05-13
chore: improve screen stop ID logic

digitalcora:
Reduce the number of special cases for specific screen types and derive stop IDs "generically" from Header, Footer, Alerts, or Departures widget configuration.

The remaining special cases are:

* DUPs, which have a unique configuration shape for departures
* Elevator screens, which need an API call to look up their stop ID


robbie-sundstrom (approved):
much easier to understand now! looks great!

PR #636, opened 2025-05-21
Pass along screen location from config

PaulJKim:
**Asana task**: [Allow editing/creation of screen label in Screens config](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209845314441909?focus=true)

Description
Removes supporting adding screen locations via the static `screen_locations.json` file.

This will be replaced by a new `location` field that can be set via the screens admin UI. The screenplay frontend already supports a `location` field and does with it what we want it to do so there are no frontend changes incurred.


digitalcora (commented):
This looks good, but I think we actually should include the new implementation of "locations" alongside it, and populate the field in the config before merging this. Reason being: if we deploy _just_ this change to production (and anything that's merged we should be able to freely deploy), it will remove all existing location labels, resulting in a worse experience until we deploy the "other half" of the change. So, better to roll them into one change?

PR #642, opened 2025-06-04
Update elixir to 1.17.3 and erlang to 27.3.4

deanshi:
**Asana task**: [Update screenplay erlang and elixir versions](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210451405410755?focus=true)

- Also update Alpine version
- Removed explict package adding and instead use erlang/alpine package
- Fixed tests by moving some objects in proper order

Description


robbie-sundstrom (approved):
Testing all looked good!

PR #644, opened 2025-06-10
enable prediction suppression

panentheos:
This removes the feature flag, enabling the prediction suppression page. Also includes a few requested tweaks for the Silver line, including adding route pills.


deanshi (approved):
:shipit:

PR #653, opened 2025-08-12
chore: switch to 4:00am service day boundary

digitalcora:
Due to the addition of limited late-night service, 4:00am should now be considered the start/end of the service day.

This affects:

* when PA messages are considered to be "active" when set to only play on certain days of the week
* the time filled in by the "start/end of service day" controls when a user is creating or editing a PA message

Incidentally reworks a function that determined the current service day to behave better around time zone changes. This used `DateTime.add`, which computes a time difference based on elapsed time, not wall time; on the date of the "fall back" from EDT to EST, shifting time backwards by 3 hours from e.g. 02:30 could land at 00:30 (still on the same day, rather than the previous day as intended).

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1211034769779824


robbie-sundstrom (approved):
LGTM!

PR #654, opened 2025-09-17
Fix redirecting to API endpoint rather than previous page

deanshi:
**Asana task**: [[bug] Screenplay login redirect to wrong URL](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210789861187901?focus=true)

Description
The idea here is that when we do hit an auth error, we will save the previous path from the previous request if it isn't an API request. We save it in the session and use that as the redirect after authorization. This keeps the old previous path as that is how we were forcing session reloads and continues to reassign previous path every time we go through the `:browser` pipeline. `previous_path_for_auth` though will be consumed as the redirect and removed. 


robbie-sundstrom (approved):

deanshi:
Going to merge this, tested this many times on dev-green and seems like it's working

PR #711, opened 2026-04-08
feat/chore: ETT: UI changes for expansion beyond Outfront screens

deanshi:
**Asana task**: [ETT: UI changes for expansion beyond Outfront screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213333802036638?focus=true)

Description
- Replaced and removed text in a few places referencing Outfront
- Removed ScreenIcon, references and scss associated with it
- Renamed Typescript, Elixir and some CSS classes to reflect the new change

The few places I didn't change the Outfront name include:
- the SFTP, Fake SFTP and SSH handling for Outfront specifically
- a bunch of the Outfront images and their uses across the board
- the `outfront_takeover_tool.md`
- the `outfront_takeover_tool_screens.exs` list of screens
- a few of the descriptions for clearing all images referencing Outfront Media screens

My thought here is that some of these descriptions/images can be updated as we're actually adding the new functionality, but if we want to preemptively remove them I can do that as well. 


assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
```diff
@@ -131,7 +131,6 @@ class OutfrontTakeoverTool extends React.Component<
             <div className="app-title">
               <img src="/images/t-identity.png" alt="Logo" className="logo" />
               <div className="stacked-title text-30">
-                <div>Outfront Media screens</div>
```
digitalcora, line 134: nit: some simplification of the markup/classes may be possible here now that this no longer needs to be a "stacked-title". (Alternatively perhaps we could keep the stacking but just change the wording to something else that fits the new scope of the tool. "In-station screens"? "Real-time info screens"?)

digitalcora (approved):
🚨 ✅

assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
deanshi: I do like the alternative text. Going to add `Real-time Info Screens`, since the extra context doesn't hurt and we have the space.

PR #714, opened 2026-04-24
include hidden signs in Places cache

panentheos:
**Asana task**: [Screenplay: track "hidden" screens for use with ETT](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370309386745?focus=true)

This adds a `hidden?` field to `ShowtimeScreen` structs, so they can be included in the Places cache, but still filtered out via the default code paths. ETT will be able to see these screens via a separate non-filtering getter.

### DIFF
--- a/lib/screenplay/places/builder.ex
+++ b/lib/screenplay/places/builder.ex
@@ -96,7 +96,13 @@ defmodule Screenplay.Places.Builder do
     |> Enum.group_by(&elem(&1, 0), fn
       {_stop_id,
        {id,
-        %Screen{app_id: app_id, disabled: disabled, app_params: app_params, location: location}}} ->
+        %Screen{
+          app_id: app_id,
+          disabled: disabled,
+          app_params: app_params,
+          location: location,
+          hidden_from_screenplay: hidden_from_screenplay
+        }}} ->
         direction_id =
           case app_params do
             %_app{direction_id: direction_id} -> direction_id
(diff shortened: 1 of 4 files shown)


lib/screenplay/places/builder.ex
```diff
@@ -108,13 +114,13 @@ defmodule Screenplay.Places.Builder do
...
           disabled: disabled,
           direction_id: direction_id,
-          location: location || ""
+          location: location || "",
+          hidden?: hidden_from_screenplay
         }
     end)
   end
 
   defp hidden_screen?({_id, %Screen{app_id: :on_bus_v2}}), do: true
```
digitalcora: This clause can also be removed, since there is no such screen type anymore, and with it the whole function.

test/screenplay/places/builder_test.exs
```diff
@@ -189,7 +189,7 @@ defmodule Screenplay.Places.BuilderTest do
                  routes: ["Blue"],
                  screens: []
                }
-             ] = PlacesCache.all(nil, return: :value)
+             ] = Screenplay.Places.get()
```
digitalcora, line 192: I think it's a bit odd now that this is testing a behavior of `Places.get/0`, but the test is located in the module for testing `Places.Builder`, and uses setup logic specific to that module. Is there a test module for `Places` itself we could move this specific case into? Then the rest could stay the same as they were before, as they are testing behaviors of `Places.Builder`.

panentheos: It was kind of already that way, since the behavior of this code is tightly coupled to the cache, and all these tests essentially had embedded a copy of `Screenplay.Places.get()`. We could split up the test logic, but then we'd lose the bigger picture, particularly the fact that this change is a no-op for existing behavior.

digitalcora: > we'd lose the bigger picture, particularly the fact that this change is a no-op for existing behavior

I agree it's nice to have it this way within the commit that makes the "include hidden screens" change. We could do any relevant test refactoring in a separate, second commit, though I don't think the way it is now is irredeemably confusing or anything.

digitalcora (approved):

PR #717, opened 2026-04-27
fix: Only fetch alerts in ETT on load and after 60s timer

robbie-sundstrom:
Noticed a bug when working on ETT work that would cause `active_alerts` and `past_alerts` APIs to both be repeatedly called. Both of their `useEffect` hooks depended on `fetchActiveAlerts` and `fetchPastAlerts`, which are recreated on every render. This caused the effects to run continuously. 

Removing these dependencies corrects this behavior, so the APIs are only called on initial load, when the 60 second timer expires, and alerts are cleared.

### DIFF
--- a/assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
+++ b/assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
@@ -17,58 +17,51 @@ interface AlertsListProps {
 const AlertsList = (props: AlertsListProps): JSX.Element => {
   const [alertsData, setAlertsData] = useState([]);
   const [pastAlertsData, setPastAlertsData] = useState([]);
-  const [lastChangeTime, setLastChangeTime] = useState(Date.now());
 
-  const fetchActiveAlerts = withErrorHandling(
-    async () => {
-      const response = await fetch(`${BASE_URL}/active_alerts`);
-      if (!response.ok) {
-        throw response;
-      }
-      return response.json();
-    },
-    { customMessage: "Failed to load active alerts. Please refresh the page." },
-  );
+  const refreshAlerts = async () => {
+    // Refresh active and past alerts on page load, alert clear, and at a regular interval.
+    const fetchActiveAlerts = withErrorHandling(
+      async () => {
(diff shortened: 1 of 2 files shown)


digitalcora:
I think the lint warnings here are pointing to the idea that [We Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect). In particular it seems here like we're basically using effects as an indirect version of just calling a function.

```tsx
useEffect(() => {
  const loadActiveAlerts = async () => {
    const data = await fetchActiveAlerts();
    if (data) setAlertsData(data);
  };
  loadActiveAlerts();
}, [lastChangeTime]);

setLastChangeTime(Date.now());
```

is roughly equivalent to


```tsx
const loadActiveAlerts = async () => {
  const data = await fetchActiveAlerts();
  if (data) setAlertsData(data);
};

loadActiveAlerts();
```

robbie-sundstrom:
@digitalcora good catch and suggestion! Didn't see that this caused the Linter warnings. The change I made now looks more complex than it is, since it essentially creates a `refreshAlerts` function that is now just called on page load, within a needed `useEffect` for the refresh timer, and when alerts are cleared.

assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
```diff
@@ -17,59 +17,51 @@ interface AlertsListProps {
...
-      const data = await fetchActiveAlerts();
-      if (data) {
-        setAlertsData(data);
-      }
-    };
-    loadActiveAlerts();
-  }, [fetchActiveAlerts, lastChangeTime]);
+  const refreshAlerts = async () => {
+    // Refresh active and past alerts on page load, alert clear, and at a regular interval.
+    const fetchActiveAlerts = withErrorHandling(
```
digitalcora: `fetchActiveAlerts` and `fetchPastAlerts` don't seem like they need to be defined within the component itself. Moving them out to the top level of the file would mean we don't have to re-construct them on every render of the component, and perhaps make the component code easier to understand as it's no longer concerned with those details directly.

digitalcora (approved):
One minor refactoring suggestion.

PR #718, opened 2026-04-28
separate indoor and outdoor content in emergency takeover messages

panentheos:
**Asana task**: [ETT: support indoor/outdoor distinction and new content](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370222596880?focus=true)

This splits the emergency takeover messages into separate indoor and outdoor content. Notes:
* We're assuming the old messages will be deleted as part of this change, so we're not making any attempt at being backward compatible.
* Tweaked the UI interactions to eliminate a few edge cases.
* The preview during message creation includes the portrait version of both messages, since that was easy. The landscape version won't fit in the current layout without some additional rework. Only the indoor content rendering is shown in the list view, to avoid taking excessive space.

The PR is split into a few separate commits for easier reviewing.

### DIFF
--- a/assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
+++ b/assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
@@ -55,7 +55,7 @@ const MessageInput = ({
     <div className="step-body">
       <div className="info">
         <div>{label} text</div>
-        <div className="text-14">(144 character max)</div>
+        <div className="text-14">&nbsp;(144 character max)</div>
       </div>
       <div>
         <div
--- a/assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
+++ b/assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
@@ -1,17 +1,16 @@
 import React from "react";
 import CANNED_MESSAGES from "Constants/messages";
 import { charLimit } from "Constants/misc";
+import { Message } from "../EmergencyTakeoverTool";
+import cx from "classnames";
+import { getMessageString } from "../../../util";
 
 interface CreateMessageProps {
-  messageOption: string;
-  cannedMessage: string;
-  customMessage: string;
-  changeMessageOption: (event: any) => void;
-  changeCannedMessage: (event: any) => void;
-  changeCustomMessage: (event: any) => void;
+  value: Message;
+  onChange: (value: Message) => void;
 }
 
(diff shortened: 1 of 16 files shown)


assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
```diff
@@ -22,60 +28,85 @@ const CreateMessage = (props: CreateMessageProps) => {
...
+}: {
+  value: Message;
+  onChange: (value: Message) => void;
+  label: string;
+}) => {
+  return (
+    <div className="step-body">
+      <div className="info">
+        <div>{label} text</div>
+        <div className="text-14">(144 character max)</div>
```
robbie-sundstrom: Nit, but can we add a space before this? It bothered me in the old version too

assets/js/components/EmergencyTakeoverTool/AlertWizard/WizardSidebar.tsx
```diff
@@ -25,24 +25,25 @@ class WizardSidebar extends React.Component<WizardSidebarProps> {
...
+    return <SVGPreviews showText prefix={prefix} message={message.text} />;
   }
 
   render() {
     return (
       <div className="wizard-sidebar">
         <span className="preview-title text-16">Preview</span>
-        {this.getPreviewImage()}
+        {this.getPreviewImage(this.props.indoorMessage, "indoor")}
+        {this.getPreviewImage(this.props.outdoorMessage, "outdoor")}
```
robbie-sundstrom, line 46: Based on the [Miro here](https://miro.com/app/board/uXjVJOrKdEk=/?moveToWidget=3458764653556348232&cot=10), do we instead want to have the Preview Images toggleable between the Indoor and Outdoor states when they differ? I noticed that showing two vertical previews made it harder to differentiate and validate the indoor, outdoor, and station list

digitalcora: Just noting that the part of the Miro you linked to is in the "Proposed Workflow" section, which is distinct from the "Workflow with Minimal UI changes" section that is the actual guide to implementation we're following.

panentheos: Wow, I didn't even realize there was more stuff "up there" in the Miro. But yes, the intention here was to make minimal UX changes for now.

robbie-sundstrom (approved):
Changes look good! Tested a few different things here and I think it's all working as intended

PR #719, opened 2026-04-29
refactor: Add Outfront ETT Screens to Places framework

robbie-sundstrom:
**Asana task**: [ETT: integrate Outfront screens with Places framework](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370222596876?focus=true)

- No user-visible changes (aside from ASHMONT and BRAINTREE no longer being capitalized on the Station Selection list
- Adds parent station ID to the hardcoded JSON of emergency-takeover-compatible Outfront screens
- Adds a new screen type in the Places framework that represents these Outfront ETT screens. These are filtered out in all cases except when finding the ETT compatible screens.

### DIFF
--- a/assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
+++ b/assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
@@ -3,6 +3,7 @@ import AlertDashboard from "./AlertDashboard/AlertDashboard";
 import AlertWizard from "./AlertWizard/AlertWizard";
 import ConfirmationModal, { ModalDetails } from "./ConfirmationModal";
 import { BASE_URL } from "Constants/constants";
+import STATION_ORDER_BY_LINE from "Constants/stationOrder";
 
 export interface Station {
   name: string;
@@ -56,7 +57,6 @@ class EmergencyTakeoverTool extends React.Component<
   constructor(props = {}) {
     super(props);
     this.state = {
-      // TODO
       alertWizardOpen: false,
       alertData: null,
       modalOpen: false,
@@ -71,11 +71,20 @@ class EmergencyTakeoverTool extends React.Component<
       stationScreenOrientationList: null,
     };
 
-    fetch(`${BASE_URL}/stations_and_screen_orientations`)
+    fetch(`${BASE_URL}/stations_and_screens`)
       .then((response) => response.json())
-      .then((result) =>
-        this.setState({ stationScreenOrientationList: result }),
-      );
+      .then((stationsMap: { [stationName: string]: Station }) => {
+        // Group stations by line using STATION_ORDER_BY_LINE
+        const stationsByLine: StationsByLine = {};
+
+        Object.keys(STATION_ORDER_BY_LINE).forEach((line) => {
+          stationsByLine[line] = STATION_ORDER_BY_LINE[line]
+            .map((stationInfo) => stationsMap[stationInfo.name])
+            .filter((station) => station !== undefined);
+        });
+
+        this.setState({ stationScreenOrientationList: stationsByLine });
+      });
 
     this.toggleAlertWizard = this.toggleAlertWizard.bind(this);
     this.openModal = this.openModal.bind(this);
--- a/assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
+++ b/assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
@@ -74,13 +74,16 @@ class EmergencyTakeoverTool extends React.Component<
     fetch(`${BASE_URL}/stations_and_screens`)
       .then((response) => response.json())
       .then((stationsMap: { [stationName: string]: Station }) => {
-        // Group stations by line using STATION_ORDER_BY_LINE
+        // Group stations by line
         const stationsByLine: StationsByLine = {};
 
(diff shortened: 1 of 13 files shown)


assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
```diff
@@ -71,11 +71,20 @@ class EmergencyTakeoverTool extends React.Component<
...
-        this.setState({ stationScreenOrientationList: result }),
-      );
+      .then((stationsMap: { [stationName: string]: Station }) => {
+        // Group stations by line using STATION_ORDER_BY_LINE
+        const stationsByLine: StationsByLine = {};
+
+        Object.keys(STATION_ORDER_BY_LINE).forEach((line) => {
+          stationsByLine[line] = STATION_ORDER_BY_LINE[line]
+            .map((stationInfo) => stationsMap[stationInfo.name])
+            .filter((station) => station !== undefined);
```
rwaskiewicz: Non-blocking comment/food for thought for future us - it might be worth it in the future to enable `noUncheckedIndexAccess` in [our tsconfig](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) to help us call our reasons we might need these `filter()` calls in the future. But we have ~100 violations of that rule right now, so certainly out of scope for this PR/probably worth its own task if we want to evaluate its usefulness/fix violation

assets/js/util.ts
```diff
@@ -119,6 +119,12 @@ export const matchStation = (
   station: string,
   stationScreenOrientationList: StationsByLine,
 ) => {
+  console.log(
```
rwaskiewicz: Double checking - is leaving this `console.log` intentional?

config/outfront_takeover_screens.exs
```diff
@@ -262,72 +304,84 @@ config :screenplay,
     ],
     blue: [
       %{
+        place_id: "place-bowd",
```
rwaskiewicz: I was looking for a way to verify/validate these place IDs, and found a file in `screens` that has a different place_id - https://github.com/mbta/screens/blob/f2ce2007850eed99db974b0b5080af781f94a40b/lib/screens/stops/subway.ex#L24

_However_, I don't know for a fact that the keys in my linked file are supposed to match these place IDs in this one. Calling this out in the off chance I'm right and this should be "place-bomnl"

config/outfront_takeover_screens.exs
```diff
@@ -262,72 +304,84 @@ config :screenplay,
...
       },
       %{
+        place_id: "place-state",
         name: "State",
         portrait: true,
         landscape: false,
         sftp_dir_name: "006_XFER_ORANGE_BLUE_STATE"
       },
       %{
+        place_id: "place-aquar",
```
rwaskiewicz, line 328: Same comment as above, questioning whether this should be "place-aqucl". Feel free to dismiss/answer above in the first comment of this nature

lib/screenplay/places.ex
```diff
@@ -33,6 +34,16 @@ defmodule Screenplay.Places do
 
   @spec get() :: list(Place.t())
   def get do
+    Cache.all(nil, return: :value)
```
rwaskiewicz, line 37: It looks like we changed the implementation of `get` here, and effectively renamed the existing `get` fn to `get_all`. For my own understanding, why did we do it this way? Or is GitHub showing the diff here in a way other than what/how we should interpret the diff?

rwaskiewicz (approved):

config/outfront_takeover_screens.exs
robbie-sundstrom: You're right, this is incorrect! I'm doing a double pass now to make sure none of the others are incorrect as well.

A quick way that I just confirmed was by going to the dotcom page for Bowdoin station, which is indexed by the parent station ID: https://www.mbta.com/stops/place-bomnl

lib/screenplay/places.ex
robbie-sundstrom: This diff here is confusing - I added another filter to `get()` that now filters out the ETT screens. This `get` is used in multiple places throughout the app where we don't want the Outfront ETT screens shown. 

`get_all` is essentially just a fetch that's used only by the ETT, which is now the old `get` implementation since there's a new screen type we want to include. Don't love the name `get_all()` but it will include screens/places we later filter out for the ETT, so it's not quite accurate to call it something like `get_ett()` either

assets/js/components/EmergencyTakeoverTool/EmergencyTakeoverTool.tsx
robbie-sundstrom: Thanks for pointing out! With that many violations it's probably worth looking at separately. A lot of them likely are constrained in a way that we could confidently say they would never become a problem, but it does seem like a helpful rule to generally have enabled

digitalcora:
@robbie-sundstrom Following this PR, there are two files `config/outfront_ett_screens.exs` and `config/outfront_takeover_screens.exs`, with similar but different contents; only the latter appears to be used. Was this an incomplete move/rename during development?

robbie-sundstrom:
@digitalcora yes that was an accidental inclusion on a commit before merging! I was just planning to remove the unecessary file in the follow up work, and have it on that branch in a [separate commit here](https://github.com/mbta/screenplay/commit/00267a69811b16d2eb1ae059379b16ea598010c7).

PR #721, opened 2026-05-04
support linked indoor/outdoor messages in emergency takeover tool

panentheos:
**Asana task**: [ETT: support indoor/outdoor distinction and new content](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370222596880?focus=true)

This adds the new gif messages and reworks the UI and data model to accommodate the new usage patterns. Notes:
* The message format is changing, so we'll need to clear the alerts in S3 for each environment.
* The SFTP handler is slightly more complex, since we need to account for multiple file extensions, and clean up files proactively so there's only at most one at a time.


lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -8,32 +8,30 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
...
+      |> Map.take(["message", "stations", "duration"])
       |> Map.merge(%{"id" => alert.id})
 
     _ = UserActionLogger.log(user, :create_alert, params_to_log)
     :ok = State.add_alert(alert)
 
-    portrait_image_data = decode_png(pngs["indoor_portrait"])
-    landscape_image_data = decode_png(pngs["outdoor_landscape"])
+    portrait_image_data = decode_image(images["indoor_portrait"])
+    landscape_image_data = decode_image(images["outdoor_landscape"])
```
robbie-sundstrom, line 34: I know that for Outfront screens, indoor and outdoor can generally correspond to portrait and landcape, but I don't think we can extend that to other screen types. Ultimately, we will want to generate up to 4 variations when needed for all combinations of indoor/outdoor and portrait/landscape. I don't think this distinction is relevant with this change aside from naming of these variables, but pointing out since this max 2 variation assumption does seem a little baked in right now

assets/js/util.ts
```diff
@@ -143,11 +144,27 @@ export const formatTime = (date: Date) => {
   }).format(date);
 };
 
-export const getMessageString = (message: Message) => {
+export const getMessageImageUrl = (
+  message: CannedMessage,
+  where: "indoor" | "outdoor",
```
robbie-sundstrom: Nit, but this `where` wording feels confusing to me throughout this code. I don't think there's a great naming answer here, but wanted to throw out `location_type`?

priv/static/images/alerts/EmergencyEnded-landscape.gif
robbie-sundstrom, line 1: Could we either modify the names of these so they have the same prefix throughout, i.e. `Alert-EmergencyEnded-landscape`, or just move them into a specific directory like `priv/static/images/alerts`

assets/js/constants/messages.ts
```diff
@@ -1,9 +1,76 @@
 const CANNED_MESSAGES = [
-  "Attention: An emergency situation has been reported. Please leave station through nearest exit.",
-  "Attention: An emergency situation has been reported outside the station. Please remain in station until advised otherwise by MBTA or Emergency Personnel.",
-  "Attention: An emergency situation has been reported. Please follow instructions of MBTA or Emergency Personnel.",
-  "Attention: Emergency situation has ended. It is safe to exit station.",
-  "This is a test of our emergency messaging system. This is NOT an emergency. This is a test.",
```
robbie-sundstrom, line 6: The corresponding images for these old messages can be deleted. They all have the naming `Outfront-Alert...` in `priv/static/images`

robbie-sundstrom (commented):
Mainly small comments that I'm submitting while I test the flows of this a bit more thoroughly before approving. Overall looks great!

assets/js/util.ts
digitalcora: `location_type` would be overloading the same-named field on Stop resources. I personally think `where` is fine here, but if we wanted to be extremely clear about it, it could be `emergencyMessagingLocation` (as that's the name of the field which has these values in the configuration).

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
panentheos: This change does generate and send all four permutations in the `images` array. Right now we're still just pulling out the Outfront ones, but I imagine this code will get reworked when we support the other screens.

assets/js/util.ts
robbie-sundstrom: True I forgot that `location_type` would overlap. I'll leave this up to Brett's discretion since `emergencyMessagingLocation` is long and I don't think I love any option

panentheos: I was hoping to avoid a long identifier that would blow out a bunch of lines. Could use `location`... it would shadow `window.location` but that's probably not going to cause any confusion.

robbie-sundstrom (approved):

PR #724, opened 2026-05-14
refactor: Move ETT canned message data to BE, store in Context on FE

robbie-sundstrom:
**Asana task**: [ETT: support takeover of Showtime screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370222596888?focus=true)

### Summary
This is a refactor that I found made it much easier to handle creating emergency takeovers for showtime screens in the following ways:
1. We don't have to pass additional message data to the BE every time that we create/modify a canned message. The full canned message information is needed on the BE for creating these takeovers on showtime screens.
2. The FE needs less data for each canned message than the BE does, namely the `audio_asset_url`, so it made little sense to store it in JS and pass it to our BE.
3. This gives us more flexibility to make additional changes to canned alerts in the future

### General Notes
- Includes two miscellaneous fixes in separate commits.
- Adds `utils/emergencyMessages.ts` as a consolidation point for canned and custom related logic from a few different places
- Creates `CannedMessagesContext` for sharing the full set of canned messages where needed in the alert creation and viewing process


assets/js/components/EmergencyTakeoverTool/AlertWizard/AlertPreview.tsx
```diff
@@ -20,17 +19,16 @@ const AlertPreview = ({ message, location, empty }: AlertPreviewProps) => {
...
   }
   if (message.type === "canned") {
-    return (
-      <img
-        className="portrait-png"
-        src={getMessageImageUrl(message, location, "portrait")}
-        alt=""
-      />
-    );
+    const imageUrl = getMessageImageUrl(message, "indoor", "portrait");
```
panentheos: I think this should be:
```suggestion
    const imageUrl = getMessageImageUrl(message, location, "portrait");
```

assets/js/utils/emergencyMessages.ts
```diff
@@ -0,0 +1,84 @@
+import { withErrorHandling } from "./errorHandler";
+import { BASE_URL } from "../constants/constants";
+
+export interface CannedMessage {
+  id: number;
+  type: "canned";
+  name?: string;
+  text?: {
```
panentheos, line 8: I read through the code paths that use this and I think everything works correctly, but if we wanted to tighten things up a bit, we could slim this type down to just the id again, and pass the `cannedMessages` as an explicit parameter to `getMessageString` and `getMessageImageUrl`. This wouldn't need much boilerplate since it's easy to pull from the context, and it would isolate the lookup and the associated type uncertainty in a central place.

panentheos (approved):
Ah yeah, I wondered if we'd need to do this eventually. See a few comments, then looks good!

assets/js/utils/emergencyMessages.ts
robbie-sundstrom: This was actually how I initially had it structured, and I changed it to this structure while cleaning up the code for PR. I think it is tighter with your suggested way, but one thing that I don't like about the approach is that it requires accessing the context in AlertWizard, which is a class component. I've got this working by consuming the context using:
```
  static contextType = CannedMessagesContext;
  declare context: React.ContextType<typeof CannedMessagesContext>;
``` 
I think this tradeoff is fine since there isn't anything technically wrong with it, but I'm just used to only consuming Context within functional classes.

panentheos: Ah, I didn't think about that. Up to you which way you think is less complex overall!

robbie-sundstrom: I think I'm going to keep it as is with some small tweaks. The setup in `AlertWizard` ends up feeling more complex with the cannedMessages being passed around through context

PR #726, opened 2026-05-15
fix alert link

panentheos:
**Asana task**: [Screenplay Bug: wrong URL on Edit Alert button ](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214354571963285?focus=true)

This, along with the associated infra change, changes the alert link so it works regardless of permission.


digitalcora (approved):
🔗 ✅

PR #732, opened 2026-05-19
feat: Support Emergency Takeovers of Showtime Screens

robbie-sundstrom:
**Asana task**: [ETT: support takeover of Showtime screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370222596888?focus=true)

## Summary
- On changes to emergency messages (create, edit, delete) this now takes up to two additional actions
  - Uploads or deletes images in S3 for custom messages
  - Modifies the screens configs for selected stations
- Adds error handling to the `alert_controller.ex`. Errors are now shown in the `ErrorModal` within `AlertWizard` and `AlertsList`
- Makes Silver Line selectable as a column
- Adds new functionality within `permanent_config.ex` for modifying the `emergency_takeover` field of screen configs

## Testing 
- Note that testing the full takeover process on showtime screens does not fully work locally, since the screens app can't point to an image that we save locally in Screenplay. I considered saving this image in the `../screens/assets/` directory to make this possible, but this seemed overall messier. I would recommend testing the full end to end of this in dev-green if you want to see the full behavior.


assets/css/stacked-stations.scss
```diff
@@ -47,9 +47,6 @@
   input[type="checkbox"] {
     margin: 0 16px 0 0;
   }
-  input[name="silver"] {
```
rwaskiewicz, line 50: Double checking - are we removing this here because there's no `<input name={x}...`
where `x` can resolve to "silver"? If that's the case, how do we know that's possible? I ask because it looks like this stylesheet is imported by `app.scss`, which if I understand correctly, could apply to anywhere in the app - but please correct me if I'm wrong!

assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
```diff
@@ -38,102 +38,96 @@ const fetchPastAlerts = withErrorHandling(
...
+      const errorData = await response.json();
+      throw new Error(errorData.error || response.statusText);
+    }
+
+    return response.json();
+  },
+  { customMessage: "Failed to clear all alerts." },
+);
+
+const AlertsList = (props: AlertsListProps): JSX.Element => {
```
rwaskiewicz: Non-blocking (I see this is lifted/moved around code), but the top level `JSX` namespace is deprecated:
```ts
// react/index.d.ts
declare global {
    /**
     * @deprecated Use `React.JSX` instead of the global `JSX` namespace.
     */
    namespace JSX {
```

Moving forward, we should access `JSX.*` through the React namespace:

```suggestion
const AlertsList = (props: AlertsListProps): React.JSX.Element => {
```

_Technically_, `(React.)JSX.Element` is a `ReactElement<P = any, T = any>`, so there may be room to improve this type a little more by specifying `P` and `T`, but that's even more non-blocking for this PR

assets/js/components/EmergencyTakeoverTool/AlertWizard/AlertWizard.tsx
```diff
@@ -35,6 +37,34 @@ interface AlertWizardState {
   showErrorMessage: boolean;
 }
 
+const handleAlertSubmit = withErrorHandling(
+  async (data: any, id: string | null) => {
```
rwaskiewicz: Non-blocking - even though we're not pulling any of the fields our of `data` in this function, `any` is still a wide type to use here. Since we're serializing this, we could change the type of data as such to demonstrate the general shape of it:

```suggestion
  async (data: Record<string, unknown> id: string | null) => {
```

assets/js/components/EmergencyTakeoverTool/AlertWizard/AlertWizard.tsx
```diff
@@ -202,15 +232,16 @@ class AlertWizard extends React.Component<AlertWizardProps, AlertWizardState> {
...
-      "[name~=csrf-token][content]",
-    ) as HTMLMetaElement;
-    const csrfToken = csrfMetaElement.content;
-
-    const stations = this.state.selectedStations.map(({ name }) => name);
+    // Map station names to whether they have an outfront screen (portrait or landscape)
+    const stations = Object.fromEntries(
+      this.state.selectedStations.map(({ name, portrait, landscape }) => [
+        name,
+        portrait || landscape,
```
rwaskiewicz: For my own understanding, what's this do? I would have thought that a screen would have to be portrait or landscape, but this implies that there's a third (or more) states where a screen is neither portrait nor landscape 🤔 Or maybe I'm misunderstanding here...maybe it's possible for neither field to be set here, regardless of that it's real world positioning/orientation is?

assets/js/components/EmergencyTakeoverTool/AlertWizard/StationColumn.tsx
```diff
@@ -23,7 +23,7 @@ const StationColumn = (props: StationColumnProps): JSX.Element => {
         line={props.line}
         checked={stationScreenOrientationList[props.line]
           // Ignore disabled stations when determining whether the whole line is selected
-          .filter(({ landscape, portrait }) => landscape || portrait)
+          .filter((lineStation) => isStationSelectable(lineStation))
```
rwaskiewicz, line 26: For my own understanding - `isStationSelectable` expands to:
```ts
    station.landscape ||
    station.portrait ||
    station.showtime_screen_ids.length > 0
```

I'm not super familiar with why `showtime_screen_ids.length > 0` makes a station selectable, which might make this a "dumb" question - is it possible that `station.landscape ||
    station.portrait` evaluates to `false`, but `station.showtime_screen_ids.length > 0` evaluates to `true`, affecting our filter here?

lib/screenplay/emergency_takeover_tool/alerts/alert.ex
```diff
@@ -212,4 +216,106 @@ defmodule Screenplay.EmergencyTakeoverTool.Alerts.Alert do
...
+    end
+  end
+
+  @spec canned_image_path(map(), :indoor | :outdoor, :landscape | :portrait) :: String.t() | nil
+  defp canned_image_path(images, where, orientation) when is_map(images) do
+    alerts_fetch_module = Application.get_env(:screenplay, :alerts_fetch_module)
+
+    image_path =
+      case get_in(images, [where, orientation]) do
+        path when is_binary(path) -> path
```
rwaskiewicz, line 268: Are we assuming that `get_in` will always return a value where `is_binary(val)` is `true`, and that we should fail fast in the event we can't find it?

lib/screenplay/emergency_takeover_tool/alerts/local_fetch.ex
```diff
@@ -14,5 +14,36 @@ defmodule Screenplay.EmergencyTakeoverTool.Alerts.LocalFetch do
...
+  def upload_takeover_image(alert_id, image_data, image_type) do
+    image_dir = asset_directory() |> Path.join(alert_id)
+    File.mkdir_p!(image_dir)
+
+    image_path = Path.join(image_dir, "#{image_type}.png")
+    File.write!(image_path, image_data)
+    :ok
+  end
+
+  # sobelow_skip ["Traversal.FileModule"]
```
rwaskiewicz: For my own knowledge, does sobelow have the concept of "explaining" why we suppress a warning/error? E.g. for ESLint, I'd do something like (rule picked at random here):
```js
// eslint-disable-next-line no-use-before-define -- this is why this suppression is OK
```

rwaskiewicz (approved):

assets/css/stacked-stations.scss
robbie-sundstrom: I'm removing this because there was previously only one Silver Line station that was shown in this column, so I believe there was no need for the checkbox to be shown for this line. Now that we are including Showtime Screens, the Silver Line column includes both South Station and Airport. It makes sense to me that we should be able to select both of these

assets/js/components/EmergencyTakeoverTool/AlertWizard/AlertWizard.tsx
robbie-sundstrom: This was already a bit confusing, and I'm realizing that I could have made what's going on clearer with this change. Each station has three different things associated with it. `portrait` and `landscape` are just a boolean that indicate there are Outfront screens with those orientations at the station. Each station also has a list of showtime screens (prefares, sectionals, busways), which we pass separately since we just need those IDs. When the station names are passed to the backend, we really just want to know if there is at least one of the two `portrait` or `landscape` Outfront screens within a station so we know to upload the takeover images to the SFTP server for the given station. Previously, we only passed stations to the backend that had Outfront screens associated with them, but we need to pass additional station names now so we keep a record of the stations that have only showtime screens but are still showing the emergency takeover.

Maybe this would be more clear if there were a key indicating this is to track that the station `has_outfront` or something of the sort

assets/js/components/EmergencyTakeoverTool/AlertWizard/StationColumn.tsx
robbie-sundstrom: Yep, this previously only allowed stations that had either a `portrait` or `landscape` Outfront screen to be selectable, which was basic logic used in a few different places. We've expanded this now to allow stations with showtime screens, so we definitely do want this to affect our filter and return true when `showtime_screen_ids.length > 0`

lib/screenplay/emergency_takeover_tool/alerts/alert.ex
robbie-sundstrom: This is controlled by the JSON that we created, so this was a place I didn't think was worth it to add error handling. In other words, I think it's fine to fail fast here, because we would need to commit incorrect JSON code for the `canned_messages` for this to cause an error.

assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
digitalcora: In Screens we normally declare components as `const AlertsList: ComponentType<AlertsListProps> = ...`, which handles the return type for you.

assets/js/components/EmergencyTakeoverTool/AlertWizard/StationColumn.tsx
digitalcora: There may be an opportunity to simplify here — I'm not sure why the client code would need to know whether there are specifically portrait/landscape Outfront screens at a station or whether there are Showtime screens (it did need to know that before, since the UI used to display icons indicating what types of screens are at each station). All it really needs to know is _whether_ a given station has _any_ screens that can be taken over; the server knows the full details and what to upload where.

lib/screenplay/emergency_takeover_tool/alerts/local_fetch.ex
digitalcora: It'd probably be a good idea to include an explanatory comment when we suppress a lint even if the tool itself doesn't parse/recognize those comments.

assets/js/components/EmergencyTakeoverTool/AlertWizard/AlertWizard.tsx
rwaskiewicz: Ah, okay - that makes sense. Thank you!

assets/js/components/EmergencyTakeoverTool/AlertWizard/StationColumn.tsx
robbie-sundstrom: I mainly kept it this way for exactly the reason you mentioned - UI showing what type of screens would be at each station. But since we aren't actively planning to show those again, this simplification could definitely be worth doing

PR #737, opened 2026-05-26
fix: Do not overwrite screen configs without emergency_messaging_location

robbie-sundstrom:
The existing pattern matching when setting an emergency takeover assumed that all screens had an `emergency_messaging_location`, and thus accidentally set the config for all screens without one (einks and bus shelters) to nil.


rwaskiewicz (approved):
Before merging, can you add context to the git commit message for the how/why we need this? I think I got there from Slack, but for future us 🙂

PR #738, opened 2026-05-27
refactor: simplify client PA Message types

digitalcora:
* Remove redundant `PaMessageFormData` type which was almost exactly `NewPaMessageBody`.

* Consolidate `NewPaMessageBody` and `UpdatePaMessageBody` into just `PaMessageChange`, which is the partial version. The server does not really have this dichotomy; it has validations on specific fields that apply to the final state of the message, regardless of whether it's a create or an update. This allows an `as` cast to be removed in `PaMessageRow`.

* Define the `message_type` of a `PaMessage` in terms of the `type` of a `StaticTemplate`, rather than the other way around. This allows an unreachable-in-practice branch to be removed in `MainForm`, and a duplicate type to be removed in `StaticTemplatePage`.


robbie-sundstrom (approved):
🧹

PR #741, opened 2026-05-29
fix: do not speak ssml markup

rwaskiewicz:
**Asana task**: [Bug: PA Messages with SSML readout markup](https://app.asana.com/1/15492006741476/project/1213833239363217/task/1214455139781074?focus=true)

Description

Update the server to conditionally no longer html escape SSML content.

In cases where we have PA message templates with predefined audio text
to send to Polly, we store the audio text as a single string. This
string may include SSML, e.g.
```
"""
To make space for other passengers and speed up boarding, please take off your backpack before entering the train and hold it at your side.

<lang xml:lang=\"es-US\"> Para dejar espacio para otros pasajeros y acelerar el embarque, por favor quítese la mochila y manténgala a su lado. </lang>
"""
```
Where the latter half of that message is to be spoken in spanish by
Polly. However, we currently HTML escape the entire string, causing the
XML tags to be transformed from `<lang>` to `&lt;lang&gt;`. This causes
Polly to speak "lang" as if it were a word, rather than an SSML
instruction.

This is done in favor of introducing an XML parser and escaping only the
text nodes in the parsed document as we assume that predefined audio
text such as this is one, valid XML and two, immutable by the client.
This eliminates the potential for handling edge cases with respect to
XML parsing.

This is done in favor of altering the data model to split audio text
into different strings per language to minimize the scope of this effort.

Add prosody rate of 90% and drc effect to spoken text to bring spoken
text into parity with RTS

### DIFF
--- a/assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
@@ -100,9 +100,6 @@ const PaMessageForm = ({
   const [phoneticText, setPhoneticText] = useState(
     defaultValues?.audio_text ?? "",
   );
-  const [phoneticTextHasSsml, setPhoneticTextHasSsml] = useState<boolean>(
-    defaultValues?.audio_text_has_ssml ?? false,
-  );
   const [audioURL, setAudioURL] = useState(defaultValues?.audio_url ?? "");
 
   const [signIds, setSignIds] = useState<string[]>(() => {
@@ -128,7 +125,6 @@ const PaMessageForm = ({
     setSelectedTemplate(null);
     setVisualText("");
     setPhoneticText("");
-    setPhoneticTextHasSsml(false);
     setAudioURL("");
     setAudioState(AudioPreview.Unreviewed);
     onChangePriority(defaultPriority);
@@ -217,7 +213,6 @@ const PaMessageForm = ({
           interval,
           navigateTo: setPage,
           phoneticText,
-          phoneticTextHasSsml,
           audioURL,
           priority,
           setDays,
@@ -293,7 +288,6 @@ const PaMessageForm = ({
             setSelectedTemplate(template);
             setVisualText(template.visual_text);
             setPhoneticText(template.audio_text ?? "");
-            setPhoneticTextHasSsml(template.audio_text_has_ssml ?? false);
             setAudioURL(template.audio_url ?? "");
             onChangePriority(template.type === "psa" ? 5 : 1);
             setAudioState(AudioPreview.Unreviewed);
--- a/assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
@@ -100,6 +100,9 @@ const PaMessageForm = ({
   const [phoneticText, setPhoneticText] = useState(
     defaultValues?.audio_text ?? "",
   );
+  const [phoneticTextHasSsml, setPhoneticTextHasSsml] = useState<boolean>(
+    defaultValues?.audio_text_has_ssml ?? false,
+  );
   const [audioURL, setAudioURL] = useState(defaultValues?.audio_url ?? "");
 
   const [signIds, setSignIds] = useState<string[]>(() => {
@@ -125,6 +128,7 @@ const PaMessageForm = ({
     setSelectedTemplate(null);
     setVisualText("");
     setPhoneticText("");
+    setPhoneticTextHasSsml(false);
     setAudioURL("");
     setAudioState(AudioPreview.Unreviewed);
     onChangePriority(defaultPriority);
@@ -213,6 +217,7 @@ const PaMessageForm = ({
           interval,
           navigateTo: setPage,
           phoneticText,
+          phoneticTextHasSsml,
           audioURL,
           priority,
           setDays,
@@ -288,6 +293,7 @@ const PaMessageForm = ({
             setSelectedTemplate(template);
             setVisualText(template.visual_text);
             setPhoneticText(template.audio_text ?? "");
+            setPhoneticTextHasSsml(template.audio_text_has_ssml ?? false);
             setAudioURL(template.audio_url ?? "");
             onChangePriority(template.type === "psa" ? 5 : 1);
             setAudioState(AudioPreview.Unreviewed);
--- a/assets/js/models/static_template.ts
+++ b/assets/js/models/static_template.ts
@@ -5,6 +5,7 @@ export interface StaticTemplate {
   title: string;
   visual_text: string;
   audio_text?: string;
+  audio_text_has_ssml?: boolean;
   audio_url?: string;
   type: TemplateType;
   archived: boolean;
--- a/assets/static/static_templates.json
+++ b/assets/static/static_templates.json
@@ -27,7 +27,7 @@
     "id": 4,
     "title": "Bicycles (English + Spanish)",
     "visual_text": "To make space for other passengers, bicycles are not permitted on trains during rush hour, or at any time of day on the Green Line. For details visit mbta.com/bikes.\n\nPara dejar espacio para otros pasajeros, bicicletas no son permitidas en los trenes durante hora pico y nunca son permitidas en la linea verde. Para mas informacion, visite mbta.com/bikes",
-    "audio_text": "To make space for other passengers, bicycles are not permitted on trains during rush hour, or at any time of day on the Green Line. For details visit mbta.com/bikes.\n\n<lang xml:lang=\"es-US\"> Para dejar espacio para otros pasajeros, bicicletas no son permitidas en los trenes durante hora pico y nunca son permitidas en la línea verde. Para más información, visite [mbta.com/bikes](http://mbta.com/bikes) </lang>",
+    "audio_text": "To make space for other passengers, bicycles are not permitted on trains during rush hour, or at any time of day on the Green Line. For details visit mbta.com/bikes.\n\n<lang xml:lang=\"es-US\"> Para dejar espacio para otros pasajeros, bicicletas no son permitidas en los trenes durante hora pico y nunca son permitidas en la línea verde. Para más información, visite mbta.com/bikes </lang>",
     "audio_text_has_ssml": true,
     "type": "psa",
     "archived": false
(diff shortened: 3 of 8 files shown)


lib/screenplay/watts/client.ex
```diff
@@ -4,7 +4,7 @@ defmodule Screenplay.Watts.ClientBehaviour do
   @doc """
   Fetches an audio file from Watts given a string.
   """
-  @callback fetch_tts(String.t()) :: {:ok, binary()} | :error
+  @callback fetch_tts(String.t(), boolean) :: {:ok, binary()} | :error
```
rwaskiewicz: Asking for opinions here - boolean parameters here feel like a code smell. Unsure if that's Uncle Bob whispering in my ear, the ghost of Crystal Reports past, or something else. Would it be better (i.e. more idiomatic) to have an `opts` arg and match on a `has_ssml` field in there?

assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
```diff
@@ -100,6 +100,9 @@ const PaMessageForm = ({
   const [phoneticText, setPhoneticText] = useState(
     defaultValues?.audio_text ?? "",
   );
+  const [phoneticTextHasSsml, setPhoneticTextHasSsml] = useState<boolean>(
```
rwaskiewicz: I _think_ between here and `MainForm` I've covered the various cases where this state is required, but asking someone with a little more experience to take a second look and evaluate that through the UI to double check me here

lib/screenplay/watts/client.ex
```diff
@@ -18,12 +18,21 @@ defmodule Screenplay.Watts.Client do
...
     watts_url = Application.fetch_env!(:screenplay, :watts_url)
     watts_api_key = Application.fetch_env!(:screenplay, :watts_api_key)
 
-    text = Phoenix.HTML.html_escape(text) |> Phoenix.HTML.safe_to_string()
-    request_data = Jason.encode!(%{text: "<speak>#{text}</speak>", voice_id: "Matthew"})
+    request_data =
+      Jason.encode!(%{
+        text:
+          ~s(<speak><amazon:effect name="drc"><prosody rate="90%">#{text}</prosody></amazon:effect></speak>),
+        voice_id: "Matthew"
```
rwaskiewicz, line 34: One call out here - we hardcode "Matthew" here. However, [for Spanish in RTS, we use "Mia"](https://github.com/mbta/realtime_signs/blob/41b7caf11ba562b9640545c261758c7dce58f905/lib/pa_ess/updater.ex#L200-L201). Any thoughts/concerns on this divergence?

digitalcora: It's not possible to switch voices in the middle of a single synthesized audio file. We'd probably prefer to use a "natively" Spanish-speaking voice such as Mia for the Spanish half of bilingual PSAs, but iirc that would have complicated the implementation quite a bit. The `lang` tag at least makes it parse the phonemes correctly, though Matt still pronounces them with an extremely English "accent".

rwaskiewicz: Ah, okay - I hadn't considered that wasn't possible, only that we had a discrepancy. In hindsight, it makes sense that this works in RTS since we separate the English and Spanish audio, and can therefore have Matthew and Mia speak. Thanks!

lib/screenplay/watts/client.ex
digitalcora: I lean YAGNI in cases like this — right now there's only one option, so having a general `opts` arg and pulling that one option out of it is unneeded conceptual overhead. If this were a published library, there might be a more compelling case for `opts` as a way of heading off possible backwards-incompatible updates later on.

assets/js/models/static_template.ts
```diff
@@ -5,6 +5,7 @@ export interface StaticTemplate {
   title: string;
   visual_text: string;
   audio_text?: string;
+  audio_text_has_ssml?: boolean;
```
digitalcora, line 8: If we omitted this property and just always assumed static templates are written in SSML, would that allow simplifying any of the client logic? It feels like a reasonable thing to do although I can appreciate the explicitness of having to "opt in" when a new template is authored.

rwaskiewicz: I think we could, but I think that would require us to look at all of the non-archived templates here and evaluate if we need to escape any of their characters. 

According to the [AWS Polly docs](https://docs.aws.amazon.com/polly/latest/dg/escapees.html):

> Additionallty, when you use the apostrophe/single quotation mark (') as an apostrophe, you must also use the escape code.

As of this moment, the only ones I see that would require escaping would be apostrophes in a handful of messages. Which, we could do easily, but we'd see that in the UI:

Image

That said, I did a cursory test of this message _without_ the escaping, which read fine IMO. However, I don't know under what conditions not-escaping would fail us here.

That being said, I think I lean towards leaving this as-is for the time being as to not affect existing templates and make future template creation a little more explicit. However, I don't feel that strongly about it if others prefer escaping the rest of our active templates here.

digitalcora: The supposed rules for escaping `'` listed in that AWS doc are really odd... "Must be escaped when used as an apostrophe" makes no sense because it's seemingly about the semantic content of the text (and they immediately violate/contradict that rule in the very next example...). And, as you've seen, not escaping it even "when used as an apostrophe" works fine, which is in line with XML escaping rules.

> I think I lean towards leaving this as-is for the time being as to not affect existing templates and make future template creation a little more explicit.

Sound reasoning 👍

assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
```diff
@@ -125,6 +128,7 @@ const PaMessageForm = ({
     setSelectedTemplate(null);
     setVisualText("");
     setPhoneticText("");
+    setPhoneticTextHasSsml(false);
```
digitalcora: I'm confused as to why this needs to be an independent piece of state from `selectedTemplate`, which if it exists would contain the `audio_text_has_ssml` field which is the same value. `selectedTemplate` appears to be present as state both here and in `MainForm`.

assets/static/static_templates.json
```diff
@@ -28,6 +28,7 @@
     "title": "Bicycles (English + Spanish)",
     "visual_text": "To make space for other passengers, bicycles are not permitted on trains during rush hour, or at any time of day on the Green Line. For details visit mbta.com/bikes.\n\nPara dejar espacio para otros pasajeros, bicicletas no son permitidas en los trenes durante hora pico y nunca son permitidas en la linea verde. Para mas informacion, visite mbta.com/bikes",
     "audio_text": "To make space for other passengers, bicycles are not permitted on trains during rush hour, or at any time of day on the Green Line. For details visit mbta.com/bikes.\n\n<lang xml:lang=\"es-US\"> Para dejar espacio para otros pasajeros, bicicletas no son permitidas en los trenes durante hora pico y nunca son permitidas en la línea verde. Para más información, visite [mbta.com/bikes](http://mbta.com/bikes) </lang>",
```
digitalcora: Unrelated to the SSML changes, but there is some stray Markdown formatting in this message we should remove.

lib/screenplay/watts/client.ex
```diff
@@ -18,12 +18,21 @@ defmodule Screenplay.Watts.Client do
   @http_client Application.compile_env!(:screenplay, :http_client)
 
   @impl true
-  def fetch_tts(text) do
+  def fetch_tts(text, false) do
```
digitalcora: Since this argument isn't actually named anywhere, it might improve readability if we named it using e.g. `fetch_tts(text, _is_ssml = false)`. Another not-mutually-exclusive possibility, which enables better LSP docs, is giving it an alias in the typespec, e.g. `fetch_tts(String.t(), is_ssml :: boolean())`.

lib/screenplay_web/controllers/pa_messages_api_controller.ex
```diff
@@ -15,8 +15,8 @@ defmodule ScreenplayWeb.PaMessagesApiController do
     json(conn, PaMessages.get_active_messages())
   end
 
-  def preview_audio(conn, %{"text" => text}) do
-    case @watts_client.fetch_tts(text) do
+  def preview_audio(conn, %{"text" => text, "has_ssml" => has_ssml}) do
+    case @watts_client.fetch_tts(text, has_ssml === "true") do
```
digitalcora: Conventionally this would just be `==`; "strict" equality in Elixir only needs to be used when you specifically don't want same-valued integers and floats to be considered equal (`1 == 1.0` is true, `1 === 1.0` is false). It only affects that one thing.

test/screenplay/watts/client_test.exs
```diff
@@ -0,0 +1,48 @@
+defmodule Screenplay.Watts.ClientTest do
+  use ExUnit.Case
+
+  import Mox
```
digitalcora, line 4: When using Mox we should also `setup :verify_on_exit!`. This adds an exit handler that ensures all the `expect`s you defined were actually called, and fails the test if not. (you can use `stub` instead if you don't care whether they're called)

assets/js/components/Dashboard/PaMessageForm/PaMessageForm.tsx
rwaskiewicz: That's fair, I hadn't thought about that as a part of the above conversation about the `audio_text_has_ssml` - here I was following the general usage pattern w/o considering your point about this field being exclusive to templates. Removed in bee742fa221ad476d5b1bd6f6fa75f408150540f

assets/static/static_templates.json
rwaskiewicz: Removed in 9161fd5c2e5aa433b8a774720a56606c64ecf9a9

lib/screenplay/watts/client.ex
rwaskiewicz: Good point! Fixed in 34786d15fcab9a6ca70db45313c3d1a93a1602a6

lib/screenplay_web/controllers/pa_messages_api_controller.ex
rwaskiewicz: Ah, that's my JavaScript side showing. TIL. Updated in 96c914dc78a89f7f97716208cc12da35036d22d0

test/screenplay/watts/client_test.exs
rwaskiewicz: Thanks! TIL. Updated in 6f22c793b49d69d4eb8fb59deee799f623665f38

digitalcora (approved):

PR #743, opened 2026-06-03
chore: address sass deprecations

rwaskiewicz:
**Asana task**: [Screenplay: address Sass deprecations](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214835715894509?focus=true)

Description

Begin to move away from `@import` statements by replacing them with the `@use` directive. Unless otherwise specified, all `@import` statements have been moved to `@use` statements.

`css/screenplay-form-variables.scss` has been introduced to break circular dependencies between `screenplay.scss` (where the variables in the former previously were defined), `variables.scss`, and users of the variables in `screenplay-form-variables.scss`.

Bootstrap `@import` statements in `app.scss` have been left alone, as Bootstrap still internally relies on `@import`. Converting these statements to `@use` in `app.scss` results in mixins and variables assumed to be globally available to no longer be found by Sass.

Update the build script following these updates. Two options are added to the Sass loader:
1. `quietDeps`, to suppress warnings from third party libraries that the Screens team has no control over
2. `silenceDeprecations`, to suppress cases where we continue to use the `@import` directive in our code base.

The above changes suppress warnings, not fix them. Once Bootstrap 6 is released, care should be taken to remove these fields from the configuration.

Update the build script when running the `deploy` command to fail when one or more warning are emitted by esbuild. This will cause the command to exit with a status code that can be picked up by CI to fail builds in GitHub Actions.

Notes

I made CI fail intentionally with only warnings from the Sass loader to verify it works:
Screenshot 2026-06-03 at 13 15 12


rwaskiewicz:
The Asana integration is failing - not a blocker per se, but I'll TAL at that this afternoon

robbie-sundstrom:
@rwaskiewicz I've seen the Asana integration fail very intermittently and seemingly randomly, but only in this repo. I don't think it's worth spending too much time looking into it since it's succeeded on retry when I've seen it happen in the past.

assets/css/components/button-group.scss
```diff
@@ -1,12 +1,14 @@
+@use "../variables.scss";
```
digitalcora: I don't think specifying the `.scss` extension on imports/uses should be necessary, at least it isn't in Screens (though maybe this is a Webpack/esbuild difference).

assets/build.mjs
```diff
@@ -1,8 +1,28 @@
...
+ * has been migrated, these options should be removed.
+ *
+ * @type {Partial<import("esbuild-sass-plugin").SassPluginOptions>}
+ */
+const useKeywordAdoptionOpts = {
+  // Suppress warnings from Bootstrap (and other third party dependencies)
+  quietDeps: true,
+  // We still need to use `@import` on Bootstrap, and `quietDeps` does not
+  // suppress those import statements.
+  silenceDeprecations: ["import"],
```
digitalcora, line 20: Is there a way to use comments to specifically silence the warnings on our Bootstrap `@import` statements, but continue to have them elsewhere so we don't accidentally introduce new `@import`s for things that don't require them?

assets/css/variables-screenplay-forms.scss
```diff
@@ -0,0 +1,7 @@
+@use "variables.scss";
+
+$form-check-input-checked-color: #0f1417;
+$form-check-input-checked-bg-color: #8ecdff;
+$form-check-input-bg: #0f1417;
+$form-check-input-border: 1px solid #8b9198;
+$form-feedback-invalid-color: variables.$text-error;
```
robbie-sundstrom, line 8: I like having all these Bootstrap styles consolidated here! My majorly small nitpicks would be adding a comment that labels them as Bootstrap styles while you're cleaning this up, and renaming the file to `variables-screenplay-forms.scss` so that it's sorted with the other `variables.scss` file. Definitely take it or leave it suggestions

rwaskiewicz: All good suggestions! The former is done in https://github.com/mbta/screenplay/pull/743/commits/4db74ff5c25d97f3a54db03e1f9cbb1a6bed1aa6, and the latter in https://github.com/mbta/screenplay/pull/743/commits/9dcd6486732300db63a4e7e4b7aa85c62f9d3fbe

assets/css/components/button-group.scss
rwaskiewicz: True, removed in https://github.com/mbta/screenplay/pull/743/commits/fd6aad9af494058f7edc039164fe8809d1929f37

assets/build.mjs
rwaskiewicz: Unfortunately not (not out of the box at least) 😞 I agree this isn't ideal, as it requires us to be more vigilant about the introduction of `@import` statements/doesn't let us offload that work to any of these tools.

robbie-sundstrom (approved):

PR #744, opened 2026-06-04
add basic tool for uploading FIFA messages

panentheos:
**Asana task**: [Match Day PA Message Configuration](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214901474689447?focus=true)

This adds a page to allow bulk uploading of FIFA messages via a defined CSV file. The feature is intentionally minimal, since it will only be used a handful of times during the next few weeks. Tested with some handcrafted dummy data (see ticket for data format).

### DIFF
--- a/assets/js/components/Dashboard/UploadPaMessages.tsx
+++ b/assets/js/components/Dashboard/UploadPaMessages.tsx
@@ -1,0 +1,166 @@
+import React, { useState } from "react";
+import fp from "lodash/fp";
+import { Button } from "react-bootstrap";
+import { parse } from "csv-parse/browser/esm/sync";
+import { createNewPaMessage } from "Utils/api";
+import { useScreenplayState } from "Hooks/useScreenplayContext";
+import { signDirection } from "../../util";
+import Toast from "Components/Toast";
+
+const gameDates = [
+  {
+    id: "6/13",
+    start: new Date("2026-06-13T04:00:00"),
+    end: new Date("2026-06-13T18:00:00"),
+  },
+  {
+    id: "6/16",
+    start: new Date("2026-06-16T04:00:00"),
+    end: new Date("2026-06-16T15:00:00"),
+  },
+  {
+    id: "6/19",
+    start: new Date("2026-06-19T04:00:00"),
+    end: new Date("2026-06-19T15:00:00"),
+  },
+  {
+    id: "6/23",
+    start: new Date("2026-06-23T04:00:00"),
+    end: new Date("2026-06-23T13:00:00"),
+  },
+  {
+    id: "6/26",
+    start: new Date("2026-06-26T04:00:00"),
+    end: new Date("2026-06-26T12:00:00"),
+  },
+  {
+    id: "6/29",
+    start: new Date("2026-06-29T04:00:00"),
+    end: new Date("2026-06-29T13:30:00"),
+  },
+  {
+    id: "7/09",
+    start: new Date("2026-07-09T04:00:00"),
+    end: new Date("2026-07-09T13:00:00"),
+  },
+];
+
+const UploadPaMessages = () => {
+  const [file, setFile] = useState<File>();
+  const [dateId, setDateId] = useState(gameDates[0].id);
+  const [loading, setLoading] = useState(false);
+  const [result, setResult] = useState<string | null>(null);
+  const { places } = useScreenplayState();
+
+  const submit = async () => {
+    try {
+      setLoading(true);
+      const { start, end } = gameDates.find((item) => item.id === dateId)!;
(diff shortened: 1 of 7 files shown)


assets/js/util.ts
```diff
@@ -170,17 +170,19 @@ export const placesWithSelectedAlert = (
     : [];
 };
 
+export const signDirection = (sign: Screen) => {
```
rwaskiewicz, line 173: Non-blocking because this is a refactor/not required for this ticket - we might be able to remove `fp` from this function in the future in favor of JS' support for intersections on `Set` objects at some point ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection))

assets/js/components/Dashboard/UploadPaMessages.tsx
```diff
@@ -0,0 +1,166 @@
...
+        const place = placesLookup[station];
+        if (!place) {
+          throw new Error(`Couldn't find place named ${station}`);
+        }
+        return place.screens
+          .filter(
+            (screen) =>
+              screen.type === "pa_ess" &&
+              signDirection(screen) === where &&
+              screen.routes![0].id.startsWith(line),
```
rwaskiewicz, line 76: Double checking - is the `id` for a route always capital cased? Does it make to to lowercase these IDs just in case we get `"blue".startsWith("Blue")`?

assets/js/components/Dashboard/UploadPaMessages.tsx
```diff
@@ -0,0 +1,166 @@
...
+        }))
+        .filter(({ message, signIds }) => message && signIds.length);
+
+      await Promise.all(
+        groupedItems.map(async ({ message, signIds }) => {
+          const res = await createNewPaMessage({
+            alert_id: null,
+            start_datetime: start.toISOString(),
+            end_datetime: end.toISOString(),
+            days_of_week: [1, 2, 3, 4, 5, 6, 7],
```
rwaskiewicz, line 102: For my own understanding, why do we set this for every day of the week? Is it that we have a start/end date, and the message will play every day between those datetimes (inclusive)?

assets/js/components/Dashboard/UploadPaMessages.tsx
```diff
@@ -0,0 +1,166 @@
...
+            priority: 5,
+            interval_in_minutes: 4,
+            visual_text: message,
+            audio_text: message,
+            audio_url: "",
+            message_type: null,
+            template_id: null,
+          });
+          if (res.status !== 200) {
+            throw JSON.stringify(res.errors);
```
rwaskiewicz, line 113: If we fail on one or more of these creation attempts, what's our remediation strategy? Do we need to delete all the successfully created announcements and try again?

rwaskiewicz (commented):
I think I'm running into data issues - using the spreadsheet (w/o the header) that we discussed offline, I'm running into place lookup issues:

Image

I wonder if this is just a data issue?

assets/js/util.ts
digitalcora: "at some point" could well be today — Set intersection has been supported in all major desktop browsers for ~2 years and I believe everyone who would have access to Screenplay in the first place should be using an evergreen browser.

assets/js/components/Dashboard/UploadPaMessages.tsx
panentheos: These should always be capitalized, yep.

assets/js/components/Dashboard/UploadPaMessages.tsx
panentheos: This is a filter that is meant for longer-running messages to make them only apply on certain days, but since we're narrowly specifying the periods with these messages, we just pass the default of all days.

assets/js/components/Dashboard/UploadPaMessages.tsx
panentheos: This is a cheap fallback and debugging tool. The plan is to pre-validate the exact CSV we're handing over, so we don't expect any errors from the server. It's unlikely but possible that one or more of the creations could fail due to e.g. a network blip, in which case yes, we'd need them to do some manual cleanup and retry.

panentheos:
That error looks like a data issue, since that place name in Screenplay is "Massachusetts Avenue". We'll need to clean those up in the final CSV.

rwaskiewicz (approved):
FWIW - once I updated:
- `Massachusetts Ave` -> `Massachusetts Avenue`
- Removed a trailing space from 'Savin Hill'

Upload appeared to work 👍

PR #745, opened 2026-06-04
fix: restore bootstrap overrides

rwaskiewicz:
**Asana task**: N/A

Description
In https://github.com/mbta/screenplay/commit/ad3b11bafdb35026bff08dccfb20f3aee29dc9ca, we moved from `@import` to `@use` statements for most
stylesheets. That change introduced a regression which led to certain
default bootstrap styles to be used instead of the custom primary theme
for screenplay. Migrating a single `@import` of Bootstrap to a `@use`
caused the theme map we define in `Screenplay.scss` to not be taken into
account when `bs-btn-*` styles are applied. As a result, the default
Bootstrap Blue was used in a handful of UI components.

This commit also includes a where the custom color of PA message
"active" toggle was not being properly applied. This was caused by the
previously global style not being applied to the toggle as a part of the
migration.

---
**Before - Action Buttons**
Screenshot 2026-06-05 at 10 14 02

**After - Action Buttons**

Screenshot 2026-06-04 at 14 51 23

---

**Before - Active Toggle**
Screenshot 2026-06-05 at 10 14 31

**After - Active Toggle**
Screenshot 2026-06-05 at 10 14 48

### DIFF
--- a/assets/css/dashboard/pa-messages-page.scss
+++ b/assets/css/dashboard/pa-messages-page.scss
@@ -1,5 +1,4 @@
 @use "../variables";
-@use "../variables-screenplay-forms";
 
 .pause-active-switch-container {
   position: relative;
@@ -9,8 +8,6 @@
       width: 86px;
       height: 24px;
       margin-top: 0;
-      background-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
-      border-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
     }
 
     .form-check-input:not(:checked) {
--- a/assets/css/dashboard/pa-messages-page.scss
+++ b/assets/css/dashboard/pa-messages-page.scss
@@ -1,4 +1,5 @@
 @use "../variables";
+@use "../variables-screenplay-forms";
 
 .pause-active-switch-container {
   position: relative;
@@ -8,6 +9,8 @@
       width: 86px;
       height: 24px;
       margin-top: 0;
+      background-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
+      border-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
     }
 
     .form-check-input:not(:checked) {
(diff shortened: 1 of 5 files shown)


assets/css/dashboard/pa-messages-page.scss
```diff
@@ -8,6 +9,8 @@
       width: 86px;
       height: 24px;
       margin-top: 0;
+      background-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
+      border-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
```
rwaskiewicz: I had to double check/verify this, but we were using the background color as the border color here prior to the Sass `@import` -> `@use` migration

assets/css/dashboard/pa-messages-page.scss
```diff
@@ -8,6 +9,8 @@
       width: 86px;
       height: 24px;
       margin-top: 0;
+      background-color: variables-screenplay-forms.$form-check-input-checked-bg-color;
```
digitalcora: tangential suggestion: Maybe we could reduce some of the duplication in these variable names — they probably made more sense with a global namespace, but now e.g. the word "form" is repeated twice. Like this could be `variables-screenplay-forms.check-input-checked-bg-color`. (The `screenplay` also seems redundant/unneeded; isn't this a module _in_ Screenplay already?)

digitalcora (approved):

assets/css/dashboard/pa-messages-page.scss
rwaskiewicz: Good point - I realized that we can eliminate `variables-screenplay-forms` entirely while fixing one of the regressions. Previously, the `$form-check-input*` variables weren't being applied to the style map defined in `screenplay.scss`. I've inlined them as an alternate solution.

There were a few direct usages of those variables (including here). I've replaced them with their `variables.scss` equivalents, which existed for every variable in `variables-screenplay-forms`

digitalcora: Even better! ✨

PR #756, opened 2026-06-10
fix: ensure only pa-message-admin role can upload csv

rwaskiewicz:
**Asana task**: [Slack](https://mbta.slack.com/archives/C039ARUM48H/p1781105097538789)

Description

Update upload component to only show the button to upload a CSV when users have the pa-message-admin role. This should (correctly) fail on the backend already, this is a presentational change only


robbie-sundstrom (approved):
🚢

PR #760, opened 2026-06-16
feat: Store Emergency Takeover Alerts in Postgres

robbie-sundstrom:
**Asana task**: [Bug - Inconsistent display of current ETT alert](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215504297867595?focus=true)

Adds a new DB table to our existing Postgres setup in RDS that all instances of Screenplay will be able to read from without consistency issues. 

- Migration to create the new Postgres table
- Data access functions are all located in `Screenplay.EmergencyTakeovers`
- Deletes a lot of code for managing the GenServer state for Emergency Takeover alerts
- Moves some code around 
  - S3 access is now limited to images within the Emergency Takeover Tool, so those are in a renamed directory
  - Some of the types defined in the `EmergencyTakeover` module are from the old `EmergencyTakeoverTool.Alerts` module
  - Screenplay.EmergencyTakeoverTool.ScreensConfig is largely the same code that existed in a section of `Screenplay.EmergencyTakeoverTool.Alerts.Alert`


assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
```diff
@@ -14,30 +14,20 @@ interface AlertsListProps {
   closeModal: () => void;
 }
 
-const fetchActiveAlerts = withErrorHandling(
+const fetchAllAlerts = withErrorHandling(
   async () => {
-    const response = await fetch(`${BASE_URL}/active_alerts`);
+    const response = await fetch(`${BASE_URL}/active_and_past_alerts`);
```
digitalcora: suggestion: "active and past" seems redundant, as those are the only two possible states (there is no such thing as a "future" emergency takeover). Any reason for this not to just be `/alerts`?

assets/js/components/EmergencyTakeoverTool/AlertDashboard/AlertsList.tsx
```diff
@@ -14,30 +14,20 @@ interface AlertsListProps {
...
     if (!response.ok) {
       throw response;
     }
     return response.json();
   },
   {
-    customMessage: "Failed to load active alerts. Please refresh the page.",
+    customMessage:
+      "Failed to load current and past alerts. Please refresh the page.",
```
digitalcora: suggestion: same here with "current and past alerts" rather than just "alerts".

assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
```diff
@@ -45,6 +45,7 @@ const CreateMessage = ({ value, onChange }: CreateMessageProps) => {
             className="message-select text-16"
             value={value.type === "canned" ? value.id : -1}
             onChange={(e) => {
+              // TODO: Bug found: this needs to be updated so that it has the images on creation of CreateMessage
```
digitalcora: Are you fixing this separately?

config/dev.exs
```diff
@@ -66,7 +66,7 @@ config :screenplay,
   sftp_password: System.get_env("SFTP_PASSWORD"),
   sftp_remote_path: System.get_env("SFTP_REMOTE_PATH"),
   sftp_local_path: System.get_env("SFTP_LOCAL_PATH"),
-  alerts_fetch_module: Screenplay.EmergencyTakeoverTool.Alerts.LocalFetch,
+  alerts_fetch_module: Screenplay.EmergencyTakeoverTool.Images.LocalFetch,
```
digitalcora: suggestion: Naming this the "alerts" fetch module doesn't seem accurate anymore since it doesn't fetch alert configuration at all; and "fetch" doesn't seem accurate either (in this config key or the module naming) since it does more than fetching, it puts/mutates. Maybe `image_store_module`? (going off your renaming of the namespace to `Screenplay.EmergencyTakeoverTool.Images`; so it would then be `Images.S3Store`, `Images.LocalStore`, etc.)

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,105 @@
...
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @derive {Jason.Encoder, except: [:__meta__]}
+
+  @type t() :: %__MODULE__{
+          id: integer(),
+          message: map(),
+          stations: [String.t()],
```
digitalcora: * This is station IDs now, right? Can we name it to make that clear, i.e. `station_ids`
* Should this type be `[station()]`? (or `[station_id()]` perhaps to align with the above)

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,105 @@
...
+  alias Screenplay.Util
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @derive {Jason.Encoder, except: [:__meta__]}
+
+  @type t() :: %__MODULE__{
+          id: integer(),
+          message: map(),
```
digitalcora: Should this type be `message()`?

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,105 @@
...
+
+  def message_from_json(%{"type" => "canned", "id" => id}) do
+    %{type: :canned, id: id}
+  end
+
+  def message_from_json(%{
+        "type" => "custom",
+        "text" => %{"indoor" => indoor, "outdoor" => outdoor}
+      }) do
+    %{type: :custom, text: %{indoor: indoor, outdoor: outdoor}}
```
digitalcora: Can we integrate this into the casting of the `message` field so "users" of this module don't have to do it themselves? Seems like it should be part of `changeset`; `Ecto.Type` could be one way to do this.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,105 @@
...
+    |> validate_required([
+      :start_time,
+      :stations,
+      :message,
+      :created_by
+    ])
+    |> validate_length(:stations, min: 1)
+  end
+
+  def new(message, stations, schedule, user) do
```
digitalcora: suggestion: this could have a typespec, all the relevant types already exist.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,105 @@
...
+    %{
```
digitalcora: suggestion: I think this could be `%__MODULE__` for better compile-time correctness checking.

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+  @spec get_active_and_past_alerts() :: {[map()], [map()]}
+  def get_active_and_past_alerts do
+    EmergencyTakeover
+    |> order_by([alert], desc: alert.start_time)
+    |> Repo.all()
+    |> Enum.map(&to_json/1)
+    |> Enum.group_by(fn %{"cleared_at" => cleared_at} ->
+      if cleared_at, do: :past, else: :current
+    end)
+    |> then(fn groups -> {groups[:current] || [], groups[:past] || []} end)
```
digitalcora: It seems like there's a disagreement in our terminology here — it's "active" in the function name, but "current" in the returned map.

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+  @type alert_update :: %{
+          message: EmergencyTakeover.message(),
+          stations: list(EmergencyTakeover.station()),
+          schedule: EmergencyTakeover.schedule()
+        }
+
+  @spec get_active_and_past_alerts() :: {[map()], [map()]}
+  def get_active_and_past_alerts do
+    EmergencyTakeover
+    |> order_by([alert], desc: alert.start_time)
```
digitalcora: I think just `order_by(desc: :start_time)` would suffice here, going by the doc examples.

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+    |> Repo.all()
+    |> Enum.map(&to_json/1)
```
digitalcora: Is the context module the appropriate place for this encoding to happen? Assuming it's for the sake of the client, maybe it should go in the controller if anything. (this seems to also be causing some awkwardness where there's a separate and largely-duplicated `_non_json` version of a function)

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+
+  @spec create_alert(map()) :: {:ok, EmergencyTakeover.t()} | {:error, Ecto.Changeset.t()}
+  def create_alert(params) do
+    %EmergencyTakeover{}
+    |> EmergencyTakeover.changeset(params)
+    |> Repo.insert()
+  end
+
+  @spec edit_alert(integer(), alert_update(), String.t()) :: :ok | {:error, String.t()}
+  def edit_alert(id, changes, user) do
```
digitalcora: suggestion: I think this would be better named `update_alert` — "update" being the act of actually mutating a record in a database (see the naming of `Repo.update`). Conventionally "edit" is used to refer to the GET route which serves a web page that lets you update a record (see `Phoenix.Router.resources/4`).

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+    |> Enum.into(%{})
+  end
+
+  defp stringify_keys(value) when is_list(value), do: Enum.map(value, &stringify_keys/1)
+  defp stringify_keys(value), do: value
+
+  @spec remove_overlapping_alerts(integer() | nil, list(String.t()), String.t()) ::
+          list(String.t())
+  def remove_overlapping_alerts(id, new_stations, user) do
+    get_overlapping_alerts(new_stations, id)
```
digitalcora, line 108: This is the only place `get_overlapping_alerts` is used, except for tests — could it be a private function, or its code included directly within this function?

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -21,17 +22,20 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
...
     with schedule <- schedule_from_duration(DateTime.utc_now(), duration_in_hours),
          user <- get_session(conn, "username"),
-         message_struct <- Alert.message_from_json(message),
-         alert <- Alert.new(message_struct, stations, schedule, user),
-         params_to_log =
-           params
-           |> Map.take(["message", "stations", "duration"])
-           |> Map.merge(%{"id" => alert.id}),
+         message_struct <- EmergencyTakeover.message_from_json(message),
+         alert <- EmergencyTakeover.new(message_struct, stations, schedule, user),
```
digitalcora: `var <- ...` always matches, so there's no point in using `<-` versus `=` here; I think `=` would be preferable to make it clear these are plain assignments, not possible branch points. Indeed, they could all be done outside the `with`, up until the first `:ok <-`.

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -49,37 +53,30 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
...
+         message_struct <- EmergencyTakeover.message_from_json(message),
          changes = %{
            message: message_struct,
            stations: stations,
            schedule: schedule
          },
          user <- get_session(conn, "username"),
-         :ok <- remove_overlapping_alerts(params, user),
-         new_alert <- Alert.update(alert, changes, user),
+         id <- String.to_integer(id_str),
```
digitalcora: Same comment as above; everything up to and including this line could be plain assignments outside the `with`.

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -97,10 +94,9 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
   def clear(conn, params = %{"id" => id}) do
     with user <- get_session(conn, "username"),
          :ok <- UserActionLogger.log(user, :clear_alert, params),
-         alert <- State.get_alert(id),
-         %Alert{stations: stations} = alert,
-         cleared_alert <- Alert.clear(alert, user),
-         :ok <- State.clear_alert(cleared_alert),
+         alert <- EmergencyTakeovers.get_alert(id),
```
digitalcora: As above, this could be a plain assignment.

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+      {:ok, _updated_alert} ->
+        :ok
+
+      {:error, changeset} ->
+        {:error,
+         "Failed to edit alert #{alert.id} with the changes: #{inspect(changeset.errors)}"}
+    end
+  end
+
+  @spec clear_alert(EmergencyTakeover.t(), String.t()) :: :ok | :error
```
digitalcora: This typespec doesn't match the returned value in the error case.

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -109,14 +105,19 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
         conn
         |> put_status(:internal_server_error)
         |> json(%{success: false, error: reason})
+
+      :error ->
```
digitalcora: Related to previous comment: this matches the (incorrect) typespec, but not what's actually returned.

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -214,13 +214,13 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
     end)
   end
 
-  def active_alerts(conn, _params) do
-    alerts_json = State.get_active_alerts() |> Enum.map(&Alert.to_json/1)
-    json(conn, alerts_json)
+  def active_and_past_alerts(conn, _params) do
+    {current_alerts, past_alerts} = EmergencyTakeovers.get_active_and_past_alerts()
+    json(conn, %{"current" => current_alerts, "past" => past_alerts})
```
digitalcora: Similar comment as previously re: "active" vs. "current".

lib/screenplay/emergency_takeovers.ex
```diff
@@ -0,0 +1,188 @@
...
+  alias Screenplay.Repo
+  alias Screenplay.Util
+
+  @type alert_update :: %{
+          message: EmergencyTakeover.message(),
+          stations: list(EmergencyTakeover.station()),
+          schedule: EmergencyTakeover.schedule()
+        }
+
+  @spec get_active_and_past_alerts() :: {[map()], [map()]}
```
digitalcora: This typespec could be made more specific, especially if we drop the JSON-encoding part from this function as suggested in the other comment.

lib/screenplay_web/router.ex
```diff
@@ -168,6 +168,7 @@ defmodule ScreenplayWeb.Router do
     post("/clear_all", AlertController, :clear_all)
     get("/active_alerts", AlertController, :active_alerts)
     get("/past_alerts", AlertController, :past_alerts)
```
digitalcora: Are both of these old endpoints still used? Does it make sense to have an "active" endpoint, a "past" endpoint, _and_ an "active and past" endpoint?

robbie-sundstrom: Good catch, the `past` endpoint isn't used; `active` is still used after you create a new alert to load the newly created active alert into the state.

lib/screenplay/emergency_takeovers.ex
robbie-sundstrom: you're right, that does work! This is my first time actually using Ecto since I learned Elixir so this kind of simplification is great to know

config/dev.exs
robbie-sundstrom: `image_store_module` is a much better name with this change

lib/screenplay/emergency_takeovers.ex
robbie-sundstrom: Framing the context module as returning the "raw" EmergencyTakeover data and leaving the JSON parsing to the controller actually cleans up my own framing of this a lot. I'm going to try a few changes related to that in the context module so thanks for the call out

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
robbie-sundstrom: This does make it a lot easier to differentiate the different break points. Fixed this throughout `alert_controller.ex`

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
robbie-sundstrom: This actually doesn't work as `%__MODULE__ ` because it doesn't yet have an `id` or timestamp fields at this point, which are required in the actual schema field. To tighten up the typing, I added a module for `NewEmergencyTakeover` that has only the required fields to create one in the DB.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
robbie-sundstrom: Added logic to cast the message field, and learned a lot about `Ecto` in the process! There's a new module for `MessageType` within `EmergencyTakeover` that casts this field now.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
robbie-sundstrom: This is still a list of station names because that still worked with the set up with Outfront and Showtime screens both being included in the `Places` setup. I did update this to `[station()]` for consistency.

assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
robbie-sundstrom: I just missed this TODO when finalizng the PR! I added a fix for this in my last commit that loads message image details for a canned message when the editing flow begins by calling the `onChange` function.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
digitalcora: I think I must have missed something in the implementation — if we are specifying station _names_, how do we know which Showtime screens this emergency message should take over?

assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx
```diff
@@ -14,6 +14,17 @@ const CreateMessage = ({ value, onChange }: CreateMessageProps) => {
...
+  useEffect(() => {
+    if (loading || value.type !== "canned" || value.id === -1) {
+      return;
+    }
+
+    const selectedMessage = cannedMessages.find((m) => m.id === value.id);
+    if (selectedMessage) {
+      onChange(selectedMessage);
+    }
+  }, [loading, value]);
```
digitalcora: This feels like another instance of [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect). Looking at just this code, I'm not even quite sure what circumstances would trigger this, what it does, and why it's necessary. Can we arrange to instead directly invoke the relevant logic within the code path that causes the update that then triggers this effect?

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,197 @@
...
+      with {:ok, normalized_message} <- cast(message) do
+        {:ok, stringify_keys(normalized_message)}
+      end
+    end
+
+    @impl true
+    def load(message), do: cast(message)
+
+    @impl true
+    def embed_as(_format), do: :self
```
digitalcora: Implementing this shouldn't be necessary as it's an optional callback and `:self` is the default behavior.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,197 @@
...
+
+    @impl true
+    def equal?(left, right) do
```
digitalcora: I'm unclear on the purpose of the `normalize` here — the basic `EctoURI` example in the docs doesn't seem to need it, and is storing data in a similar way. If we could remove it, implementing this shouldn't be necessary as comparing with `==` is the default behavior.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,197 @@
...
+      |> Enum.map(fn {k, v} -> {to_string(k), stringify_keys(v)} end)
+      |> Enum.into(%{})
+    end
+
+    def stringify_keys(value), do: value
+
+    @impl true
+    def dump(message) do
+      with {:ok, normalized_message} <- cast(message) do
+        {:ok, stringify_keys(normalized_message)}
```
digitalcora: I don't think we need this `stringify_keys` — the `EctoURI` example in the docs doesn't do this, it just returns a map with atom keys from `dump` (obtained from `Map.from_struct` in that case) and it seems Ecto handles the key conversion itself for `map`-type fields.

Assuming that is the case, this is the only use of `stringify_keys` within this module, and it could perhaps just be a private function in the controller... if it's necessary at all (if this is just for sending JSON to the client, Phoenix should automatically handle the stringifying of map keys).

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,197 @@
...
+      value
+      |> Enum.map(fn {k, v} -> {to_string(k), stringify_keys(v)} end)
+      |> Enum.into(%{})
+    end
+
+    def stringify_keys(value), do: value
+
+    @impl true
+    def dump(message) do
+      with {:ok, normalized_message} <- cast(message) do
```
digitalcora: Check me on this one but I think this is "double-casting" as the value which arrives at `dump` will have already been cast *if* indeed it has gone through a `Changeset.cast` — in the `EctoURI` example, their `dump` implementation only accepts "pre-cast" values and errors on anything else. Basically, it's your own fault if you get an error by directly setting a schema field to a value that doesn't actually match the type, without running it through casting first.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
```diff
@@ -0,0 +1,197 @@
...
+      start_time: schedule.start_time,
+      end_time: schedule.end_time,
+      created_by: Util.trim_username(user),
+      edited_by: Util.trim_username(user),
+      cleared_at: nil,
+      cleared_by: nil
+    }
+  end
+
+  def message_from_json(%{"type" => "canned", "id" => id}) do
```
digitalcora: Shouldn't this function be unnecessary now that we have casting integrated in the custom type? It doesn't seem like it's doing anything that `cast` isn't.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
digitalcora: > which are required in the actual schema field

I don't think this is correct, then, because a "new" record (which has not been saved yet) *will* have `nil` values for all of these fields, and that's as expected. We shouldn't need a separate duplicative `NewEmergencyTakeover` because the main record definition is already intended to represent both "new" (unsaved) and saved records.

lib/screenplay/emergency_takeovers.ex
digitalcora: Unresolving; I see you 👍'd this and then committed a change that made this function private, but then committed another change that made it public again.

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -250,11 +249,12 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
...
-  @spec remove_overlapping_alerts(map(), String.t()) :: :ok | {:error, String.t()}
-  defp remove_overlapping_alerts(params, user) do
-    stations_to_delete = State.remove_overlapping_alerts(params, user)
-
-    with :ok <- SFTP.clear_takeover_images(stations_to_delete),
+  @spec remove_overlapping_alerts(integer() | nil, map(), String.t()) ::
+          :ok | {:error, String.t()}
+  defp remove_overlapping_alerts(id, params, user) do
+    with stations_to_delete <-
+           EmergencyTakeovers.remove_overlapping_alerts(id, params["stations"], user),
```
digitalcora: Relevant to previous comments about removing unnecessary `<-` that will always match, this change seems to *introduce* one?

lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex
```diff
@@ -191,7 +182,7 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.AlertController do
   @spec upload_takeover_images(String.t(), %{String.t() => String.t()}) ::
           :ok | {:error, String.t()}
   def upload_takeover_images(alert_id, images) do
-    alerts_fetch_module = Application.get_env(:screenplay, :alerts_fetch_module)
+    image_store_module = Application.get_env(:screenplay, :image_store_module)
```
digitalcora: This seems like it could be at the module level as

```ex
@image_store Application.compile_env!(:screenplay, :image_store_module)
```

...since the field should be present at compile time in all environments.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
robbie-sundstrom: Ah good call! Modifying the `MODULE` type to have `id` and `inserted_at/created_at` be optionally `nil` allows this to work with dialyzer.

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
robbie-sundstrom: We actually do this Showtime Screen lookup in the same way that we would need to do it if it were based on station IDs by using the `Places` module. The specific screen IDs are pulled out in `showtime_screens_at_stations`. We actually do this Showtime Screen lookup in the same way that we would need to do it if it were based on station IDs by using the `Places` module. The specific screen IDs are pulled out in `showtime_screens_at_stations` within the `alert_controller`

lib/screenplay/emergency_takeover_tool/emergency_takeover.ex
robbie-sundstrom: I don't think this necessarily should give a warning, but it does flag a warning when this is not included, which is why I added it in the first place:
```function embed_as/1 required by behaviour Ecto.Type is not implemented (in module Screenplay.EmergencyTakeoverTool.EmergencyTakeover.MessageType)```

Comment (digitalcora) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: Instead of `@behaviour Ecto.Type`, you can `use Ecto.Type`, per the module docs:

> When you `use Ecto.Type`, it will set `@behaviour Ecto.Type` and define default, overridable implementations for `embed_as/1` and `equal?/2`.

Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: Ahhhh I see that now. Configuring it with `use Ecto.Type` indeed works!

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 25 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: This was to ensure that `cast` data is only compared with other `cast` data. I don't think in practice with the current implementation this would ever not be the case, so it should be fine to remove

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 28 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: Yes this should be updated to use `MessageType.cast()`

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 5 above)
Comment (digitalcora) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: Thanks, I see now the comment you made on the original implementation task that you found using station IDs was not necessary.

Regardless of the strict *necessity* of this, I still think it would have been a good idea to implement — station names can change, and from a data perspective we really shouldn't be using them as "primary keys" identifying a station. But at this point the cleanup would be optional and certainly outside the scope of this PR.

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 25 above)
Comment (digitalcora) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: I think in general implementers of these callbacks are not expected to do this, [same idea as this comment](https://github.com/mbta/screenplay/pull/760#discussion_r3476663427). The "contract" of Ecto schemas is that if you (the "user" of the schema) want to ensure the data you're storing is valid, it's your responsibility to run it through a changeset using `Changeset.cast` — casting and validation is never "done for you" anywhere else. If you bypass validation using `Changeset.change` or putting values directly into a struct, it's assumed you know what you're doing. It's sort of like the difference in TypeScript between `const foo: Foo = ...` and `const foo = ... as Foo`.

File: lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex (hunk 29 above)
Comment (robbie-sundstrom) [reply lib/screenplay_web/controllers/emergency_takeover_tool/alert_controller.ex]: missed this one, but fixed!

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 27 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: you're right here! This message will have already been `cast` by the time it's used by `dump`

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 26 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: The `stringify_keys` isn't needed in fact! didn't realize that Phoenix would handle turning the map keys into strings

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 25 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: ah thanks for the explanation! Makes sense that this would only happen for cast values unless we explicitly bypass

File: lib/screenplay/emergency_takeovers.ex (hunk 14 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeovers.ex:108]: I did make this private again and made sure the behaviors that were being tested were covered by the tests for `remove_overlapping_alerts` This just required adding one new test case for an empty new station list.

File: assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx (hunk 23 above)
Comment (robbie-sundstrom) [reply assets/js/components/EmergencyTakeoverTool/AlertWizard/CreateMessage.tsx]: Fair callout! I've got to stop relying on effects so much 😅  I made this logic happen during step advancement, because we need this before step 2 of the alert creation process to show the image preview. This structure makes a lot more sense and then is only invoked once during the flow when it's specifically needed.

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex (hunk 5 above)
Comment (robbie-sundstrom) [reply lib/screenplay/emergency_takeover_tool/emergency_takeover.ex]: I figured this would be the right time, so I updated this to store station IDs instead of names in my most recent commit. Wasn't too complex of a change, and I think it is a good idea for a long term data perspective

File: lib/screenplay/emergency_takeover_tool/emergency_takeover.ex [hunk 31]
```diff
@@ -68,7 +68,7 @@ defmodule Screenplay.EmergencyTakeoverTool.EmergencyTakeover do
   @type t() :: %__MODULE__{
           id: integer() | nil,
           message: message(),
-          stations: [station()],
+          station_ids: [station()],
```
digitalcora: suggestion: this type could now be named `station_id`

lib/screenplay/jobs/reminders.ex
```diff
@@ -27,8 +27,8 @@ defmodule Screenplay.Jobs.Reminders do
         Logger.debug("No outdated alerts found")
 
       alerts ->
-        Enum.each(alerts, fn %EmergencyTakeover{stations: stations} ->
-          stations
+        Enum.each(alerts, fn %EmergencyTakeover{station_ids: station_ids} ->
+          station_ids
```
digitalcora, line 32: Noting that this changes the behavior of the Slack integration to send the stop IDs in the message rather than their user-friendlier names — though I think this is acceptable given the frequency of this tool's use, and the target audience of the message can just sign into the ETT to see the actual station names.

lib/screenplay_web/controllers/emergency_takeover_tool/page_controller.ex
```diff
@@ -58,7 +58,8 @@ defmodule ScreenplayWeb.EmergencyTakeoverTool.PageController do
 
         {place.name,
          %{
-           name: place.name,
+           # name: place.name,
```
digitalcora: ✂️

digitalcora (approved):
Looking good!! Really appreciate you working through this including the totally optional side quest I proposed 😸 I think we should deploy to dev-green and give it a shot in a "real" environment before merging, given the database migration is probably final, though of course for the full takeover functionality it should also be tested in prod with the real OFM screens.

Also, there is a probably minor type issue being surfaced by CI on the current commit.

PR #765, opened 2026-06-24
fix: filter pa messages with an ended alert

rwaskiewicz:
**Asana task**: [Bug - Delay in closing Alert-associated PA Messages](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215504297867591)

Description

Filter PA messages that are associated with an ended alert. Prior to this commit, a PA message associated with an ended alert would continue to be shown in the PA messages section of the application until the alert fell off the data feed. With this commit, the end date of the associated alert is picked up faster than the alert dropping off, allowing the PA message to be removed from the PA messages list before the ~five minute period has elapsed.

Note: When working on this fix, I found a related bug where ended alerts are not being removed from the "associate this new pa message with an alert" list. Fixed in https://github.com/mbta/screenplay/pull/766

### DIFF
--- a/lib/screenplay/pa_messages.ex
+++ b/lib/screenplay/pa_messages.ex
@@ -57,14 +57,14 @@ defmodule Screenplay.PaMessages do
       |> Map.put_new_lazy(:now, &DateTime.utc_now/0)
       |> Map.put_new(:state, :all)
 
-    alert_ids = AlertsCache.alert_ids()
+    alerts = AlertsCache.alerts()
 
     signs_for_routes = RoutesToSigns.signs_for_routes(opts[:routes])
 
     now = opts[:now] || DateTime.utc_now()
 
     PaMessage
-    |> PaMessage.Queries.state(opts[:state], alert_ids, now)
+    |> PaMessage.Queries.state(opts[:state], alerts, now)
     |> PaMessage.Queries.signs(opts[:signs])
     |> PaMessage.Queries.signs(signs_for_routes)
     |> order_by(desc: :inserted_at)
@@ -88,7 +88,7 @@ defmodule Screenplay.PaMessages do
   @spec get_active_messages() :: [PaMessage.t()]
   @spec get_active_messages(now :: DateTime.t()) :: [PaMessage.t()]
   def get_active_messages(now \\ DateTime.utc_now()) do
-    AlertsCache.alert_ids()
+    AlertsCache.alerts()
     |> PaMessage.Queries.active(now)
     |> Repo.all()
   end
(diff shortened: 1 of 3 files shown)


lib/screenplay/pa_messages/pa_message/queries.ex
```diff
@@ -42,20 +43,32 @@ defmodule Screenplay.PaMessages.PaMessage.Queries do
...
+    alert_ids = alerts |> Enum.reject(&alert_will_fall_off?(&1, now)) |> Enum.map(& &1.id)
+
     from m in q,
       where:
         m.start_datetime <= ^now and
           ((is_nil(m.end_datetime) and m.alert_id in ^alert_ids) or m.end_datetime >= ^now)
   end
 
+  @spec alert_will_fall_off?(Alert.t(), DateTime.t()) :: boolean()
+  defp alert_will_fall_off?(%Alert{active_period: [{_, period_end}]}, now)
```
digitalcora, line 58: suggestion: this could be better-named, "will fall off" is a description of an API behavior that happens to be the case today, but could be different in the future (it's also not super clear what "fall off" refers to; this isn't a terminology we commonly use anywhere). `alert_closed?` would be more descriptive of the concept we're getting at.

lib/screenplay/pa_messages.ex
```diff
@@ -57,14 +57,14 @@ defmodule Screenplay.PaMessages do
       |> Map.put_new_lazy(:now, &DateTime.utc_now/0)
       |> Map.put_new(:state, :all)
 
-    alert_ids = AlertsCache.alert_ids()
+    alerts = AlertsCache.alerts()
```
digitalcora, line 60: Is there a reason we need `AlertsCache.alerts/0` to return all alerts that are in the feed, even ones that are closed? I guess I'm wondering why this filter was implemented "downstream" of the alerts cache in `PaMessage.Queries`, when presumably no part of Screenplay cares about closed alerts and we could perhaps filter them out "at the source". (this would also automatically fix the issue from #766)

rwaskiewicz: Hmmm good point 🤔 Regardless of putting the change in the cache or in queries, I suppose we're always going to be subject to "waiting" on our call-loop to V3 to get the latest Alert data, making this change (and making the issue in #766) more verbose than putting it in the cache.

I've made the change in #767 since that was more of change in direction + #766 was dependent on this PR. If #767 is more appropriate, I'll close this and #766

rwaskiewicz:
Superseded by #767

PR #766, opened 2026-06-24
 fix: hide stale alerts when associating with pa messages 

rwaskiewicz:
**Asana task**: [Bug - Delay in closing Alert-associated PA Messages](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215504297867591)

Prior to this commit, when associating a new PA message with an alert,
closed alerts were displayed. This commit updates the logic as to not
retrieve said alerts.

Depends On: https://github.com/mbta/screenplay/pull/765

Description

- [ ] For features with a design/UX component, deployed branch to dev-green and let product know it's ready for review.

PR #767, opened 2026-06-26
fix: hide stale alerts, close pa messages with assoc alerts

rwaskiewicz:
**Asana task**: [Bug - Delay in closing Alert-associated PA Messages](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215504297867591)

Description

Prior to this commit, when associating a new PA message with an alert, closed alerts were displayed. Filtering alerts at retrieval time based on whether they are closed or not keeps closed alerts out of the cache. By preventing said alerts from entering the cache:

1. Alerts that are closed will not be shown when associating a new PA message with an alert
2. Exising PA messages associated with an alert whose alert has been closed will be removed from the PA Messages list without having to wait for the alert to "fall off" the data feed.

This change does not alter the Alert fetching logic. A consequence of that is changes to Alerts will not instantly be refelcted in Screenplay. Rather, we are (still) reliant on fetching Alerts from the V3 API to get the latest alert information.

Replaces: #765, #766 

### DIFF
--- a/lib/screenplay/alerts/alert.ex
+++ b/lib/screenplay/alerts/alert.ex
@@ -165,6 +165,15 @@ defmodule Screenplay.Alerts.Alert do
     DateTime.compare(t, start_t) in [:gt, :eq] && DateTime.compare(t, end_t) in [:lt, :eq]
   end
 
+  @spec closed?(t(), DateTime.t()) :: boolean()
+  def closed?(alert, now \\ DateTime.utc_now())
+
+  def closed?(%__MODULE__{active_period: [{_, period_end}]}, now) when period_end != nil do
+    DateTime.compare(period_end, now) in [:lt, :eq]
+  end
+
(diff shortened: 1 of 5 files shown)


lib/screenplay/alerts/alert.ex
```diff
@@ -165,6 +165,15 @@ defmodule Screenplay.Alerts.Alert do
     DateTime.compare(t, start_t) in [:gt, :eq] && DateTime.compare(t, end_t) in [:lt, :eq]
   end
 
+  @spec closed?(t(), DateTime.t()) :: boolean()
+  def closed?(alert, now \\ DateTime.utc_now())
```
digitalcora, line 169: A test for this function and/or the `now`-filtering behavior of the alerts cache would be good to have.

digitalcora (approved):

PR #776, opened 2026-07-07
start endpoint process last

panentheos:
**Asana task**: [Screenplay: Fix process ordering to reduce errors on shutdown](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1211731814923660?focus=true)

This makes the endpoint process the last one to launch, and the first one to shut down. This should avoid ETS or GenServer errors on shutdown due to the controllers trying to call various supporting processes after they've been terminated.


digitalcora (approved):
Sensible!

PR #777, opened 2026-07-08
chore: group dependabot updates

rwaskiewicz:
**Asana task**: N/A

Description

In 8c6bfdac, Dependabot was configured to update GitHub Actions. In the 2026.07.07 retro, we commented that the number of individual PRs was getting high. This commit attempts to cut down on the noise by grouping updates of all actions across all semver version types.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,10 @@ updates:
     cooldown:
       default-days: 7
     open-pull-requests-limit: 5
+    groups:
+      all-updates:
+        patterns:
+          - "*"
   - package-ecosystem: mix
     directory: "/"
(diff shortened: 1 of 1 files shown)


digitalcora (approved):
This seems fine, although maybe we should consider only grouping non-major updates together, as we do for other ecosystems? The rationale that major updates are more likely to be breaking changes for us, and that we wouldn't want them to hold up other updates in that case, still seems like it would apply.

rwaskiewicz:
@digitalcora SGTM - done in https://github.com/mbta/screenplay/pull/777/commits/e1e8b959732a146842cab922ca77044b9dc9357d

PR #783, opened 2026-07-14
add line filtering for alert places

panentheos:
**Asana task**: [Screenplay: Sort/filter stations in Posted Alerts](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215387822203482?focus=true)

This adds the line/mode filter to the places list on the alert detail page. The filter defaults to a certain line if all places are on the same line, and uses the corresponding station sort order. Otherwise, it defaults to All, and uses alphabetical sorting. In either case, the filter can be changed to see a subset of the impacted places.

Note that several pages now wait until the places and alerts are loaded before displaying their content, which is necessary to properly initialize the defaults, and has the side effect of removing some flashes of empty content, e.g. the erroneous "not found" message on the alert details page.

### DIFF
--- a/assets/js/components/Dashboard/AlertDetails.tsx
+++ b/assets/js/components/Dashboard/AlertDetails.tsx
@@ -28,7 +28,7 @@ const AlertDetails: ComponentType = () => {
 
   const navigate = useNavigate();
 
-  const foundAlert = alerts.find((alert) => alert.id === id);
+  const foundAlert = (alerts ?? []).find((alert) => alert.id === id);
   useEffect(() => {
     if (foundAlert) {
       setSelectedAlert(foundAlert);
@@ -37,9 +37,17 @@ const AlertDetails: ComponentType = () => {
 
   const validAlertId = id && allAPIAlertIds.includes(id) ? id : undefined;
 
+  if (!alerts || !places) {
+    return null;
+  }
+  const alertPlaces = placesWithSelectedAlert(
+    selectedAlert,
+    places,
+    screensByAlertMap,
+  );
   return selectedAlert ? (
     // Define a new ContextProvider so state is not saved to Context used on the PlacesPage.
-    <PlacesListStateContainer>
+    <PlacesListStateContainer places={alertPlaces}>
       <div className="alert-details">
         <div className="page-content__header">
           <div>
@@ -69,14 +77,10 @@ const AlertDetails: ComponentType = () => {
           <div className="page-content__body">
             <AlertCard alert={selectedAlert} classNames="selected-alert" />
             <PlacesList
-              places={placesWithSelectedAlert(
-                selectedAlert,
-                places,
-                screensByAlertMap,
-              )}
-              noModeFilter
+              places={alertPlaces}
               isAlertPlacesList
               showAnimationForNewPlaces
+              showLineMap={false}
             />
           </div>
         )}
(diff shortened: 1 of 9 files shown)


assets/js/components/Dashboard/AlertDetails.tsx
```diff
@@ -37,9 +37,17 @@ const AlertDetails: ComponentType = () => {
 
   const validAlertId = id && allAPIAlertIds.includes(id) ? id : undefined;
 
+  if (!alerts || !places) {
+    return null;
+  }
```
rwaskiewicz, line 42: Does it make sense to hoist the letter cahnge above the definition of `foundAlert`? That would eliminate the need for the change to add nullish coalescing to `alerts` (Note: GitHub is rendering this suggestion in a nonsensical way on my end, hopefully it comes out the other side looking better):

```suggestion
  if (!alerts || !places) {
    return null;
  }
  
  const foundAlert = alerts.find((alert) => alert.id === id);
  useEffect(() => {
    if (foundAlert) {
      setSelectedAlert(foundAlert);
    }
  }, [foundAlert]);

  const validAlertId = id && allAPIAlertIds.includes(id) ? id : undefined;
```

assets/js/components/Dashboard/PlacesPage.tsx
```diff
@@ -175,6 +179,11 @@ const PlacesList: ComponentType<PlacesListProps> = ({
     statusFilterValue !== STATUSES[0] ||
     screenTypeFilterValue !== SCREEN_TYPES[0];
 
+  const allPlaceRoutes = fp.uniq(places.flatMap(({ routes }) => routes));
+  const modeOptions = MODES_AND_LINES.filter(({ ids }) => {
+    return ids[0] === "All" || fp.intersection(ids, allPlaceRoutes).length > 0;
+  });
```
rwaskiewicz, line 185: No action required - FWIW, we could accomplish this without lodash (but would require us to update our TS config for 2024's `Set` methods):
```ts
  const allPlaceRoutes = new Set(places.flatMap(({ routes }) => routes));
  const modeOptions = MODES_AND_LINES.filter(({ ids }) => {
    return ids[0] === "All" || allPlaceRoutes.intersection(new Set(ids)).size > 0;
  });
```

assets/js/hooks/useScreenplayContext.tsx
```diff
@@ -156,10 +157,30 @@ interface PlacesListState {
...
+      return places
+        ? fp
+            .reverse(PLACES_PAGE_MODES_AND_LINES)
+            .find(
+              ({ ids }) =>
+                ids[0] === "All" ||
+                places.every(
+                  (place) => fp.intersection(ids, place.routes).length > 0,
+                ),
+            )!
```
rwaskiewicz, line 181: No action required - Similar comment above RE doing this in vanilla JS:
```ts
    ? PLACES_PAGE_MODES_AND_LINES.toReversed().find(
        ({ ids }) =>
          ids[0] === "All" ||
          places.every(
            (place) =>
              new Set(ids).intersection(new Set(place.routes)).size > 0,
          ),
      )!;
```

rwaskiewicz (commented):
When testing this out, it looked like "Ferry" was auto selected as filter, which doesn't seem to be happening on `main`. Could you take a look? 

https://github.com/user-attachments/assets/17dc73b7-b71b-4cf6-a7db-ac993d63e7cb

assets/js/components/Dashboard/AlertDetails.tsx
panentheos: Unfortunately we can't do that because that would conditionally short-circuit the `useEffect`, which isn't allowed.

panentheos:
Ah, that Ferry behavior is happening because the list of places is empty, so the `every` check succeeds on the first reversed (e.g. last) filter option. I'll add a conditional to choose "All" in that case, which seems like the sensible choice.

rwaskiewicz (approved):

PR #784, opened 2026-07-15
fix: add text validation to pa/ess text messages

rwaskiewicz:
**Asana task**: [Screenplay: invalid characters in PA messages](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1216523723126553?focus=true)

Description

Add text validation to visual messages sent to signs. The set of allowed
characters is largely borrowed from [Realtime Signs](https://github.com/mbta/realtime_signs/blob/831f805027295590bd5f21c77a9720a8e160e203/lib/pa_ess/utilities.ex#L609)
with the following additions:
* hyphens ('-'), as these were the basis of the original issue where an
endash was added to a message and not processed properly by the hardware
* periods ('.'), to write whole sentences that are visually complete

No validation was added to the text area for predefined templates, as we
assume that the text provided there are correct.

Screenshot 2026-07-15 at 15 17 21
Screenshot 2026-07-15 at 15 24 17

\* For posterity `@ol` is meant to be a dial-up joke, not Orange Line commentary 😆 

### DIFF
--- a/assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
+++ b/assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
@@ -25,6 +25,7 @@ import * as paMessageStyles from "Styles/pa-messages.module.scss";
 import type { StaticTemplate, TemplateType } from "Models/static_template";
 
 const MAX_TEXT_LENGTH = 2000;
+const INVALID_VISUAL_TEXT_CHARACTERS = /[^a-zA-Z0-9,\/!@':\s\-.]/g;
 
 interface Props {
   title: string;
@@ -157,6 +158,12 @@ const MainForm = ({
   const popoverText = "The service day starts and ends at 4:00 AM";
 
   const isEndTimeInvalid = validated && !moment(endTime, "HH:mm").isValid();
+
+  const invalidVisualTextInputs = new Set(
+    [...visualText.matchAll(INVALID_VISUAL_TEXT_CHARACTERS)].map(
+      (match) => match[0],
+    ),
+  );
 
   return (
     <div className={paMessageStyles.editPage}>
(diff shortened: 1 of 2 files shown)


assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
```diff
@@ -67,6 +69,14 @@ interface Props {
   isReadOnly?: boolean;
 }
 
+export const getInvalidVisualText = (visualText: string): Set<string> => {
```
rwaskiewicz, line 72: I don't love exporting this just to test it from a philosophical level, but practically this was an easier way to get this under test without setting up the form and interacting with it.

assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
```diff
@@ -25,6 +25,8 @@ import * as paMessageStyles from "Styles/pa-messages.module.scss";
 import type { StaticTemplate, TemplateType } from "Models/static_template";
 
 const MAX_TEXT_LENGTH = 2000;
+// A negated set regex used to limit the visual text sent to signs.
+const INVALID_VISUAL_TEXT_CHARACTERS = /[^a-zA-Z0-9,/!@':\s\-.]/g;
```
rwaskiewicz, line 29: See my comment in the PR summary regarding the addition of hyphens and periods. Should we extend the RTS characters to include hyphens and periods?

robbie-sundstrom: I thought based on this ticket that regular hyphens were allowed in RTS, so this seems like a good catch. Might want to confirm with Brett that there isn't a reason it's not included in the RTS allow list.

rwaskiewicz: Likewise, unless I'm referencing the wrong source of truth in RTS 🤔

rwaskiewicz: @robbie-sundstrom in talking to Brett, this is fine for now. At some point, we should standardize between here, Signs UI and RTS. To do that, we should come up with a verified list of all allowed characters (this scope could include Spanish accented chars). From there we use that list in all three. So the plan (for now) is to proceed with this as-is, and I'll create a ticket in the backlog (if I can't find one) for that reconciliation work

robbie-sundstrom (approved):
love using regex to actually pull out the invalid characters for the error message 😎

assets/js/components/Dashboard/PaMessageForm/MainForm.tsx
rwaskiewicz: [Created this task](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216635602500735) to reconcile the three systems

